PR #77, opened 2018-06-14
Composing sebak network using docker

spikeekips:
### Background
Current docker support to deploy SEBAK network is limited to run 3 nodes. To make it more flexible, created new command, `sebak-compose-network`. In future it can be integrated with the main `sebak`.

With `--n` option, it runs the given number of nodes using docker. To make this process simple, instead of generate or make 'docker-compose' file or others, tried to use native docker API.

```
$ cd cmd/sebak-compose-network
$ go run main.go -h
sebak composing network

Usage:
  /var/folders/7q/cfmzp61d23scjv7645llr25r0000gn/T/go-build012630019/b001/exe/main [flags]

Flags:
      --force                    remove the existing sebak containers
  -h, --help                     help for /var/folders/7q/cfmzp61d23scjv7645llr25r0000gn/T/go-build012630019/b001/exe/main
      --image string             docker image name for sebak (default "boscoin/sebak/compose-network:latest")
      --log-level string         log level, {crit, error, warn, info, debug} (default "info")
      --n uint                   number of node (default 3)
      --sebak-log-level string   sebak log level, {crit, error, warn, info, debug} (default "dbug")
```

Before running command, building docker image for 'compose-network' should be proceeded. The details was already described in the bottom of README.md.

This composing network feature will be developed like [running various cases of ISAAC network](https://github.com/bosnet/isaac-consensus-protocol/tree/master/examples/cases).


Geod24:
We don't need another Docker image, the one already in the repo is enough. We also don't need the SSH key for deployment (I did the same mistake in my original PR, until @outsideris made me realize it was unnecessary).
And as noted before, integrating it with sebak feels just wrong, as it completely break separation of concern.

spikeekips:
> @Geod24 : We don't need another Docker image, the one already in the repo is enough. 
The current docker image just has final build binary, `/go/bin/sebak`. The image of this PR has the whole git enabled source code, I intended this image to pull the changes from remote repository.

> We also don't need the SSH key for deployment (I did the same mistake in my original PR, until @outsideris made me realize it was unnecessary).

Yes, it will be unnecessary if SEBAK will be open source. The included ssh will be used temporarily to access the private repo.

> And as noted before, integrating it with sebak feels just wrong, as it completely break separation of concern.
I agree. Providing docker related files and features can be managed by other sub-projects.

Geod24:
> Yes, it will be unnecessary if SEBAK will be open source. The included ssh will be used temporarily to access the private repo.

Yeah but the problem here is that we are `go get`ing inside a docker image, instead of just `COPY`ing the repository. This also means that when we want to test something, we won't be able to with this setup, because it will get whatever is remote, not the working directory.

spikeekips:
@Geod24 Dockerfile.compose-network use ADDed source at build time and it’s entrypoint.sh has SEBAK_SRC_COMMAND env variable, using this we can inject and run the arbitrary commands like pull PR or change branch.

outsideris:
I disagreed with the new command because of it is not sebak's responsibility, at least I feel like that. It is only for test purpose. Delagating to other tools is better. Or, we can make another new repository for these kinds of tools.

Geod24:
> Dockerfile.compose-network use ADDed source at build time and it’s entrypoint.sh has SEBAK_SRC_COMMAND env variable, using this we can inject and run the arbitrary commands like pull PR or change branch.

Yeah but it's a tedious process: you have to know it's here, get the right command, and it clearly doesn't scale. However I agree the original Dockerfile needs a bit of improvement and will submit a PR soon.

Maybe we can sit down tomorrow and see what is missing from the functionality in master ? That should reduce communication overhead!

spikeekips:
@outsideris Yes, I agree. I just want to introduce the new way to deploy network. It does not matter this feature exists in this project or not. 

If @outsideris @Geod don’t mind, I will cancel this PR :)

Geod24:
I don't mind

spikeekips:
ㅠㅠㅠ ㅋㅋㅋㅋ

outsideris:
@Geod24 👍

spikeekips:
This changes will be continued in my personal project :)

Issue #76, opened 2018-06-12
Proposal : Using custom import path “blockchainos.org/sebak”? 


anarcher:
Go has a mechanism[1] for defining custom import paths by serving HTML meta tags with repository information. Sometimes, Using github import path (github.com/org/repo/...) bothers me. and it doesnt make easy to change repo location too. 
[1]: https://golang.org/cmd/go/#hdr-Remote_import_paths

outsideris:
Set up `https://boscoin.io/sebak/`.
Now it works.

PR #78, opened 2018-06-14
'github.com/owlchain/sebak' -> 'boscoin.io/sebak'

spikeekips:
### Github Issue
#76 

### Background
The current repository org, 'owlchain' is for private development, so to publishing SEBAK as open-source, we need to move to 'bosnet' org. The keep our identity to be clearer, the new go import path will be better, the name is 'boscoin.io'.

* New source repository will be "https://github.com/bosnet/sebak"
* New go import path will be "boscoin.io/sebak"

I already renamed the import path to "boscoin.io/sebak" from "github.com/owlchain/sebak" and check to pass the `go test ./...` also too.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,8 +3,8 @@ FROM golang:alpine AS builder
 LABEL maintainer="BOSCoin Developers <devteam@boscoin.io>"
 
 #golang:alpine set $GOPATH to `/go`
-COPY ./ /go/src/github.com/owlchain/sebak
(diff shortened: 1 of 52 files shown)


Geod24:
IIUC this is supposed to implement #76 ?
Then we need a change on our website to serve the metadata, don't we ? At the moment the repo doesn't build.

spikeekips:
@Geod24 Yes, right. #76 is proceeded by @anarcher, so all the remaining work will be finished ASAP :)

Issue #74, opened 2018-06-08
`sebakstorage.LevelDBBackend.GetRaw()` should return the 'ErrorStorageRecordDoesNotExist' error


spikeekips:
Currently `sebakstorage.LevelDBBackend.GetRaw()` returns the simple error when record does not exist, but to distinguish the error clearly, we need to define new error, `sebakerror.ErrorStorageRecordDoesNotExist`. if `GetRaw()` returns `leveldb.ErrNotFound` or `Has()` returns false, `sebakerror.ErrorStorageRecordDoesNotExist` will return.

anarcher:
@spikeekips @zzim2x Plz review it again~

anarcher:
I will close it.

PR #81, opened 2018-06-14
LevelDBBackend.GetRaw() can return ErrorStorageRecordDoesNotExit #74

anarcher:
#74 

-  Add `ErrorStorageRecordDostNotExit` in `lib/error`
- `LevelDBBackend.GetRaw()` returns the error when record dose not exist. 

ps) It's reopened PR  because my forked repo is broken after changing repo.

### DIFF
--- a/lib/storage/leveldb.go
+++ b/lib/storage/leveldb.go
@@ -5,12 +5,14 @@ import (
 	"errors"
 	"fmt"
 
-	"boscoin.io/sebak/lib/common"
 	"github.com/syndtr/goleveldb/leveldb"
 	leveldbIterator "github.com/syndtr/goleveldb/leveldb/iterator"
 	leveldbOpt "github.com/syndtr/goleveldb/leveldb/opt"
 	leveldbStorage "github.com/syndtr/goleveldb/leveldb/storage"
 	leveldbUtil "github.com/syndtr/goleveldb/leveldb/util"
+
+	"boscoin.io/sebak/lib/common"
+	"boscoin.io/sebak/lib/error"
 )
 
 type LevelDBCore interface {
@@ -100,8 +102,8 @@ func (st *LevelDBBackend) Has(k string) (bool, error) {
 func (st *LevelDBBackend) GetRaw(k string) (b []byte, err error) {
 	var exists bool
 	if exists, err = st.Has(k); !exists || err != nil {
-		if !exists {
-			err = fmt.Errorf("key, '%s' does not exists", k)
+		if !exists || err == leveldb.ErrNotFound {
+			err = sebakerror.ErrorStorageRecordDoesNotExist
 		}
 		return
 	}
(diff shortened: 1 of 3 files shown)


anarcher:
@spikeekips @zzim2x Plz review it again~

lib/storage/leveldb_test.go
```diff
@@ -133,6 +134,11 @@ func TestLevelDBBackendGetRaw(t *testing.T) {
 	input := "findme"
 
 	st.New(key, input)
+
+	//when record does not exist,it should return ErrorStorageRecordDoesNotExist
```
outsideris: Add a whitespace after comma.

lib/storage/leveldb.go
```diff
@@ -100,8 +102,8 @@ func (st *LevelDBBackend) Has(k string) (bool, error) {
 func (st *LevelDBBackend) GetRaw(k string) (b []byte, err error) {
 	var exists bool
 	if exists, err = st.Has(k); !exists || err != nil {
-		if !exists {
-			err = fmt.Errorf("key, '%s' does not exists", k)
+		if !exists || err == leveldb.ErrNotFound {
+			err = sebakerror.ErrorStorageRecordDoesNotExist
```
outsideris: I didn't understand this code because I didn't know golang initialize variables in return type.
I don't decide my stands for this feature, but I want to talk if we use this feature heavily or not.

Geod24: I personally find named return very confusing, but it's quite heavily used through Sebak. It would be nice to have this discussion, yes.

cmcm2222: um.. 
according to this feature, should we define existing error like this??
for example; 

''' func (st *LevelDBBackend) Discard() error {
	ts, ok := st.core.(*leveldb.Transaction)
	if !ok {
		return errors.New("this is not *leveldb.Transaction")
	}

	ts.Discard()
	return nil
}

func (st *LevelDBBackend) Commit() error {
	ts, ok := st.core.(*leveldb.Transaction)
	if !ok {
		return errors.New("this is not *leveldb.Transaction")
	}

	return ts.Commit()
} ''' 

are have same error message.

outsideris: @cmcm2222 your code block is broken.

anarcher: @Geod24 @outsideris IMHO, the named return is outside of this PR.  :-) for discussing it, let make a new issue or do something else. :-)

@cmcm2222 I am not sure but that we should return a custom error from that methods. I think that it also is outside of this PR :-) Let we discuss it! :-)  But in my opinion, that errors must not be returned because of this cause by type miscasting. When we meet that errors, It is a bug.

cmcm2222: @outsideris yep I've changed that. thanks. 
@anarcher Thank you for kind explanation. so helpful.

anarcher:
Is it ok to merge? @Geod24

PR #82, opened 2018-06-14
Use the checked out dependencies for Docker

Geod24:
```
Previously, building `sebak` would result in a ~1.2GB image every single time.
Due to `dep ensure`'s inneficiency, it would also take several minutes.
It would also prevent messing with the dependencies for testing purpose.

As it's much more likely that this image is used during development / testing,
we do not need clean dependencies every time,
and the speed increase makes development / testing cycle realistic again.
```


Geod24:
@cmcm2222 : I added a second commit to allow calling sebak with any argument. Can you review it as well ?

cmcm2222:
@Geod24  OK. I wil.. :)

Dockerfile
```diff
@@ -6,12 +6,12 @@ LABEL maintainer="BOSCoin Developers <devteam@boscoin.io>"
 COPY ./ /go/src/boscoin.io/sebak
```
spikeekips: Using `go install` will be good idea for fast build without `dep ensure` if our project is still in private repo. After this project will be opened, docker image should be applied the latest updated source code from 'master' branch. I think, instead of `COPY` the local source code, `go get boscoin.io/sebak/...` would be natural.

docker/entrypoint.sh
```diff
@@ -1,9 +1,19 @@
...
-for VALIDATOR in ${SEBAK_VALIDATORS}; do
-    VALIDATOR_ARGS="${VALIDATOR_ARGS} --validator=${VALIDATOR}"
-done
-./sebak node ${VALIDATOR_ARGS} --network-id=${SEBAK_NETWORK_ID} --secret-seed=${SEBAK_SECRET_SEED} --log-level debug
+## This entrypoint has 2 modes: if any argument is provided to `docker run`,
+## the arguments are passed directly to sebak
+## Otherwise, it just starts a node with the environment file
+if [ $# -gt 1 ]; then
+    ## Argument mode
+    ./sebak $@
```
spikeekips: Good feature. If extra options is given, the `SEBAK_XX` env variables seem to be ignored. It will be pressure to set the entire options to add one option.

Dockerfile
spikeekips: I missed something. `go get ...` does not fit our dependencies. I think `dep ensure` should be done.

```
$ time go get boscoin.io/sebak/...
# boscoin.io/sebak/lib/common
src/boscoin.io/sebak/lib/common/util.go:23:29: multiple-value uuid.NewV1() in single-value context
src/boscoin.io/sebak/lib/common/util.go:27:29: multiple-value uuid.NewV4() in single-value context

real	3m45.278s
user	0m35.393s
sys	0m17.996s
```

This is caused by 'github.com/google/uuid', we use '0.2.0' version of uuid, but the master of uuid was updated.

docker/entrypoint.sh
Geod24: I just tested it, the environment variable are not ignored. `VALIDATOR_ARGS` is ignored, since it's not a native env variable for sebak, but `SEBAK_NETWORK_ID` for example will work.

Note that the first mode is designed to quick CLI testing, while the second mode is more useful if you want to spawn a network. For example @cmcm2222 and I use this mode to test our implementation of #1 quickly.

spikeekips: @Geod24 yes, I was wrong, `sebak` already knows the env variables :)

Dockerfile
Geod24: It depends what the intended usage is.

This Dockerfile is designed for developer / testers, that's why it lives in the same repository. It provide a very quick edit-compile-test cycle. It allows one to spawn a network quickly because that's part of the cycle: while at the moment we're working on the CLI usage, later on when we work on the node aspect, we can use the same Dockerfile to change one or more instances in a network quickly and feed them messages, to see if our changes make sense.

What you are describing, if I am not mistaken, is a Dockerfile which "just works", provides a clean build, and uses the latest version of the code, I suggest we put that image in another repository, because there's no point in having it within the sebak source tree, since it will be totally independent from the state of the repository.

Does it make sense ?

spikeekips: @Geod24 Most of other project to provide `Dockerfile`, the `Dockerfile` in the root means that it can make official image for the project. If you intended this `Dockerfile` to be only for testing, the current location of `Dockerfile` makes to be confused.

outsideris: IMO, Dockerfile should depend on the local source, not GitHub repository. It is common use cases.
And Only pushed docker image is based on latest source code. But we can use automated build on Docker Hub with integrating GitHub.

Geod24: `COPY` cannot access path outside of the build context. If `Dockerfile` did not live at the root of the repository, it wouldn't be able to `COPY` the context (repository). So while an official image can live in a subdirectory, the one used for testing the code in the repo can't.

spikeekips: @outsideris Hmm, depending on the local source code, does not guarantee it is based on the latest code or not, so not good way~, I think.

Geod24: @spikeekips : That's exactly what @outsideris mentioned. We're not going to push images based on the root Dockerfile. The one we're going to push will depend on a `Dockerfile` which does a clean build (with `go get` and `dep ensure`).

spikeekips: > @Geod24 : COPY cannot access path outside of the build context. If Dockerfile did not live at the root of the repository, it wouldn't be able to COPY the context (repository). So while an official image can live in a subdirectory, the one used for testing the code in the repo can't.

```
$ docker build -f docker/testing/Dockerfile -t SEBAK-testing .
```
This way would be possible with sub directory of `Dockerfile`.

Dockerfile
```diff
@@ -6,12 +6,12 @@ LABEL maintainer="BOSCoin Developers <devteam@boscoin.io>"
...
-# `git` and `openssh` are needed for `go get`
-RUN apk add --no-cache git openssh          \
-    && go get github.com/golang/dep/cmd/dep \
-    && dep ensure -v                        \
-    && go install -v ./...
+## Note that we do not get the dependencies anew
+## We carry over whatever is in `vendor`, so the user MUST run `dep ensure` in their local copy
+## This make building the container orders of magnitude faster (`dep ensure` is extremely slow),
+## greatly reduce the container's size, and gives more control to the user as to what is tested
+## (one can replace a dependency, if needed).
```
outsideris: I'm not sure about this.
What people expected when they see Dockerfile, `docker build ...` should be just worked.
Some pre-process is a little bit confused.

Additionally, why container's size reduced even if we use docker multi-staged build?

Dockerfile
spikeekips: > @Geod24 : We're not going to push images to the Docker hub which are based on the root Dockerfile.

Yeah, you already said, this `Dockerfile` is for testing, not publicly opened. I just pointed the location of `Dockerfile`.

outsideris: If `Dockerfile` use source code from GitHub, Docker images can be different by the build timing.
For instance, I build a docker image and build again 5 mins later, two images can be different if we push some commits during the builds. It can be very confused.

Geod24: That's correct, and I did not consider this solution.
I'm not so found of it because it forces you to get more context (the `-f` option is not optional anymore, and Docker's default expects the `Dockerfile` to be at the root of the context) but if we decide to store multiple Docker file in this repository it's a viable solution 👍

spikeekips: @outsideris It depends on how we manage the branch to build ;) New version release, new docker image could be released with tags :)

Dockerfile
outsideris: I don't have a current solution, but...
Our problem is our dependencies are too large. So, our direction should be reducing our dependencies. If this approach is temporary, it is ok. However, I'm worrying about we forget the dependencies issue if this mitigates the docker build time. Docker's build time is not a critical issue for now.
And I don't know `dep` mechanism much as well.

Geod24: > Docker's build time is not a critical issue for now.

It makes the edit-compile-run cycle very slow because the time spent `dep ensure`ing dominates by at least an order of magnitude, which is why it came as a problem to me.

Another approach I thought of this weekend would be to use volumes and mount the `vendor` directory, so that `dep ensure` cache is kept. Not sure that's the best solution though. Are we fine with merging this as a temporary solution ?

PR #84, opened 2018-06-15
API: Account changes to event stream

kfangw:
### Requirements

* Implement SSE(Sever Sent Event)

### Github Issue
#3 

### Solution
I changed two point
- server-sent-event api with go-observerble
- Adopt mux

After this change, i will add bellows
- client
- APIs for each issues

### DIFF
--- a/lib/api.go
+++ b/lib/api.go
@@ -6,12 +6,15 @@ import (
 	"net/http"
 
 	"boscoin.io/sebak/lib/network"
+	"boscoin.io/sebak/lib/storage"
 	"github.com/GianlucaGuarini/go-observable"
 )
 
-func AddAPIHandlers(ctx context.Context, t *sebaknetwork.HTTP2Network) {
-
-	t.AddAPIHandler(ctx, "/account/{address}", GetAccountHandler).Methods("GET")
+func AddAPIHandlers(s *sebakstorage.LevelDBBackend) func(ctx context.Context, t *sebaknetwork.HTTP2Network) {
+	fn := func(ctx context.Context, t *sebaknetwork.HTTP2Network) {
+		t.AddAPIHandler("/account/{address}", GetAccountHandler(s)).Methods("GET")
+	}
+	return fn
 }
 
 func streaming(o *observable.Observable, w http.ResponseWriter, event string, callBackFunc func(args ...interface{}) ([]byte, error), once []byte) {
--- a/lib/api.go
+++ b/lib/api.go
@@ -1,0 +1,63 @@
+package sebak
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"boscoin.io/sebak/lib/network"
+	"github.com/GianlucaGuarini/go-observable"
+)
+
+func AddAPIHandlers(ctx context.Context, t *sebaknetwork.HTTP2Network) {
+
+	t.AddAPIHandler(ctx, "/account/{address}", GetAccountHandler).Methods("GET")
+}
+
+func streaming(o *observable.Observable, w http.ResponseWriter, event string, callBackFunc func(args ...interface{}) ([]byte, error), once []byte) {
+	cn, ok := w.(http.CloseNotifier)
+	if !ok {
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+		return
+	}
+
+	flusher, ok := w.(http.Flusher)
+	if !ok {
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+		return
+	}
+
+	closeChan := make(chan bool)
+	messageChan := make(chan []byte)
+
+	observerFunc := func(args ...interface{}) {
+		s, err := callBackFunc(args...)
+		if err != nil {
+			closeChan <- true
+			return
+		}
+		messageChan <- s
+	}
+
+	o.On(event, observerFunc)
+	defer o.Off(event, observerFunc)
+
+	w.Header().Set("Content-Type", "application/json")
+
+	if len(once) != 0 {
+		fmt.Fprintf(w, "%s\n", once)
+		flusher.Flush()
+	}
+
+	for {
+		select {
+		case <-cn.CloseNotify():
+			return
+		case <-closeChan:
+			return
+		case message := <-messageChan:
+			fmt.Fprintf(w, "%s\n", message)
+			flusher.Flush()
+		}
+	}
+}
--- a/lib/api_test.go
+++ b/lib/api_test.go
@@ -1,0 +1,95 @@
+package sebak
+
+import (
+	"boscoin.io/sebak/lib/storage"
+	"bufio"
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/gorilla/mux"
+	"net/http"
+	"sync"
+	"testing"
+	"time"
+)
+
+func TestStreaming(t *testing.T) {
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	// Setting Server
+	storageConfig, err := sebakstorage.NewConfigFromString("memory://")
+	if err != nil {
+		t.Errorf("failed to initialize file db: %v", err)
+
+	}
+	storage, err := sebakstorage.NewStorage(storageConfig)
+	if err != nil {
+		t.Errorf("failed to initialize file db: %v", err)
+	}
+	defer storage.Close()
+
+	ctx := context.WithValue(context.Background(), "storage", storage)
+	router := mux.NewRouter()
+	router.HandleFunc("/account/{address}", GetAccountHandler(ctx, nil)).Methods("GET")
+	server := &http.Server{Addr: ":5000", Handler: router}
+	go server.ListenAndServe()
+
+	// Make Dummy BlockAccount
+	ba := testMakeBlockAccount()
+	ba.Save(storage)
+	prev := ba.GetBalance()
+
+	// Do stream Request to the Server
+	go func() {
+		req, err := http.NewRequest("GET", fmt.Sprintf("http://localhost:5000/account/%s", ba.Address), nil)
+		if err != nil {
+
+		}
+		req.Header.Set("Accept", "text/event-stream")
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			t.Error("Server not yet initialized")
+		}
+
+		if err != nil {
+			for {
+				resp, err = http.DefaultClient.Do(req)
+				if err == nil {
+					break
+				}
+				time.Sleep(time.Second)
+			}
+		}
+		reader := bufio.NewReader(resp.Body)
+		var n Amount
+		for n = 1; n < 10; n++ {
+			line, _ := reader.ReadBytes('\n')
+			var cba = &BlockAccount{}
+			json.Unmarshal(line, cba)
+
+			if ba.Address != cba.Address {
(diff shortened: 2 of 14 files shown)


kfangw:
@spikeekips @zzim2x Please review

lib/broker_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	time.Sleep(time.Second)
+
+	w.Flush()
+	w2.Flush()
+	ret1 := buffer.String()
+	ret2 := buffer2.String()
+
+	fmt.Println(ret1)
+	fmt.Println(ret2)
```
outsideris: These are unnecessary.

lib/broker.go
```diff
@@ -0,0 +1,142 @@
...
+import (
+	"io"
+)
+
+type EventType string
+
+const (
+	EventDummy  EventType = "dummy"
+	EventDummy2           = "dummy2"
+)
```
outsideris: Why is this needed? `dummy` looks like test purpose.

lib/broker.go
```diff
@@ -0,0 +1,142 @@
...
+
+type EventBroker struct {
+	brokers map[EventType]*Broker
```
anarcher: IMHO, `EventBroker` seems like  `EventBrokers`.

lib/broker.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func NewEventBroker() *EventBroker {
+	e := &EventBroker{
+		brokers: map[EventType]*Broker{},
+	}
+	e.brokers[EventDummy] = NewBroker()
```
anarcher: IMHO, This statement `e.brokers[EventDummy] = ..` is not necessary.

lib/broker.go
kfangw: Event types are not yet defined. Hence, I added those two.
Maybe those are changed to practical names

lib/broker.go
kfangw: AddBroker much better 👍

lib/broker_test.go
kfangw: yeap. I will remove it.

kfangw:
I am working on another testcase.. the testcase may fail.

spikeekips:
I understand why you invented `EventBroker`. I thought you already deeply considered the other implementation of internal event systems. We can make it easier like https://github.com/GianlucaGuarini/go-observable or https://github.com/ReactiveX/RxGo .

cmcm2222:
Wow. 
Reactive programming can be in golang too?? 
Awesome.

kfangw:
Good to know those projects.
I will adopt one one them or make a similar one. Thanks spike!!

kfangw:
@spikeekips @zzim2x @outsideris @anarcher 
please review again.
At this moment, please ignore broken ci build. The broken caused by CI itself

I changed two point
1. server-sent-event api with go-observerble
2. Adopt mux

After this change, i will add bellows
1. client
2. APIs for each issues

spikeekips:
@kfangw Please ask us after finishing work :) Too often changes, I am confused when I start to review.

kfangw:
Writing is done but there is a CI problem...T^T...
Sorry for the annoying...
I will ask again when the problem is solved

spikeekips:
@kfangw Thanks :)

kfangw:
Lets move to https://github.com/bosnet/sebak/pull/114

kfangw:
CI build failed occurred occasionally.
However, the original HANG problem is not seen anymore.
I am sorry for the annoying until now.

Please review this PR

outsideris:
DId you find what caused the hang? just curious.

kfangw:
circleci ran the build with old source.
When I submit another PR with the same branch, It turns to normal...
I don't know what happened in the CI exactly.

kfangw:
@outsideris 
refer this https://circleci.com/gh/bosnet/sebak/190
At that time, I removed the testcase "TestStreaming"
but it ran test with old source.

I guess that the CI couldn't fetch latest source code. 

I have trying to reproduce the problem, but I can't...
Unfortunately, I squashed the commit and force pushed... so i lost the commit T^T

lib/api.go
```diff
@@ -0,0 +1,127 @@
...
+			return
+		case message := <-messageChan:
+			fmt.Fprintf(w, "%s\n", message)
+			flusher.Flush()
+		}
+	}
+}
+
+func GetAccountHandler(ctx context.Context) HandlerFunc {
+	storage, ok := ctx.Value("storage").(*sebakstorage.LevelDBBackend)
```
anarcher: IMHO,  Using `LevelDBBackend` with `context.Context.Value` is not suitable in normal cases.

lib/api_test.go
```diff
@@ -0,0 +1,95 @@
...
+	storage, err := sebakstorage.NewStorage(storageConfig)
+	if err != nil {
+		t.Errorf("failed to initialize file db: %v", err)
+	}
+	defer storage.Close()
+
+	ctx := context.WithValue(context.Background(), "storage", storage)
+	router := mux.NewRouter()
+	router.HandleFunc("/account/{address}", GetAccountHandler(ctx)).Methods("GET")
+	server := &http.Server{Addr: ":5000", Handler: router}
```
anarcher: IHMO, Using `net/http/httptest` can be better for using testing :-D.

lib/block_account.go
```diff
@@ -42,6 +43,8 @@ func (b *BlockAccount) String() string {
 	return string(sebakcommon.MustJSONMarshal(b))
 }
 
+var BlockAccountObserver = observable.New()
```
anarcher: IHMO, In general, **package level vars** as using  **Package-global objects** can have surprising side effects.  Anyway, This var can be in another place. :)

kfangw: I don't like this usage also. I am totally agreed what you worry about.
However, the object should be accessed from everywhere.
So, I will just move it to another separated package.
Is it enough? or is there any alternative solution? @anarcher

lib/api.go
```diff
@@ -0,0 +1,119 @@
...
+	"boscoin.io/sebak/lib/error"
+	"boscoin.io/sebak/lib/storage"
+	"github.com/GianlucaGuarini/go-observable"
+	"github.com/gorilla/mux"
+	"boscoin.io/sebak/lib/network"
+)
+
+func AddAPIHandlers(ctx context.Context, t *sebaknetwork.HTTP2Network) {
+
+	t.AddAPIHandler(ctx, "/account/{address}", GetAccountHandler).Methods("GET")
```
kfangw: I think that we may need a bunch of features provided by mux package.
hence, we need `*mux.Route` reference here.

@spikeekips

lib/block_account.go
anarcher: @kfangw  I understand why it has to exist in pkg-global var. It makes sense :-)

lib/api_test.go
kfangw: I tried the httptest.
May I change when I add another APIs?

lib/api.go
kfangw: Issue what you picked is spread all around of handlers for sebak.
I think we need to talk about this in another issue.

kfangw:
@spikeekips @anarcher May I merge this PR?
remaining issues pointed in this PR are ... 

1. unittest with httptest
2. usage of context

those will be solved in another PR.

lib/api_test.go
anarcher: Using httptest is optional :-)!

anarcher:
IMHO, unittest with httptest is optional. You can ignore it!! :-D

lib/api.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"boscoin.io/sebak/lib/network"
+	"github.com/GianlucaGuarini/go-observable"
+)
+
```
anarcher: I know my approach is ugly. `T_T`
```
⚠ func (nr *NodeRunner) Ready() {
      nr.network.SetContext(nr.ctx)
      nr.network.AddHandler(nr.ctx, AddAPIHandlers(nr.storage))
      nr.network.Ready()
  }
```

```
⚠ func AddAPIHandlers(s *sebakstorage.LevelDBBackend) func(ctx context.Context, t *sebaknetwork.HTTP2Network) {
      fn := func(ctx context.Context, t *sebaknetwork.HTTP2Network) {
          //t.AddAPIHandler(ctx, "/account/{address}", GetAccountHandler).Methods("GET")
          t.AddAPIHandler("/account/{address}", GetAccountHandler(ctx, t, s)).Methods("GET")
      }
      return fn
  }
```

```
func GetAccountHandler(ctx context.Context, t *sebaknetwork.HTTP2Network, s *sebakstorage.LevelDBBackend) http.Handle
  rFunc {
      /*
          storage, ok := ctx.Value("storage").(*sebakstorage.LevelDBBackend)
          if !ok {
              panic(errors.New("storage is missing in context"))
          }
      */

      storage := s
```

```
⚠ func (t *HTTP2Network) AddAPIHandler(pattern string, handlerFunc http.HandlerFunc) (router *mux.Route) {
      //pattern := args[0].(string)
      //handler := args[1].(func(context.Context, *HTTP2Network) HandlerFunc)
      apiRouter := t.routers["api"]
      return apiRouter.HandleFunc(pattern, handlerFunc)
  }
```

lib/network/http2_network.go
```diff
@@ -61,15 +69,22 @@ func NewHTTP2Network(config HTTP2NetworkConfig) (h2n *HTTP2Network) {
...
 		server:         server,
+		router:         baseRouter,
 		tlsCertFile:    config.TLSCertFile,
 		tlsKeyFile:     config.TLSKeyFile,
 		receiveChannel: make(chan Message),
 	}
+	h2n.handlers = map[string]func(http.ResponseWriter, *http.Request){}
+	h2n.routers = map[string]*mux.Router{
+		"node": baseRouter.PathPrefix("/node").Subrouter(),
+		"api":  baseRouter.PathPrefix("/api").Subrouter(),
```
spikeekips: We can use `UrlPrefixXX` in this place even in map key.

lib/network/http2_network.go
```diff
@@ -10,25 +10,33 @@ import (
...
 	"golang.org/x/net/http2"
 )
 
 type Handlers map[string]func(http.ResponseWriter, *http.Request)
 
+const (
+	UrlNodePrefix = "/node"
+	UrlApiPrefix  = "/api"
+)
+
```
spikeekips: In the below code for `HTTP2Network.routers`, you can define url path prefix and router name too.
```go
const (
	RouterNameNode = "node"
	RouterNameAPI  = "node"
)

var (
	UrlPathPrefixNode = fmt.Sprintf("/%s", RouterNameNode)
	UrlPathPrefixAPI  = fmt.Sprintf("/%s", RouterNameAPI)
)
```

spikeekips:
@kfangw I think this PR is the 1st implementation of API using **server side stream**, so how about change the PR title for the original issue, #3 .

lib/api.go
kfangw: It is not ugly. elegance. 👍 
I have been suffered from circular import, hence my dirty code been written.
Your idea makes me free from circular import and refer to the storage
Thanks @anarcher

lib/network/http2_network.go
kfangw: Yeap. I changed it to named const.

lib/api.go
anarcher: @kfangw  I am honored that I have helped you. `:-]`

PR #64, opened 2018-06-07
#60 Fixed checkpoint validation

spikeekips:
### Github Issue
Please see #60.

Issue #60, opened 2018-06-06
Unreasonable 'Checkpoint` Validation


spikeekips:
When transaction stores, it updates the `Checkpoint` of target account. This causes problem that target account can not know his/her latest checkpoint when he/her tries to make new tranaction.

For example, there are 3 account, `a0`, `a1` and `a2`, each account has 1 BOS, and send payment transactions between each others.

| souce | description       | target  | `a2` balance   | `a2` checkpoint |
| ----- | ----------------- | ------- | -------------- | --------------- |
| `a0`  | `tx0` sends 1 BOS | `a2`    | 1 BOS -> 2 BOS | `chk0`          |
| .... |  |  |  |  |
| `a2`  | `tx2` sends 1 BOS | `a0`    |                |                 |
| .... |  |  |  |  |
| `a1`  | `tx1` sends 1 BOS | `a2`    | 2 BOS -> 3 BOS | `chk1`          |

If `a2` sends `tx2` before `tx1` is stored, but `tx1` is stoed during `tx2` under consensus. Of course, the `Checkpoint` of `tx2` is `chk0`. What happened?

Current implementation will think, `tx2` is invalid, that is, `tx2` does not have latest checkpoint, so `tx2` will be failed to get consensus.

This problem is caused because the validation just checks whether `Checkpoint` of transaction is the latest checkpoint of account, or not.

If the validation checks that the `Checkpoint` exists in transactions of the account and at that `Checkpoint` the balance is sufficient or not, the problem will be solved.

This is new strategy.
### How `Checkpoint` can be made

checkpoint is: `<subtracted>-<added>`
* `<subtracted>`: hash of last paid transaction, it means balance is *subtracted*
* `<added>`: hash of last added transaction, it means balance is *added*

### Validating `Checkpoint`

* if checkpoint is final checkpoint of source, check balance
* if not, try to find existing checkpoint of source
    * if not, say no
    * if found,
        * check, `<subtracted>` is same with `<added>` of final check
            * if not, say no
            * if same, check the balance of that time.

### Checkpoint Changes Example

| account | initial checkpoint |
| --      | --                 |
| a0      | `tx0-tx0`          |
| a1      | `tx1-tx1`          |
| a2      | `tx2-tx2`          |

| source | target | transaction | checkpoint of transaction | checkpoint of source | checkpoint of target |
| --     | --     | --          | --                        | --                   | --                   |
[...]

PR #88, opened 2018-06-16
#60 Fixed checkpoint validation

spikeekips:
This PR is from #64.

### Github Issue
Please see #60.

### DIFF
--- a/lib/block_account.go
+++ b/lib/block_account.go
@@ -38,10 +38,6 @@ func NewBlockAccount(address string, balance Amount, checkpoint string) *BlockAc
 	}
 }
 
-func NewBlockAccountFromTransaction(address string, balance Amount, tx Transaction) *BlockAccount {
-	return NewBlockAccount(address, balance, tx.NextCheckpoint())
-}
-
 func (b *BlockAccount) String() string {
 	return string(sebakcommon.MustJSONMarshal(b))
 }
--- a/lib/block_account.go
+++ b/lib/block_account.go
@@ -21,6 +21,8 @@ import (
 
 const BlockAccountPrefixAddress string = "ba-address-"
 const BlockAccountPrefixCreated string = "ba-created-"
+const BlockAccountCheckpointPrefix string = "bac-ac-"
+const BlockAccountCheckpointByAddressPrefix string = "bac-aa-"
 
 type BlockAccount struct {
 	Address    string
@@ -61,6 +63,9 @@ func (b *BlockAccount) Save(st *sebakstorage.LevelDBBackend) (err error) {
 		createdKey := GetBlockAccountCreatedKey(sebakcommon.GetUniqueIDFromUUID())
 		err = st.New(createdKey, b.Address)
 	}
+
+	bac := NewBlockAccountCheckpoint(b, b.Checkpoint)
+	err = bac.Save(st)
 
 	return
 }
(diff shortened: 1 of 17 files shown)


lib/node_runner_isaac_test.go
```diff
@@ -292,8 +299,29 @@ func TestNodeRunnerConsensusSameSourceWillNotIgnored(t *testing.T) {
 		wg.Done()
 	}
 
+	var secondHandleBallotCheckerFuncs = []sebakcommon.CheckerFunc{
+		CheckNodeRunnerHandleBallotIsWellformed,
+		CheckNodeRunnerHandleBallotCheckIsNew,
+		CheckNodeRunnerHandleBallotReceiveBallot,
+		CheckNodeRunnerHandleBallotHistory,
+		// CheckNodeRunnerHandleBallotStore
+		func(c sebakcommon.Checker, args ...interface{}) (err error) {
```
anarcher: This func seems like just doing return but doing nothing. Is it necessary?

lib/block_account.go
```diff
@@ -21,6 +21,8 @@ import (
 
 const BlockAccountPrefixAddress string = "ba-address-"
 const BlockAccountPrefixCreated string = "ba-created-"
+const BlockAccountCheckpointPrefix string = "bac-ac-"
+const BlockAccountCheckpointByAddressPrefix string = "bac-aa-"
```
outsideris: What is `aa` mean?

kfangw:
It may takes about 1+ day(s) for me to review this PR. .
Could you give me enough time? 
If then, I want to catch up the context.
If not, you can finish this PR without my review.

lib/block_account.go
outsideris: It's not straightforward. I don't have another idea yet. 

* `bac` is `BlockAccountCheckpoint`
* `ac` is `AccountCheckpoint`
* `aa` is `Account..Address`.

spikeekips: @outsideris As you know, the middle part is only for making distinct key, anything can be possible. Please suggest good middle name ㅎㅎ

lib/node_runner_isaac_test.go
anarcher: I guess that  I get what you mean.  :-)  Although the test code works without it(the empty func), it exists for to point to `CheckNodeRunnerHandleBallotStore`. Is it right?

spikeekips: @anarcher Without this empty function, test will not be passed. :)

lib/transaction.go
```diff
@@ -125,6 +126,20 @@ func (o Transaction) Equal(m sebakcommon.Message) bool {
 	return o.H.Hash == m.GetHash()
 }
 
+func (o Transaction) IsValidCheckpoint(checkpoint string) bool {
```
Charleslee522: I figure, this method is the most important one in this pr :)

lib/node_runner_isaac_test.go
anarcher: Really?  I thought this test will be passed without the empty func :-)

spikeekips: @anarcher Oh, yes, you right. You reminds me I'm stupid ㅋㅋㅋ

anarcher: Never! HAHA.  IMO,  The empty function represents that `CheckNodeRunnerHandleBallotStore` is empty. So I would like to keep this empty function. :-D

anarcher:
Needs a rebase or resolve (sorry), because I added new error ~_~

lib/transaction.go
```diff
@@ -161,14 +176,19 @@ func (o *Transaction) Sign(kp keypair.KP, networkID []byte) {
...
-		base58.Encode(
-			sebakcommon.MakeHash(
-				[]byte(fmt.Sprintf("%s%s", o.B.Checkpoint, o.GetHash())),
-			),
-		),
-	)
+// NextSourceCheckpoint generate new checkpoint from current Transaction. It has
+// 2 part, "<head>-<body>".
+//
+// <head>: hash of last paid transaction of source account
```
leejames00: comment should follow https://github.com/bosnet/sebak/issues/60 -- `<subtracted> - <added>`

lib/block_account.go
```diff
@@ -159,3 +160,97 @@ func (b *BlockAccount) Withdraw(fund Amount, checkpoint string) error {
...
+// block. the storage should support,
+//  * find by `Address`:
+// 	- key: "`Address`-`Checkpoint`": value: `ID` of BlockAccountCheckpoint
+//  * get list by created order:
+//
+// models
+//  * 'address' and 'checkpoint'
+// 	- 'bac-<BlockAccountCheckpoint.Address>-<BlockAccountCheckpoint.Checkpoint>': `BlockAccountCheckpoint`
+type BlockAccountCheckpoint struct {
+	Checkpoint string
```
Charleslee522: I think, if checkpoint type is the struct like this,
```
type Checkpoint struct {
  Source string
  Target string
}
```
then it would be good for improving performance. But it is quite big changes so, we can handle it next issue #98.

lib/transaction.go
```diff
@@ -125,6 +126,20 @@ func (o Transaction) Equal(m sebakcommon.Message) bool {
 	return o.H.Hash == m.GetHash()
 }
 
+func (o Transaction) IsValidCheckpoint(checkpoint string) bool {
+	if o.B.Checkpoint == checkpoint {
+		return true
+	}
+
+	a := strings.SplitN(checkpoint, "-", 2)
+	b := strings.SplitN(o.B.Checkpoint, "-", 2)
```
Charleslee522: Why don't you use [`ParseCheckpoint`](https://github.com/bosnet/sebak/pull/88/files#diff-376d399af0732b484ed220968101362cR163) method instead of `strings.SplitN`?

lib/transaction.go
```diff
@@ -177,10 +177,10 @@ func (o *Transaction) Sign(kp keypair.KP, networkID []byte) {
 }
 
 // NextSourceCheckpoint generate new checkpoint from current Transaction. It has
-// 2 part, "<head>-<body>".
+// 2 part, "<subtracted>-<added>".
 //
-// <head>: hash of last paid transaction of source account
-// <body>: hash of last transaction
+// <subtracted>: hash of last paid transaction, it means balance is subtracted
+// <added>: hash of last added transaction, it means balance is added
```
Geod24: Suggestion: We could use "deposit" and "withdrawal" instead. That would be in line with the name introduced in `account.go` and those terms are used commonly in association with bank account.

lib/block_account.go
spikeekips: @Charleslee522 We can make it in new issue :)

lib/transaction.go
spikeekips: @Geod24 I like new terms. :)

spikeekips: @Geod24 Anyway, if we limit the feature of impl. to the payment, it will be good, but if the transaction will be not used only for payment, bank related terms is not suitable, because current transaction structure can contain the other type of operations.

Geod24: What kind of other operations do you envision ? And if we do, won't we run into the same problem as we do now ?

leejames00: maybe voting?

lib/node_runner_checker.go
```diff
@@ -236,58 +235,90 @@ func CheckNodeRunnerHandleBallotIsBroadcastable(c sebakcommon.Checker, args ...i
...
+	}
+
+	totalAmount := tx.TotalAmount(true)
+	// check, have enough balance at checkpoint
+	if MustAmountFromString(bac.Balance) < totalAmount {
+		return
+	}
+
+	// check, have enough balance now
+	if totalAmount > MustAmountFromString(ba.Balance) {
```
kfangw: To make it pretty.
MustAmountFromString(ba.Balance) < totalAmount

above condition is like this
MustAmountFromString(bac.Balance) < totalAmount

lib/block_account.go
```diff
@@ -21,6 +21,8 @@ import (
 
 const BlockAccountPrefixAddress string = "ba-address-"
 const BlockAccountPrefixCreated string = "ba-created-"
+const BlockAccountCheckpointPrefix string = "bac-ac-"
```
kfangw: why do we need those prefix?
To avoid hash conflict?

spikeekips: @kfangw Yes, right.

lib/node_runner_checker.go
spikeekips: @kfangw No problema!

PR #94, opened 2018-06-18
Set the log level on the node runner too

Geod24:
This way when we set the debugging information, it actually applies to the core of the business logic and we can debug the consensus protocol.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -39,9 +39,5 @@
   version = "1.3.0"
 
 [[constraint]]
-  name = "github.com/mattn/go-isatty"
-  version = "0.0.3"
-
-[[constraint]]
   name = "github.com/spf13/cobra"
   version = "0.0.3"
--- a/cmd/sebak/cmd/node.go
+++ b/cmd/sebak/cmd/node.go
@@ -10,7 +10,6 @@ import (
 	"golang.org/x/net/http2"
 
 	logging "github.com/inconshreveable/log15"
-	"github.com/mattn/go-isatty"
 	"github.com/spf13/cobra"
 	"github.com/stellar/go/keypair"
 
@@ -195,15 +194,7 @@ func parseFlagsNode() {
 		common.PrintFlagsError(nodeCmd, "--log-level", err)
 	}
 
-	var logHandler logging.Handler
-
-	var formatter logging.Format
-	if isatty.IsTerminal(os.Stdout.Fd()) {
-		formatter = logging.TerminalFormat()
-	} else {
-		formatter = logging.JsonFormatEx(false, true)
-	}
-	logHandler = logging.StreamHandler(os.Stdout, formatter)
+	logHandler := logging.StdoutHandler
 
 	if len(flagLogOutput) < 1 {
 		flagLogOutput = "<stdout>"


cmcm2222:
LGTM

spikeekips:
Can you explain more details? [`nr.log`](https://github.com/Geod24/sebak/blob/abb0b6a41cce62a170f534c4ca2d6afff3e36828/lib/node_runner.go#L47) already set the same `logLevel` and same `logHandler` with the main module.

Geod24:
Without this PR, all I get during a normal run is:
```
127.0.0.1 - - [18/Jun/2018:09:33:22 +0000] "POST /connect HTTP/2.0" 200 126 "" "v-GDIR.M6CC"
```

which comes, I believe, from the [network layer](https://github.com/bosnet/sebak/blob/9bc1d56f468635c66f724459c2584d0f2e33d7f7/cmd/sebak/cmd/node.go#L219).
With this PR, I get:
```
127.0.0.1 - - [18/Jun/2018:09:33:22 +0000] "POST /connect HTTP/2.0" 200 126 "" "v-GDIR.M6CC"
DBUG[06-18|09:33:22] got connect                              module=sebak node=GAYG.TLY7 message="{\"Type\":\"connect\",\"Data\":\"eyJhZGRyZXNzIjoiR0RJUkY0"
```

Geod24:
How do you run it ? I use the `docker` command.

Geod24:
Okay, I see the issue now. Why is [this check](https://github.com/bosnet/sebak/blob/fb78d4ef02551230baf3294235a4e30c864cbdba/cmd/sebak/cmd/node.go#L201-L205) performed ?
When you run `sebak` in `docker`, it evaluates to `false`, leading to the situation I described.

spikeekips:
> @Geod24: Okay, I see the issue now. Why is [this check](https://github.com/bosnet/sebak/blob/fb78d4ef02551230baf3294235a4e30c864cbdba/cmd/sebak/cmd/node.go#L201-L205) performed ?
When you run `sebak` in `docker`, it evaluates to `false`, leading to the situation I described.

 As you know, 'log15' can colorize log message by log level, so when not in terminal, the escape sequence strings will be shown. `isatty.IsTerminal(os.Stdout.Fd())` checks the stdout is attached to the tty or not.

Geod24:
> As you know, 'log15' can colorize log message by log level, so when not in terminal, the escape sequence strings will be shown.

I run `docker` with `-it`, and `-t` is short for `--tty`, so a TTY does get allocated but somehow is not recognized.

Geod24:
@spikeekips : What do you think about getting rid of the check for the time being ? I think having logging in `docker` is more valuable than having files without control chars. And we can always do `less -R file.txt` to interpret those!

spikeekips:
> @Geod24 : @spikeekips : What do you think about getting rid of the check for the time being ? I think having logging in `docker` is more valuable than having files without control chars. And we can always do `less -R file.txt` to interpret those!

Yeah, I agree. We can provide new option for turn off colorizing instead of checking tty or not.

Geod24:
@spikeekips : Actually, log15 provides a builtin way to do this, so I used this instead, and AFAICS it works on docker :)

Geod24:
PR update, ready for review

PR #96, opened 2018-06-18
Fix #85 `sebak generate` triggers `value out of range` error

### LINKED ISSUE #85: `sebak generate` triggers `value out of range` error
```
$ sebak genesis SBXBRFM4UDBHREM2XRM6IIOXNR52N6NAKWIMR7MR4XMNJ5VA4WC27QDY
error: invalid '--balance'; strconv.ParseInt: parsing "10000000000000000000": value out of range
```

`sebak genesis` with the default balance, `1,000,000,000,000.0000000` occurs error, it is caused by `sebak.AmountFromString`. `strconv.ParseInt` should be changed to use `ParseUint`.

zzim2x:
initial balance is out of range of a 64 bit integer value.

### DIFF
--- a/lib/data.go
+++ b/lib/data.go
@@ -97,7 +97,7 @@ func (a Amount) MustSub(sub Amount) Amount {
 
 // Implement JSON's Marshaler interface
 func (a Amount) MarshalJSON() ([]byte, error) {
-	return []byte(fmt.Sprintf("\"%s\"", strconv.FormatUint(uint64(a), 10))), nil
+	return []byte(fmt.Sprintf("\"%s\"", a.String())), nil
 }
 
 // Implement JSON's Unmarshaler interface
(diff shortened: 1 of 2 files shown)


Geod24:
Needs a rebase (sorry), since I deleted some of the code.

lib/data.go
```diff
@@ -97,7 +97,7 @@ func (a Amount) MustSub(sub Amount) Amount {
 
 // Implement JSON's Marshaler interface
 func (a Amount) MarshalJSON() ([]byte, error) {
-	return []byte(fmt.Sprintf("\"%s\"", a.String())), nil
+	return []byte(fmt.Sprintf("\"%s\"", strconv.FormatUint(uint64(a), 10))), nil
```
Geod24: I think that should be reverted ;)

zzim2x: I got it.

PR #105, opened 2018-06-19
CircleCI run test with go 1.10 and 1.9

outsideris:
CircleCI test config updated:

- Use go 1.10 as default like our local environments
- Run test on go 1.9 because [our minimum support version is 1.9](https://github.com/bosnet/sebak/wiki/SEBAK-Installation-Guide#prerequisite) via [CircleCI workflow](https://circleci.com/docs/2.0/workflows/).
- Specify timeout for `go test`. Default timeout is 10mins. It's too long.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,20 +2,10 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/golang:1.9
-
+      - image: circleci/golang:1.10
     working_directory: /go/src/boscoin.io/sebak
     steps:
       - checkout
-      - run:
-          name: Checkout merge commit
-          command: |
-            if [[ -n "${CIRCLE_PR_NUMBER}" ]]
-            then
-              FETCH_REFS="${FETCH_REFS} +refs/pull/${CIRCLE_PR_NUMBER}/merge:pr/${CIRCLE_PR_NUMBER}/merge"
-              git fetch -u origin ${FETCH_REFS}
-              git checkout "pr/${CIRCLE_PR_NUMBER}/merge"
-            fi
       - run:
           name: install dep
           command: curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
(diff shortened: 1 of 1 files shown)


outsideris:
Wait! I'm trying to fix circleci config.
Let you know if I fix them.

outsideris:
Additionally:

- separate merged test into different job after all test are succeed.
- run fmt first and then run test.

.circleci/config.yml
```diff
@@ -28,9 +18,79 @@ jobs:
...
+      - test_go1_10:
+          requires:
+            - fmt
+      - test_go1_9:
+          requires:
+            - fmt
+      - test_merged:
+          requires:
+            - test_go1_10
+            - test_go1_9
```
Geod24: Wait, if I understand correctly we are:
- Running the test on the HEAD of the PR with Go 1.9
- Running the test on the HEAD of the PR with Go 1.10
- Running the test on the merge commit of the PR with Go 1.10

I would expect:
- Running the test on the merge commit of the PR with Go 1.9
- Running the test on the merge commit of the PR with Go 1.10

.circleci/config.yml
```diff
@@ -28,9 +18,79 @@ jobs:
           key: dependency-cache-{{ checksum "Gopkg.lock" }}
           paths:
             - vendor
+      - persist_to_workspace:
+          root: /go/src/boscoin.io
+          paths:
+            - sebak
```
Geod24: Why ?

.circleci/config.yml
outsideris: There are two code bases.
1. HEAD of the PR
2. merged commit with master and the PR

I think the PR should be tested in both cases. (And Travis ci do that).

Sometimes, we experience a case of that tests are passed in local, but CI is failed. Under that situation, we should know if the PR itself is failed or it is failed after merging with the master.

If we test only 2nd case, it is hard to find the problem. That is why I test both cases on CI.

Geod24: Yeah but do we really care about `HEAD` of the PR ?
I don't see any scenario where it really matters for reviewing.

> (And Travis ci do that).

According to [the documentation](https://docs.travis-ci.com/user/pull-requests/#How-Pull-Requests-are-Built) this is not the case:

> Rather than build the commits that have been pushed to the branch the pull request is from, we build the merge between the source branch and the upstream branch.


> If we test only 2nd case, it is hard to find the problem. That is why I test both cases on CI.

My opinion here is that if you see a CI failure when it passes locally, the first step to take is to rebase your PR. Then there's no difference anymore.

outsideris: As [the documentaiton](https://docs.travis-ci.com/user/pull-requests/#%E2%80%98Double-builds%E2%80%99-on-pull-requests)

> If you see two build status icons on your GitHub pull request, it means there is one build for the branch, and one build for the pull request itself (actually the build for the merge of the head branch with the base branch specified in the pull request).

I'm not sure it is default or not. But many projects which I'm using has two builds.

I understand what you said. I just think if we run tests for both cases, we can find the problem easily. Otherwise, when CI is failed, but local tests are passed, we didn't know it is failed because of Linux or merging.

So, I should try to fetch upstream and rebase first. And then, if there is no problem, I try to it on Linux or docker. I think it is wasted time.

Anyway, I didn't think about testing on a merge commit much. Give me time. I'm considering it.

Issue #111, opened 2018-06-19
CI randomly failing with cache


outsideris:
* [Failed job](https://circleci.com/gh/bosnet/sebak/149)
* [Passed job without cache](https://circleci.com/gh/bosnet/sebak/154)

I don't know why.  However, it's annoying.

It can be CI's cache problem.

outsideris:
It looks like timing issue, not cache.

outsideris:
To find this problem, I also will add TravisCI as @spikeekips suggestion.

outsideris:
We didn't find the cause and there is wasted time a lot.

If you encounter some random fail in CI, please paste the error message below as comments.

kfangw:
https://circleci.com/gh/bosnet/sebak/491
```
=== RUN   TestNodeRunnerCreateAccount
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x7faaa5]

goroutine 21313 [running]:
boscoin.io/sebak/lib.VotingStateStaging.IsClosed(...)
	/go/src/boscoin.io/sebak/lib/voting.go:309
boscoin.io/sebak/lib.(*VotingResult).IsClosed(...)
	/go/src/boscoin.io/sebak/lib/voting.go:123
boscoin.io/sebak/lib.(*ISAAC).receiveBallotVotingStates(0xc4205ee640, 0xc429f702f8, 0x6, 0xc4204dcff0, 0x2c, 0xc420648780, 0x58, 0xc4204dd020, 0x2c, 0xc429f7a400, ...)
	/go/src/boscoin.io/sebak/lib/isaac.go:173 +0x1b5
boscoin.io/sebak/lib.(*ISAAC).ReceiveBallot(0xc4205ee640, 0xc429f702f8, 0x6, 0xc4204dcff0, 0x2c, 0xc420648780, 0x58, 0xc4204dd020, 0x2c, 0xc429f7a400, ...)
	/go/src/boscoin.io/sebak/lib/isaac.go:85 +0x145
boscoin.io/sebak/lib.CheckNodeRunnerHandleBallotReceiveBallot(0xbb4f00, 0xc4327d2000, 0x0, 0x0, 0x0, 0x0, 0x0)
	/go/src/boscoin.io/sebak/lib/node_runner_checker.go:167 +0xec
[...]

PR #117, opened 2018-06-20
Logs in tests are enabled/disabled now

outsideris:
### Github Issue
Related #111 

### Background
Logs are too verbose during running tests.
To digging #111, this feature is needed

### Solution
`SEBAK_LOG_HANDLER` env var can enable or disable logger.

This PR is made by pair(triple?) programming with @anarcher and @zzim2x


lib/common/test/log.go
```diff
@@ -0,0 +1,25 @@
...
+	logging "github.com/inconshreveable/log15"
+)
+
+func LogHandler() logging.Handler {
+	handlers := map[string]func() logging.Handler{
+		"null": func() logging.Handler {
+			return logging.DiscardHandler()
+		},
+		"stdout": func() logging.Handler {
+			return logging.StreamHandler(os.Stdout, logging.TerminalFormat())
```
Geod24: Use `StdoutHandler` instead

anarcher: @Geod24 @outsideris 

```
  var  StdoutHandler = StreamHandler(os.Stdout, LogfmtFormat())
```
https://godoc.org/github.com/inconshreveable/log15#pkg-variables

I think `TerminalFormat` is better than `LogfmtFormat` in console :-D

Geod24: It is actually a `TerminalFormat` with support for color, but it's [initialized at runtime](https://github.com/inconshreveable/log15/blob/d4e174bf71609b5afbd4a559f7477f370450053d/root.go#L17-L20).
I've replaced similar code with `StdoutHandler` in #94

outsideris:
@Geod24 I fixed it.

astonshnam:
I reviewed it now because I did a different job. This looks very good to me.

Issue #59, opened 2018-06-05
To improve performance, keeping sources in VotingBox


Charleslee522:
`SourceManager` manages a source list which has unique sources in the boxes.
The type of source list should be `map[/* Source */ string]bool` for checking uniqueness.
When a new ballot came, the source from the ballot is inserted to the list. Then, the ballot process finished, it deleted from the list.
While the ballot is in the list, new ballots from the same source are not processed.

It is related with #13(https://github.com/bosnet/sebak/pull/49/files#diff-18683a7542195386fa853b4683d31b7bR38)

PR #119, opened 2018-06-21
#59 `BallotBoxes` has `Sources` to filter the transactions from same source

spikeekips:
### Github Issue
#59 

### Background
Please check #59 

### Solution
To find the ballot, which has same source with incoming transaction, `ISAAC.Boxes` has sources map, so we can easily lookup by source.


lib/ballot.go
```diff
@@ -400,6 +413,15 @@ func (b *BallotBoxes) AddBallot(ballot Ballot) (isNew bool, err error) {
 	return
 }
 
+func (b *BallotBoxes) AddSource(m sebakcommon.Message) {
+	b.Sources[m.Source()] = m.GetHash()
```
Charleslee522: @spikeekips I have a question.
`b.Sources` save only one recently received message from the `source`, right?

lib/isaac.go
```diff
@@ -130,6 +130,7 @@ func (is *ISAAC) receiveBallotStateINIT(ballot Ballot) (vs VotingStateStaging, e
 		is.Boxes.WaitingBox.RemoveVotingResult(vr) // TODO detect error
 		if !vs.IsClosed() {
 			is.Boxes.VotingBox.AddVotingResult(vr) // TODO detect error
+			is.Boxes.AddSource(ballot)
```
Charleslee522: Calling `AddSource` only here?
I figure it is also called in https://github.com/bosnet/sebak/pull/119/files#diff-2231e829007317368e86620b46b6bfcaR343. Did I miss something?

astonshnam: @Charleslee522 I think your opinion is right.

spikeekips:
@leejames00 and @enessech Please complete your review :)

Charleslee522:
@spikeekips If you write `resolves #59`, the issue #59 is automatically closed after being merged.
> ### Github Issue
>

PR #130, opened 2018-06-26
random port for testing

kfangw:
We have been struggling with CI failed.
I guess that It is caused by http server testing.

When the CI runs builds in parallel, a build may effect another build.

What do you think about this?

### DIFF
--- a/lib/network/http2_network_test.go
+++ b/lib/network/http2_network_test.go
@@ -26,7 +26,7 @@ func getPort() string {
 		for {
 			s := rand.NewSource(int64(time.Now().Nanosecond()))
 			r := rand.New(s)
-			testPort = strconv.Itoa(r.Intn(65535))
+			testPort = strconv.Itoa(r.Intn(16383) + 49152) // ephemeral ports range 49152 ~ 65535
 
 			ln, err := net.Listen("tcp", ":"+testPort)
 			if err == nil {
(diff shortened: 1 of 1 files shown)


lib/network/http2_network_test.go
```diff
@@ -11,9 +11,35 @@ import (
...
+var once sync.Once
+
+func getPort() string {
+	once.Do(func() {
+		for {
+			s := rand.NewSource(int64(time.Now().Nanosecond()))
+			r := rand.New(s)
+			testPort = strconv.Itoa(r.Intn(65535))
+
+			ln, err := net.Listen("tcp", ":"+testPort)
```
Charleslee522: I think, this checking part is good :)

spikeekips: In terms of information, the range of port can be changed. As we know, 0 ~ 1023 is well-known ports and from 1024 to 49151 is also registered ports(https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Well-known_ports), so we can use the private port range to escape the conflict.(https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Dynamic,_private_or_ephemeral_ports)

kfangw: use ephemeral ports
range 49152 ~ 65535
@spikeekips

Issue #26, opened 2018-06-01
Graceful Stop


spikeekips:
Useful library: https://github.com/facebookgo/grace

spikeekips:
@anarcher At first, the reason for 'graceful restart' was that the node needed to be restarted to update new version or new validator list, or anything else to be applied new environments.

anarcher:
> I'm wondering that too. I don't see much value in implementing graceful restart in app. Graceful stop is a nice feature though.
@Geod24 I agree with your opinion! :-) 

> @anarcher At first, the reason for 'graceful restart' was that the node needed to be restarted to update new version or new validator list, or anything else to be applied new environments.

IHMO, 'graceful restart' **without dropping any connections** (facebookgo/grace) is not suitable for node. 

For changing validator list, we should stop and start a node with the new validator arguments.
At some time, If old node process and new node process tries to access same leveldb, one of them get a failure to connect. 

I feel safe to stop and start a node but, with graceful stop :-)
So, I would like to change this issues title 'Graceful Stop' :-)

spikeekips:
@anarcher It will be unnecessary for receiving message from client, because it does not need to keep connection, and the other cases like 'event-stream' can be ignored. As you suggested, 'graceful stop' will be good.

PR #131, opened 2018-06-26
Add `oklog.run` for graceful stop

anarcher:
### Github Issue

#26 

### Solution

Using `oklog/run` for a universal mechanism to manage goroutine lifecycles. 

`oklog/run` : https://github.com/oklog/run

### Possible Drawbacks

### DIFF
--- a/cmd/sebak/cmd/node.go
+++ b/cmd/sebak/cmd/node.go
@@ -10,6 +10,7 @@ import (
 	"golang.org/x/net/http2"
 
 	logging "github.com/inconshreveable/log15"
+	"github.com/oklog/run"
 	"github.com/spf13/cobra"
 	"github.com/stellar/go/keypair"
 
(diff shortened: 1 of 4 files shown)


cmd/sebak/cmd/node.go
```diff
@@ -280,10 +281,32 @@ func runNode() {
 
 		os.Exit(1)
 	}
-	nr := sebak.NewNodeRunner(flagNetworkID, currentNode, policy, nt, isaac, st)
-	if err := nr.Start(); err != nil {
-		log.Crit("failed to start node", "error", err)
 
+	// Execution group.
+	var g run.Group
+	{
```
Charleslee522: Just curious, why this brace needs?

spikeekips: @Charleslee522 If you cover your code with curly braces, the newly assigned variables in the block will be local variable, that is, it will not be available from outside of the block. Without this feature, it can divide the code block **VISUALLY** :)

spikeekips:
@anarcher Please merge this PR if there are no further work.

cmd/sebak/cmd/node.go
Geod24: Also known as [lexical scoping](https://whatis.techtarget.com/definition/lexical-scoping-static-scoping) ;)

PR #136, opened 2018-06-27
marshal node

kfangw:
To get validators address
`func (v *Validator) MarshalJSON()` should be modified like this.

### Github Issue
#9

### DIFF
--- a/lib/common/node.go
+++ b/lib/common/node.go
@@ -134,10 +134,15 @@ func (v *Validator) RemoveValidators(validators ...*Validator) error {
 }
 
 func (v *Validator) MarshalJSON() ([]byte, error) {
+	var neighbors = make(map[string]struct{})
+	for _, neighbor := range v.validators {
+		neighbors[neighbor.Address()] = struct{}{}
+	}
 	return json.Marshal(map[string]interface{}{
 		"address":  v.Address(),
 		"alias":    v.Alias(),
 		"endpoint": v.Endpoint().String(),
+		"validators": neighbors,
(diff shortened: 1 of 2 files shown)


kfangw:
@spikeekips please review this PR
it is related to #9

lib/common/node.go
```diff
@@ -134,10 +134,15 @@ func (v *Validator) RemoveValidators(validators ...*Validator) error {
...
+	}
 	return json.Marshal(map[string]interface{}{
-		"address":  v.Address(),
-		"alias":    v.Alias(),
-		"endpoint": v.Endpoint().String(),
+		"address":    v.Address(),
+		"alias":      v.Alias(),
+		"endpoint":   v.Endpoint().String(),
+		"validators": neighbors,
 		//"validators": v.validators,
```
Charleslee522: 👍 Looks good to me!
There are no tests about marshaling, so checking these code is a bit difficult.
I added related tests #134 already, so we could be able to check it after merging this and #134.

ps. comment seems to be removed.

spikeekips:
@kfangw This PR tried to prevent marshaling **loop**?

kfangw:
@spikeekips exactly.

spikeekips:
**Marshal**(@kfangw) prevents **Marshal**ing loop ㅎㅎㅎ

kfangw:
I have one more `l`. `Marshall`  :)
This needed to be hold until #134 
Hence, I will make a #9 without validators.

PR #143, opened 2018-06-28
#9 NodeInfo API

kfangw:
### Github Issue
#9 

/api/node

### DIFF
--- a/lib/api_node.go
+++ b/lib/api_node.go
@@ -1,0 +1,55 @@
+package sebak
+
+import (
+	"boscoin.io/sebak/lib/common"
+	"boscoin.io/sebak/lib/error"
+	"boscoin.io/sebak/lib/observer"
+	"context"
+	"fmt"
+	"net/http"
+)
+
+const GetNodePattern = "/node"
+
+func GetNodeHandler(ctx context.Context) http.HandlerFunc {
+
+	return func(w http.ResponseWriter, r *http.Request) {
+
+		var err error
+
+		cn := ctx.Value("currentNode").(sebakcommon.Node)
+
+		switch r.Header.Get("Accept") {
+		case "text/event-stream":
+			var readyChan = make(chan struct{})
+			iterateId := sebakcommon.GetUniqueIDFromUUID()
+			go func() {
+				<-readyChan
+				observer.NodeObserver.Trigger(fmt.Sprintf("iterate-%s", iterateId), cn)
+			}()
+
+			event := "change"
+			event += " " + fmt.Sprintf("iterate-%s", iterateId)
+			callBackFunc := func(args ...interface{}) (btSerialized []byte, err error) {
+				cn := args[1].(sebakcommon.Node)
+				var cnSerialized []byte
+				if cnSerialized, err = cn.Serialize(); err != nil {
+					return []byte{}, sebakerror.ErrorSerialized
+				}
+				return cnSerialized, nil
+			}
+			streaming(observer.NodeObserver, w, event, callBackFunc, readyChan)
+		default:
+			var cnSerialized []byte
+			if cnSerialized, err = cn.Serialize(); err != nil {
+				http.Error(w, "Error reading request body", http.StatusInternalServerError)
+				return
+			}
+
+			if _, err = w.Write(cnSerialized); err != nil {
+				http.Error(w, "Error reading request body", http.StatusInternalServerError)
+				return
+			}
+		}
+	}
+}
--- a/lib/api_test.go
+++ b/lib/api_test.go
@@ -471,20 +471,33 @@ func TestNodeHandler(t *testing.T) {
 	resp, _ := http.DefaultClient.Do(req)
 	reader := bufio.NewReader(resp.Body)
 
+	var dummyValidatorIDs [10]string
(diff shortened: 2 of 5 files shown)


kfangw:
@spikeekips @Charleslee522 Please review.

kfangw:
@Charleslee522 please give me your comment here.
#134 and this PR share conflict code.
What I want to change by this PR is that prevent circular (un)marshalling validator object.
If this issue can be solved by #134, then i postpone this PR until #134  merged.

lib/common/node.go
```diff
@@ -133,29 +142,40 @@ func (v *Validator) RemoveValidators(validators ...*Validator) error {
...
+	return json.Marshal(ValidatorFromJSON{
+		v.Alias(),
+		v.Address(),
+		v.Endpoint().String(),
+		neighbors,
 	})
 }
 
-func (v *Validator) UnmarshalJSON(b []byte) error {
+func (v *Validator) UnmarshalJSON(b []byte) (err error) {
```
Charleslee522: I figure it's better because `err` variable already declared in return statement(`err error`) when you want to use return variable err
```
func (v *Validator) UnmarshalJSON(b []byte) (err error) {
	var va ValidatorFromJSON
+	if err = json.Unmarshal(b, &va); err != nil {
+		return
	}

	v.alias = va.Alias
	v.address = va.Address
	v.endpoint, err = NewEndpointFromString(va.Endpoint)
	if err != nil {
+		return
	}
	if v.validators == nil {
		v.validators = make(map[string]*Validator)
	}
	for _, neighbor := range va.Validators {
		v.validators[neighbor] = &Validator{}
	}

+	return
}
```

lib/api_node.go
```diff
@@ -0,0 +1,55 @@
...
+		case "text/event-stream":
+			var readyChan = make(chan struct{})
+			iterateId := sebakcommon.GetUniqueIDFromUUID()
+			go func() {
+				<-readyChan
+				observer.NodeObserver.Trigger(fmt.Sprintf("iterate-%s", iterateId), cn)
+			}()
+
+			event := "change"
+			event += " " + fmt.Sprintf("iterate-%s", iterateId)
```
Charleslee522: How about this?
```
event := fmt.Sprintf("change iterate-%s", iterateId)
```

lib/api_test.go
```diff
@@ -439,5 +441,85 @@ func TestGetTransactionsHandler(t *testing.T) {
...
+				t.Errorf("%s validator is not exist", dummyValidatorIDs[i])
+			}
+		}
+		resp.Body.Close()
+		wg.Done()
+	}()
+	for i := 0; i < 10; i++ {
+		var newValidator *sebakcommon.Validator
+		endpoint, _ := sebakcommon.NewEndpointFromString(sebakcommon.GetUniqueIDFromUUID())
+		newValidator, _ = sebakcommon.NewValidator(dummyValidatorIDs[i], endpoint, "")
```
Charleslee522: How about this without ```var newValidator```?
```newValidator, _ := sebakcommon.NewValidator(dummyValidatorIDs[i], endpoint, "")```

File: lib/api_test.go [hunk 4]
```diff
@@ -439,5 +441,85 @@ func TestGetTransactionsHandler(t *testing.T) {
...
+
+	var dummyValidatorIDs [10]string
+	for i := 0; i < 10; i++ {
+		key, _ := keypair.Random()
+		dummyValidatorIDs[i] = key.Address()
+	}
 
+	// Do stream Request to the Server
+	go func() {
+		for i := 0; i < 5; i++ {
```
Charleslee522: Isn't this 10? not 5.
How about using constant to readability? like
```
const nValiators = 10
```

lib/api_test.go
```diff
@@ -439,5 +441,85 @@ func TestGetTransactionsHandler(t *testing.T) {
...
+	}()
+	for i := 0; i < 10; i++ {
+		var newValidator *sebakcommon.Validator
+		endpoint, _ := sebakcommon.NewEndpointFromString(sebakcommon.GetUniqueIDFromUUID())
+		newValidator, _ = sebakcommon.NewValidator(dummyValidatorIDs[i], endpoint, "")
+		validator.AddValidators(newValidator)
+	}
+
+	wg.Wait()
+	// No streaming
```
Charleslee522: I think, it would be better to split `No streaming` to the other test case for readability.
Like below :)
```
func TestNodeHandlerWithStreamRequest(t *testing.T) {
and
func TestNodeHandlerWithNoStreaming(t *testing.T) {
```

kfangw:
the issue what I concerned was solved by #134 ? @Charleslee522

Charleslee522:
@kfangw Yes i think it was solved by #134. Because there are no validators in Validator struct. Only LocalNode struct have validators so, it can’t be possible to reference validators recursively.

Please check this to make it clear. ‘TestNodeMarshalJSONWithValidator‘

Geod24:
What's the status of this ?

kfangw:
implementation is almost finished. but it will be resumed with refactoring API structure.

spikeekips:
@kfangw The PR is already far behind from latest master :) Please add `/api/node` :)

PR #154, opened 2018-06-29
Add .env file

zzim2x:
This file is automatically generated when a container is initialized.

SEBAK doesn't support config file, so environment variables are put dynamically in that file.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -8,6 +8,9 @@ if [ $# -gt 1 ]; then
     ./sebak $@
 else
     # Node mode
+    if [ -f ".env" ]; then
+      source ./.env
+    fi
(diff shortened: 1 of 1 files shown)


Geod24:
Isn't that duplicating functionalities offered by Docker `--env-file` ? Except with a slightly different syntax (sh vs KVP)

zzim2x:
@Geod24 k8s does not support env-file and it is static file.

Geod24:
@zzim2x : So it's automatically generated by Kubernetes ? From the wording of the commit message, I thought that was a Docker thing. LGTM then (but if you can add it to the commit message, even better !).

zzim2x:
@Geod24 sorry! 

There is that functionality like multi-stage of docker. A initial container generates .env file before node started.

PR #165, opened 2018-07-02
FIX streaming recent data

kfangw:
To send recent data, iterate's order needed to be `reverse`

### DIFF
--- a/lib/api_account.go
+++ b/lib/api_account.go
@@ -94,17 +94,23 @@ func GetAccountTransactionsHandler(storage *sebakstorage.LevelDBBackend) http.Ha
 			iterateId := sebakcommon.GetUniqueIDFromUUID()
 			go func() {
 				<-readyChan
+				var iterList []BlockTransaction
 				count := maxNumberOfExistingData
-				iterFunc, closeFunc := GetBlockTransactionsByAccount(storage, address, false)
+				iterFunc, closeFunc := GetBlockTransactionsByAccount(storage, address, true)
 				for {
 					bt, hasNext := iterFunc()
 					count--
 					if !hasNext || count < 0 {
 						break
 					}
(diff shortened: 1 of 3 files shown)


kfangw:
@spikeekips review please~

spikeekips:
넹넹. I got it.

kfangw:
@spikeekips ping ~

lib/api_account.go
```diff
@@ -94,17 +94,23 @@ func GetAccountTransactionsHandler(storage *sebakstorage.LevelDBBackend) http.Ha
...
 						break
 					}
-					observer.BlockTransactionObserver.Trigger(fmt.Sprintf("iterate-%s", iterateId), &bt)
+					iterList = append(iterList, bt)
 				}
 				closeFunc()
+
+				for idx := range iterList {
+					bt := iterList[len(iterList)-1-idx]
+					observer.BlockTransactionObserver.Trigger(fmt.Sprintf("iterate-%s", iterateId), &bt)
```
Geod24: Wouldn't it be better to trigger as we iterate, instead of duplicating the list and then iterate on it to trigger ?

kfangw: Iterate is reversed. From the latest to 10 item before. Hence we need to reverse it again.
If there is better idea then let me know

kfangw:
If there are 1,2,3...100 items
Reverse iterator iterate from 100, 99, 98 ... 1
So the reason why i make a new list is to iterate reverse again from 91, 92, ... 100

Geod24:
Oh I see. TBH that sounds like a pretty wasteful usage of resources, but I guess as a temporary solution it should be fine.

kfangw:
Yeap. I agreed. Because i have lack of knowledge of levelDB, it is only possible solution now. I will enhance this later. 
Thanks for your concerns.

Geod24:
Could you rebase on master ?

PR #167, opened 2018-07-02
remove not neccessary check

kfangw:
### Requirements

It is not necessary to validate source/target of the transaction which indicated by the checkpoint.

By definition of `checkpoint`(#60), only thing needed to be validate is `<sub>` part.
That validation is done at https://github.com/bosnet/sebak/blob/167c97afe15193a58806f1462ece06ba17d016e4/lib/node_runner_checker.go#L301-L304

### DIFF
--- a/lib/block_transaction.go
+++ b/lib/block_transaction.go
@@ -38,7 +38,6 @@ type BlockTransaction struct {
 	Source             string
 	Fee                Amount
 	Operations         []string
-	Targets            []string
 	Amount             Amount
 
 	Confirmed string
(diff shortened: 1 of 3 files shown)


kfangw:
@spikeekips 
existing `<sub>` of Sender's checkpoint can validate what you concerned.

spikeekips:
> @kfangw : existing `<sub>` of Sender's checkpoint can validate what you concerned.

Yes, you rigtht. `IsValidCheckpoint()` checks the checkpoint points the latest from source, but it does not prove that `tx.Checkpoint` really exists in `BlockTransaction`. One more I want check is that source account had the ability to pay at that checkpoint.

kfangw:
If then only existent of the related `BlockTransaction` needed to be validated.

Amount availability is validated by bellow code. 
https://github.com/bosnet/sebak/blob/167c97afe15193a58806f1462ece06ba17d016e4/lib/node_runner_checker.go#L306-L317

spikeekips:
@kfangw Oh~~ sorry. ㅋㅋㅋ

spikeekips:
:+1

kfangw:
@Charleslee522 Could you make a testcase for the CheckPoint?
I think that we need to have confidence that invalid transaction should be caught by CheckPoint validation logic.

Charleslee522:
@kfangw Yes. I think that we need a testcase too. But we will talk about changing checkpoint to sequence point or not in the afternoon. I figure I can do it after that.

PR #178, opened 2018-07-09
Use check Amount primitives (+code cleanup)

Geod24:
Pretty straightforward, as long as you review it commit by commit.
The lack of overload or sane implicit type conversion led to some duplication, unfortunately.

### DIFF
--- a/lib/data.go
+++ b/lib/data.go
@@ -3,8 +3,8 @@
 //
 // One BOSCoin accounts for 10 million currency units.
 // In addition to the `Amount` type, some member functions are defined:
-// - Add / Sub do an addition / substraction and return an error object
-// - AddCheck / SubCheck call `Add` / `Sub` and turn any `error` into a `panic`.
+// - `Add` / `Sub` do an addition / substraction and return an error object
+// - `MustAdd` / `MustSub` call `Add` / `Sub` and turn any `error` into a `panic`.
 //   Those are provided for testing / quick prototyping and should not be in production code.
 // - Invariant `panic`s if the instance it's called on violates its invariant (see Contract programming)
 //
--- a/lib/data.go
+++ b/lib/data.go
@@ -85,6 +85,49 @@ func (a Amount) Sub(sub Amount) (Amount, error) {
 	return a - sub, nil
 }
 
+//
+// Add this `Amount` to itself, `n` times
+//
+// If the resulting value would overflow maximumAmount, an error is returned,
+// along with the value (which would trigger a `panic` if used).
+//
+func (a Amount) MultInt(n int) (Amount, error) {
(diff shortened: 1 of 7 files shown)


lib/data.go
```diff
@@ -85,6 +85,49 @@ func (a Amount) Sub(sub Amount) (Amount, error) {
...
+// Counterpart of `Mult` which panic instead of returning an error
+// Useful for debugging and testing, should be avoided in regular code
+func (a Amount) MustMult(n int) Amount {
+	if v, err := a.Mult(n); err != nil {
+		panic(err)
+	} else {
+		return v
+	}
+}
+
```
Charleslee522: How about using function name like this for readability? It's not mandatory :)
```
MultiplyInt
MultiplyUint
MultiplyInt64
MultiplyUint64
MustMultiply
```

lib/transaction.go
```diff
@@ -150,17 +150,29 @@ func (tx Transaction) Source() string {
...
 
+//
+// Returns:
+//   the total monetary value of this transaction,
+//   which is the sum of its operations,
+//   optionally with fees
+//
+// Params:
+//   withFee = If fee should be included in the total
+//
```
Charleslee522: 👍

lib/transaction.go
```diff
@@ -150,17 +150,29 @@ func (tx Transaction) Source() string {
...
+	var amount Amount
 	for _, op := range tx.B.Operations {
-		amount += int64(op.B.GetAmount())
+		amount = amount.MustAdd(op.B.GetAmount())
 	}
 
+	// TODO: This isn't checked anywhere yet
 	if withFee {
-		amount += int64(len(tx.B.Operations)) * int64(tx.B.Fee)
+		amount = amount.MustAdd(tx.B.Fee.MustMult(len(tx.B.Operations)))
```
Charleslee522: It's clearer than before :)

lib/data.go
Geod24: Hum I attempted to keep the name of the operation shorts (e.g. with have `Sub` not `Subtract`). I'll add the types though.

Geod24:
Added some tests but codecov don't seem to want to update the report.
@Charleslee522 : I cannot see the report correctly [here](https://codecov.io/gh/bosnet/sebak/pull/178/changes). Any idea why ?

Geod24:
> I think because your new test doesn't cover new statements.

`MultUint` was missing, it is fixed now.
`MultInt64` should be covered by `MultInt` (and by extension, `MustMult`).

Charleslee522:
@Geod24 I guess, a case which overflow maximumAmout in `MultUint` function was missing too.

Geod24:
Should be okay now. Coverage looks 👌

PR #186, opened 2018-07-11
Properly link files and converage in Codecov

Geod24:
Before the code was more complicated and not working.
This new code use the example provided in the official documentation:
https://github.com/codecov/example-go


Charleslee522:
@Geod24 `File not found in report` also in this pr.
https://codecov.io/gh/bosnet/sebak/pull/186/changes

outsideris:
Ooops! I thought `N/A` means `no difference`.

Geod24:
Yeah there's still something off. It detects the proper path now, but still doesn't link... I'll ping when I find the solution.

Geod24:
So with the current iteration, I managed to get a file view [per commit](https://codecov.io/gh/bosnet/sebak/src/6358a2d541b68e4ad277e509b6af08c803ab489b/lib/data.go).

However, it does not show up in the PR 🤔

Geod24:
So I've changed this to run all 3 phases in parallel. I have no clue why it doesn't show the file on the PR when it does show it on the commit. We won't have a diff since the base branch has no coverage for those files. I'm itching to just merge as is since it (indirectly) works.

outsideris:
Try to change source code for forcing to change coverage.

outsideris:
I guess it is codecov's problem.
How about open a new PR to test it.

Geod24:
> How about open a new PR to test it.

I can try it, but I don't think that's going to help.
I'm wondering if its flopping because there is no "base" report.

PR #193, opened 2018-07-23
Add verbose and dry run mode to sebak wallet pay

Geod24:
```
This will allow people to create test cases for our integration test suite more easily.
```

E.g. look at the JSON files in #188

### DIFF
--- a/cmd/sebak/cmd/wallet/payment.go
+++ b/cmd/sebak/cmd/wallet/payment.go
@@ -20,6 +20,7 @@ var (
 	PaymentCmd        *cobra.Command
 	flagEndpoint      string
 	flagCreateAccount bool
+	flagDry           bool
 )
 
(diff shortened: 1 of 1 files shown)


outsideris:
Do you want to comapre with printed output in tests?

Geod24:
At the moment the tests don't allow us to do that. Also, the testsuite as designed in #188 is not intended to test the CLI interface (although it would be possible), but focuses on the `node`, which is IMO the most important part ATM.

Geod24:
In fact, after implementing another test, I think it would be beneficial to also add a verbose option which both sends the request and prints it.

Geod24:
@outsideris : I forgot to push the verbose flag originally. Can you give it a second look ?

outsideris:
Why not. Looks good.

PR #195, opened 2018-07-25
Validate fund before sending a transaction

Geod24:
Fairly simple and can avoid some mistakes.

### DIFF
--- a/cmd/sebak/cmd/wallet/payment.go
+++ b/cmd/sebak/cmd/wallet/payment.go
@@ -30,6 +30,7 @@ func init() {
 		Run: func(c *cobra.Command, args []string) {
 			var err error
 			var amount sebak.Amount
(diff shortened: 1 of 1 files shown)


cmd/sebak/cmd/wallet/payment.go
```diff
@@ -82,6 +83,21 @@ func init() {
 				os.Exit(1)
 			}
 
+			// Check that account's balance is enough before sending the transaction
+			{
+				newBalance, err = sebak.MustAmountFromString(senderAccount.Balance).Sub(amount)
```
outsideris: Why not just `Sub(amount - sebak.BaseFee)` here?

outsideris: Ok. I see.

Issue #194, opened 2018-07-24
Trying to overdraft kills an account


Geod24:
While writing the integration tests, I came accross this bug. If I try to overdraft from an account, the account gets stuck. Using the docker containerized network:

```console
% sebak wallet payment $PUBKEY2 100_000 $SECRET1 --endpoint=https://127.0.0.1:2822 --create --verbose
Account before transaction:  {GDIRF4UWPACXPPI4GW7CMTACTCNDIKJEHZK44RITZB4TD3YUM6CCVNGJ 10000000000000000000 5nc4PvPKMxDTM6ets3hudktkA-5nc4PvPKMxDTM6ets3hudktkA}
{
  "T": "transaction",
  "H": {
    "version": "",
    "created": "2018-07-24T01:23:11.644513500Z",
    "hash": "8urE17YWtVEMxGa8dwB1oZsaanm5Cisk9JtAVozVTgxY",
    "signature": "4pVJ496WGWuSAfQxG3pEUkMUwnGSSaBxpPtJZTT5kChZt7HEkAQB4xQRmfSG3zSQRVdxGpBSbG1EWh5WfXKjmYZo"
  },
  "B": {
    "source": "GDIRF4UWPACXPPI4GW7CMTACTCNDIKJEHZK44RITZB4TD3YUM6CCVNGJ",
    "fee": "10000",
    "checkpoint": "5nc4PvPKMxDTM6ets3hudktkA-5nc4PvPKMxDTM6ets3hudktkA",
    "operations": [
      {
        "H": {
          "type": "create-account"
        },
        "B": {
          "target": "GAYGELM74WJMKSLDN5YP2VAMP64WC4IXIGICUNK2SCVIT7KPTLY7M3MW",
          "amount": "100000"
        }
      }
    ]
  }
}
Receiver account after 5 seconds:  {GAYGELM74WJMKSLDN5YP2VAMP64WC4IXIGICUNK2SCVIT7KPTLY7M3MW 100000 5nc4PvPKMxDTM6ets3hudktkA-8urE17YWtVEMxGa8dwB1oZsaanm5Cisk9JtAVozVTgxY}
% sebak wallet payment $PUBKEY1 1_000_000 $SECRET2 --endpoint=https://127.0.0.1:2822 --create --verbose
[...]

PR #199, opened 2018-08-01
Add tests for overdraft

Geod24:
A test case for #194 - will not pass until the bug is fixed, but a good starting point if anyone wants to tackle the issue.


Geod24:
Fixed the tests, now it's passing and testing much more stuff

PR #207, opened 2018-08-02
Update Gopkg.lock to v0.5.0 format

Geod24:
A recent update to dep changed the package format, so this update it

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,150 +3,197 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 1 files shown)


outsideris:
If we merge this, we should update `dep` to v0.5.0+?

Geod24:
You mean, developers ? Yes
Because `docker` is already [`go get`ing it](https://github.com/bosnet/sebak/blob/358d22157863478311d3d552eeee18dfb0fe2cef/Dockerfile#L15)

PR #208, opened 2018-08-02
Remove pull request template

Geod24:
Literally no one fills it.
We can reintroduce a more sensible one if it proves to be an issue with external contributors,
when we get external contributors.


Charleslee522:
I don't agree with this pr. I always fill the pr template and it is not only for external contributors. And I didn't know this pr from now...
I figure, it was related with development process, so it needed more reviewers.

outsideris:
@Charleslee522 I believe that we need a ground role for merge.
I think that consensus teams will heavily use this repo and follow more formal rule than TC or infra.
So, it is best that members of consensus make the rule.

And if you don't agree you can revert it and we can make new consensus again on the new PR.

Issue #99, opened 2018-06-19
Stop flooding /connect and use Keep-Alive instead


Geod24:
We use HTTP2 and setup [keep-alive](https://github.com/bosnet/sebak/blob/bbed51cdeb85fb40a10d01c771e1a5889cd9bf99/lib/common/http2_client.go#L23-L26) but we keep flooding the `/connect` method every second or so.

The [handling](https://github.com/bosnet/sebak/blob/bbed51cdeb85fb40a10d01c771e1a5889cd9bf99/lib/node_runner.go#L201-L206) of the message is also [not used](https://github.com/bosnet/sebak/blob/bbed51cdeb85fb40a10d01c771e1a5889cd9bf99/lib/common/node.go#L187-L194).

@spikeekips is there any reason this was designed this way ?

spikeekips:
@Geod24 Yeah, log messages is noisy. To check the connection is live, `Keep-Alive` seems not to be good enough, because if connection is dead, it will not be caught until idle timeout. 

We can change to log only when first failure of connection.

Geod24:
> it will not be caught until idle timeout.

We can set a low timeout, to match the current behavior.

spikeekips:
> @Geod24 : We can set a low timeout, to match the current behavior.

Yes, it makes sense.

spikeekips:
@Geod24 I have question, if node is down, can the other nodes know that? If not, the broadcasting may be matter.

Geod24:
Fixed in #211

PR #211, opened 2018-08-02
Do not reconnect every second after the initial connection

Geod24:
```
Once we are connected to a node, stop sending `/connect` messages.
By setting an `idleTimeout` of `0` and `keepAlive` to `true`
we ensure that the connection is never closed.
We're still suffering from bugs when a node comes back up,
but since Sebak does not currently support it,
it isn't a problem.
```

### DIFF
--- a/lib/common/http2_client.go
+++ b/lib/common/http2_client.go
@@ -28,8 +28,8 @@ func NewHTTP2Client(timeout, idleTimeout time.Duration, keepAlive bool) (client 
 		IdleConnTimeout:   idleTimeout,
 		DisableKeepAlives: !keepAlive,
 		DialContext: (&net.Dialer{
-			Timeout:   1 * time.Second,
-			KeepAlive: 100000 * time.Second,
+			Timeout:   3 * time.Second,
+			KeepAlive: 1 * time.Second,
 			DualStack: true,
 		}).DialContext,
 	}
--- a/lib/network/http2_config.go
+++ b/lib/network/http2_config.go
@@ -28,7 +28,6 @@ func NewHTTP2NetworkConfigFromEndpoint(nodeName string, endpoint *common.Endpoin
 	var ReadTimeout time.Duration = 0
 	var ReadHeaderTimeout time.Duration = 0
 	var WriteTimeout time.Duration = 0
-	var IdleTimeout time.Duration = 5
 	var TLSCertFile, TLSKeyFile string
 
 	if ReadTimeout, err = time.ParseDuration(common.GetUrlQuery(query, "ReadTimeout", "0s")); err != nil {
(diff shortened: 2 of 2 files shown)


Geod24:
Fixes #99

lib/network/connection_manager.go
```diff
@@ -127,6 +127,7 @@ func (c *ConnectionManager) connectingValidator(v *sebaknode.Validator) {
 		if c.setConnected(v, err == nil) {
 			if err == nil {
 				c.log.Debug("validator is connected", "validator", v)
+				break
```
cmcm2222: In case of connection error, How about graceful kill node itself?

lib/network/http2_network_config.go
```diff
@@ -31,7 +31,6 @@ func NewHTTP2NetworkConfigFromEndpoint(endpoint *sebakcommon.Endpoint) (config H
 	var ReadTimeout time.Duration = 0
 	var ReadHeaderTimeout time.Duration = 0
 	var WriteTimeout time.Duration = 0
-	var IdleTimeout time.Duration = 5
```
cmcm2222: @Geod24 For easy source code recognition, IMHO it'd better leave previous code. 
what do you think? this is just suggestion. :)

lib/network/connection_manager.go
cmcm2222: IMHO Some error message came out , I think it'd be better shut down node and turn it on again.

Geod24: That's only for connecting to peer, it's not a node error in itself.

cmcm2222: Ok. then I have no problem.

lib/network/http2_network_config.go
Geod24: I don't see why ? This is just a variable in the config parsing function.
I left `IddleTimeout` in the structure (L86) even though it is fixed to `0` for the reason you mention, but I don't think it's more readable to have `IdleTimeout: IdleTimeout` and look up `IdleTimeout` just to find out that it's constant.

Geod24:
Ping ?

spikeekips:
@Geod24 Sorry too late :) I have question, if node is down, can the other nodes know that? If not, the broadcasting may be matter.

Geod24:
@spikeekips yes it can be detected, because sending the request will error out. This situation is currently not well handled by sebak, and this PR doesn't fix that issue, but it's a first step to do so.
Note that currently this bug can happen, but since we're flooding `/connect` every second, it requires automated testing to trigger.

spikeekips:
@Geod24 Currently the network part supports to know the disconnected nodes, so it would be perfect this PR supports it.

Geod24:
> @Geod24 Currently the network part supports to know the disconnected nodes, so it would be perfect this PR supports it.

No it does not. Currently [when we broadcast](https://github.com/bosnet/sebak/blob/1be170628de3387e8976313382b7019d628f8bce/lib/network/connection_manager.go#L158-L169) we just iterate over `connected`, without checking the boolean in it. This code comes from the defunct `AllConnected` method [which was doing exactly the same thing](https://github.com/bosnet/sebak/commit/8c059e3a102051dad4718c1dc6c613bf98438528).

Geod24:
> If one node is disconnected, it is not counted.

Can you clarify ? It sounds like if one node is disconnected, it should be counted as not voting / no. Are you saying that if I have 3 validators, but 2 of them are disconnected, I effectively have only one validator ?

spikeekips:
@Geod24 only for INIT -> SIGN, the disconnected nodes are not counted.

Geod24:
What's the rationale behind this ?

spikeekips:
@Geod24 Under the new design of consensus, this PR can be applied. Please keep going on.

Geod24:
@spikeekips : It's green so it's just waiting for your approval.

PR #213, opened 2018-08-06
Cleanup to NodeRunner / LocalNode / Validator

Geod24:
I was working on some test code (as my branch name suggests...) when I had some issues with the design. After digging a bit deeper, I saw that some code could be simplified / removed, so I did it, and added some comments to the modules so it would be easier for people to grasp the design.

Best reviewed commit-by-commit.


lib/node/localNode.go
```diff
@@ -1,3 +1,11 @@
+//
+// Defines the `LocalNode` type of Node, which is our node
```
cmcm2222: LGTM

lib/node/validator.go
```diff
@@ -1,3 +1,9 @@
+//
+// Defines the `Validator` type of Node, which is a remote node
```
cmcm2222: LGTM

lib/node/localNode.go
```diff
@@ -83,7 +81,7 @@ func (n *LocalNode) SetTerminating() {
 }
 
 func (n *LocalNode) Address() string {
-	return n.address
+	return n.keypair.Address()
```
cmcm2222: LGTM

lib/node/localNode.go
```diff
@@ -30,6 +30,35 @@ type LocalNode struct {
 	validators map[ /* Node.Address() */ string]*Validator
 }
 
+func NewLocalNode(address string, endpoint *sebakcommon.Endpoint, alias string) (n *LocalNode, err error) {
```
cmcm2222: LGTM

lib/node/localNode.go
```diff
@@ -30,6 +30,35 @@ type LocalNode struct {
...
+		alias:      alias,
+		address:    address,
+		endpoint:   endpoint,
+		validators: map[string]*Validator{},
+	}
+
+	return
+}
+
+func NewLocalNodeFromString(b []byte) (*LocalNode, error) {
```
cmcm2222: This func have other functionality?? 
it seems just deserialize object isn't it?

Geod24: Oh yeah, it's removed in a later commit actually (it's not added, just moved)

sconeman:
It is late but, LGTM.

Issue #212, opened 2018-08-06
API: Get transactions from proposer


Charleslee522:
- Background
In the new SEBAK, it is not checked that all nodes have same transactions after sharing. Therefore, if some transactions in the proposed block are not included in the transaction pool, then it needs to request and receive that transactions from proposer.

- Source position to fix
node_runner_round_transaction_checker.go:50(CheckNodeRunnerHandleTransactionsGetMissingTransaction method)

Issue #205, opened 2018-08-01
Transaction sharing without INIT voting


Charleslee522:
The old version of ISAAC shared the transaction by transferring the INIT ballot voting of the existing ISAAC.
It was not good for performance and resource. So we need a better solution for transferring transactions.

sconeman:
@spikeekips A ballot state called `TXSHARE` was created. Is this right before `INIT`?

Charleslee522:
@spikeekips @sconeman I already removed this state, but it has revived...
I'm going to check it again :)

sconeman:
@Charleslee522 Aha... So I will remove again.

Charleslee522:
@sconeman That's ok. I can handle it :)

Issue #204, opened 2018-08-01
Changing checkpoint to sequenceID to defend replay attack


Charleslee522:
The transaction has the Checkpoint to defend replay attack. But it could be changed to sequenceID of sender. Then, it would be efficient and simple.

```
Checkpoint string -> SequenceID uint64
```

Charleslee522:
#203

kfangw:
how is going? @Geod24

spikeekips:
This issue can be closed, right?

Geod24:
Right!

Issue #203, opened 2018-08-01
Transaction validation for the new ISAAC


Charleslee522:
Validating every transactions in proposed block is needed.
It should validates,
- balance of source
- sequenceID
- base fee
- verify signature
- transaction hash
- etc...

Charleslee522:
@spikeekips which features do you want to check for tx validation?

spikeekips:
@Charleslee522 The things you pointed is done by `Transaction.IsWellformed()`, so I try to add like these,

* check there are same source in the proposed transactions
* check source and target accounts exist
* check the total amount of transaction is valid
* etc.

Charleslee522:
@spikeekips I want you to explain more about "check there are same source in the proposed transactions" please :)

spikeekips:
@Charleslee522 If there are 2 or more transactions from same source account in the proposed transactions, the first one will be considered as valid, the left will not be included in `ValidTransactions`.

Charleslee522:
@spikeekips Why the second tx from the same source is invalid? If balance is enough and tx has no dependency with the first one, then isn't it ok.

Geod24:
> If balance is enough and the second tx has no dependency with the first one, then isn't it ok?

It's one or the other, but you cannot have both.
If we use sequence ids, they either have the same sequence ID (= replay) or one has sequence id `S` and the other `S + 1`.

spikeekips:
@Charleslee522 👍

PR #215, opened 2018-08-07
WIP: New proposer consensus model

### LINKED ISSUE #202: Proposer selection function
Every nodes in network must select the proposer by itself using the function below.

IsaacRound.go:CalculateProposer()

Input
- Connected node addresses
- Block Height
- Round

Output
- Proposer node address

spikeekips:
### Github Issue

Resolves #202 #203 #205 

Related #200 #204 #212 

### Background

The consensus model of the previous ISAAC is based on the state transition, the new model is based on the round voting. The more details will be introduced in [How ISAAC Consensus Protocol Works](https://github.com/bosnet/sebak/wiki/How-ISAAC-Consensus-Protocol-Works).

### Solution

With new round model,

* The transactions will be stored in the new `Block` structure.
* The `Block` has height and it's hash, this will make the entire stored data to be more structured, so the catchup can be easily processed.
* Each validator selects the proposer in every round.
* The confirmed transactions will have order.
* The incoming transactions from client will be broadcasted without consensus.

#### Voting Process
Originally the new model will have 2 voting process,

* `SIGN`: voting for received `Ballot` from proposer
* `ACCEPT`: voting for confirmed `Ballot` of each validator

In `ACCEPT` state, validator will compare the voting result of the other validators with his/her voting result.

At this implementation, `SIGN` and `ACCEPT` voting process is not yet implemented.

#### Etc

The existing `Ballot` and `NodeRunner` was replaced by the new design.


Charleslee522:
The 2 voting process will be implemented in #216

sconeman:
I am adding a test case of proposer selection function(#202) in `lib/node_runner_test.go`.

lib/consensus.go
```diff
@@ -9,9 +9,9 @@ type Consensus interface {
...
 	HasMessageByHash(string) bool
-	ReceiveMessage(sebakcommon.Message) (Ballot, error)
-	ReceiveBallot(Ballot) (VotingStateStaging, error)
+	ReceiveMessage(sebakcommon.Message) error
+	//ReceiveBallot(Ballot) (VotingStateStaging, error)
 
-	AddBallot(Ballot) error
-	CloseConsensus(Ballot) error
+	//AddBallot(Ballot) error
+	//CloseConsensus(Ballot) error
```
Charleslee522: These methods(GetNode, HasMessage, HasMessageByHash, ReceiveMessage) are not used.
@spikeekips How about removing consensus.go or redesign interface? :)

lib/common/util_benchmark_test.go
```diff
@@ -0,0 +1,170 @@
+package sebakcommon
```
Geod24: I think any kind of benchmark should be in its own source tree and run on demand.
We should not have benchmark as part of the unit tests.

lib/consensus.go
Geod24: Don't comment out code, remove it instead, please. That's what VCS are for.

lib/node_runner_transaction_checker.go
```diff
@@ -0,0 +1,157 @@
+package sebak
```
Geod24: Please add documentation for all new modules, and for public symbols please.

lib/consensus.go
spikeekips: @Charleslee522 Yes, I agree. We can remove `Consensus` interface.

Charleslee522:
@Geod24 Yes, you're right. It is true that it is necessary to break down and document issues.
But I figure, this pr is a little different with other prs because this pr is bottleneck of our group members and also other groups now. So, after this pr is resolved, most of our team members can join,  suggest and handle the following issues.
Therefore, I think it would be better to fix problems and document with new issues after merging it.
How do you think?

Geod24:
@Charleslee522 : The consensus protocol is without doubt the most crucial part of Sebak: changes can have far-reaching consequences, and just one bug going in unnoticed could destroy our entire system when we switch MainNet to Sebak.

In addition to that, if there is any flaws in the design, how can reviewers see it if they are not able to review this PR ? Designs flaws can easily lead us to waste weeks or months of development time. I think the benefits brought by making this PR reviewable far outweight its cost.

The more complex a PR is, the more quality I expect from it. Accepting lower quality code for higher-impact PR is **not** the way to deliver a successful PR.

Charleslee522:
@Geod24 I agree with you if the baseline is solid as in the following assumptions.
1. We have been running the publishing for a few months without big problems,
2. Because of this, we believe that we have established a proven reliable network.

But,
1. We did not publish yet, we did not even run a test-net yet.
2. Therefore, the existing SEBAK is not fully verified.
I think the consensus protocol is still being built, just as it was when first introduced the SEBAK implementation.

Besides, the design hasn't been slightly upgraded from the existing one, but it has been reorganized into a completely new design. So I think it would be faster to follow the consensus scenario and understand the new design than just to read the code changes in the existing base. I did it.

lib/node_runner.go
```diff
@@ -186,93 +199,267 @@ func (nr *NodeRunner) SetHandleBallotCheckerFuncs(
...
+	if len(nr.consensus.RunningRounds) > 0 {
 		return
 	}
 
-	if err = nr.Consensus().CloseConsensus(checker.Ballot); err != nil {
-		nr.Log().Error("new failed to close consensus", "error", err)
+	nr.StartNewRound(0)
+}
+
+func (nr *NodeRunner) CalculateProposer(blockHeight uint64, roundNumber uint64) string {
```
sconeman: Unfortunately, I found this function is not fair for candidates.
I make a testcase like below.
```
func TestNodeRunnerCalculateProposer(t *testing.T) {
	defer sebaknetwork.CleanUpMemoryNetwork()

	numberOfNodes := 3
	nodeRunners := createTestNodeRunnerWithReady(numberOfNodes)

	nr0 := nodeRunners[0]

	proposer00 := nr0.CalculateProposer(0, 0)
	proposer01 := nr0.CalculateProposer(0, 1)
	proposer02 := nr0.CalculateProposer(0, 2)
	proposer03 := nr0.CalculateProposer(0, 3)
	proposer04 := nr0.CalculateProposer(0, 4)
	proposer05 := nr0.CalculateProposer(0, 5)

	if proposer00 == proposer01 {
		t.Error("failed to calculate correct proposers. proposer00", proposer00, "proposer01", proposer01)
	}
	if proposer00 == proposer02 {
		t.Error("failed to calculate correct proposers. proposer00", proposer00, "proposer02", proposer02)
	}
	if proposer01 == proposer02 {
		t.Error("failed to calculate correct proposers. proposer01", proposer01, "proposer02", proposer02)
	}
	if proposer00 != proposer03 {
		t.Error("failed to calculate correct proposers. proposer00", proposer00, "proposer03", proposer03)
	}
	if proposer01 != proposer04 {
		t.Error("failed to calculate correct proposers. proposer01", proposer01, "proposer04", proposer04)
	}
	if proposer02 != proposer05 {
		t.Error("failed to calculate correct proposers. proposer02", proposer02, "proposer05", proposer05)
	}

	proposer10 := nr0.CalculateProposer(1, 0)
	proposer11 := nr0.CalculateProposer(1, 1)
	proposer12 := nr0.CalculateProposer(1, 2)
	proposer13 := nr0.CalculateProposer(1, 3)
	proposer14 := nr0.CalculateProposer(1, 4)
	proposer15 := nr0.CalculateProposer(1, 5)

	if proposer10 == proposer11 {
		t.Error("failed to calculate correct proposers. proposer10", proposer10, "proposer11", proposer11)
	}
	if proposer10 == proposer12 {
		t.Error("failed to calculate correct proposers. proposer10", proposer10, "proposer12", proposer12)
	}
	if proposer11 == proposer12 {
		t.Error("failed to calculate correct proposers. proposer11", proposer11, "proposer12", proposer12)
	}
	if proposer10 != proposer13 {
		t.Error("failed to calculate correct proposers. proposer10", proposer10, "proposer13", proposer13)
	}
	if proposer11 != proposer14 {
		t.Error("failed to calculate correct proposers. proposer11", proposer11, "proposer14", proposer14)
	}
	if proposer12 != proposer15 {
		t.Error("failed to calculate correct proposers. proposer12", proposer12, "proposer15", proposer15)
	}

	for _, nr := range nodeRunners {
		nr.Stop()
	}
}
```
And I got messages like below.
```
--- FAIL: TestNodeRunnerCalculateProposer (3.32s)
	node_runner_test.go:180: failed to calculate correct proposers. proposer01 GAJKIJVMZKWQWAH74YOHTRREZT6CO6F2BVK6ABWXMWSM5SRLYNFIAO5T proposer04 GDSVSVTQF3LENGRR2VC2HWWFVVPD5HG2ITMKGFBOE2OLKRYZCL6OJ3CC
	node_runner_test.go:197: failed to calculate correct proposers. proposer10 GCCMSRGVCN5GZAP7YDRRRTYGGNDEA2AJKAFJOKTFI2PXQOE2WX3QSXXT proposer12 GCCMSRGVCN5GZAP7YDRRRTYGGNDEA2AJKAFJOKTFI2PXQOE2WX3QSXXT
	node_runner_test.go:203: failed to calculate correct proposers. proposer10 GCCMSRGVCN5GZAP7YDRRRTYGGNDEA2AJKAFJOKTFI2PXQOE2WX3QSXXT proposer13 GDSVSVTQF3LENGRR2VC2HWWFVVPD5HG2ITMKGFBOE2OLKRYZCL6OJ3CC
	node_runner_test.go:206: failed to calculate correct proposers. proposer11 GAJKIJVMZKWQWAH74YOHTRREZT6CO6F2BVK6ABWXMWSM5SRLYNFIAO5T proposer14 GDSVSVTQF3LENGRR2VC2HWWFVVPD5HG2ITMKGFBOE2OLKRYZCL6OJ3CC
	node_runner_test.go:209: failed to calculate correct proposers. proposer12 GCCMSRGVCN5GZAP7YDRRRTYGGNDEA2AJKAFJOKTFI2PXQOE2WX3QSXXT proposer15 GDSVSVTQF3LENGRR2VC2HWWFVVPD5HG2ITMKGFBOE2OLKRYZCL6OJ3CC
```
According to my analysis, proposers at block 1 and round 0, 1, 2 are same. Like this, this function seems to be more fairness.

sconeman:
#202 done.

lib/node_runner_test.go
```diff
@@ -148,3 +149,95 @@ func TestNodeRunnerCreateAccount(t *testing.T) {
...
+		if p == nr0.localNode.Address() {
+			numN0inProposers++
+		} else if p == nr1.localNode.Address() {
+			numN1inProposers++
+		} else if p == nr2.localNode.Address() {
+			numN2inProposers++
+		}
+	}
+
+	passCriteria := int(maximumBlockHeight) * int(maximumRoundNumber) / numberOfNodes * 2
```
Charleslee522: @sconeman I want to know that why the criteria is double of average. 
ps. Using variable like `maximumBlockHeight` is better than just using constant(magic number) for readability.

lib/common/util_benchmark_test.go
Charleslee522: @Geod24 Is that common rule or just your think? It's just curious because I'm not familiar with benchmark test :)
As you know, the benchmark tests are not running when you didn't include '-bench' option.

PR #219, opened 2018-08-07
Contract

kfangw:
Contract executor/deployer and related apis


anarcher:
But ONT wasm is failed to vendoring with dep..

kfangw:
I removed the ont. added dummy code with go-wagon

PR #228, opened 2018-08-08
contract deploy/execute operation

kfangw:
New operation deploy/execute for contract 

When I write this features, I realize that we need more implementation for this.
- StateStore layer
- A passage for contract executor(like a wallet) to get the result of the execution

Additionally, I want to refactor FinishXXX as `Do` function.


kfangw:
@anarcher I want to refactor `Operation` but It will be done when the proposer PR merged.
I will merge this branch without the Do function.

Issue #221, opened 2018-08-07
`Value` Package


kfangw:
- Define data types for contract
- (De)Serialization of the data types
- Conversion from the value to each contract execution engine, vice versa

PR #230, opened 2018-08-08
value encode/decode

kfangw:
Define types for Value
Encode/Decode

#221


kfangw:
@anarcher I've rewritten the value package

Value hold interface{} and serialize only when needed.

kfangw:
This will be merged to `contract` branch

PR #232, opened 2018-08-08
Cleanup some checkpoint-related code

Geod24:
Since I started to work on replacing checkpoints (again) this iteration, I stumbled upon a few places where they were not necessary, and they are good first place to act on.
I also included a mini commit which removes a needless (AFAICS, maybe there's something I'm not seeing @spikeekips ?) `select {}`

Best reviewed commit-by-commit.

### DIFF
--- a/lib/test.go
+++ b/lib/test.go
@@ -4,11 +4,10 @@ import (
 	"fmt"
 	"math/rand"
 
-	"github.com/btcsuite/btcutil/base58"
+	"boscoin.io/sebak/lib/common"
+
 	"github.com/google/uuid"
 	"github.com/stellar/go/keypair"
-
-	"boscoin.io/sebak/lib/common"
 )
 
 func testMakeBlockAccount() *BlockAccount {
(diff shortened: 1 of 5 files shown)


lib/test.go
```diff
@@ -98,7 +97,7 @@ func TestMakeTransaction(networkID []byte, n int) (kp *keypair.Full, tx Transact
 }
 
 func TestGenerateNewCheckpoint() string {
-	return base58.Encode([]byte(uuid.New().String()))
+	return uuid.New().String()
```
sconeman: I want to know your intentions for this part. I was just curious to ask :)

Geod24: That's how it's done everywhere else:
- https://github.com/bosnet/sebak/blob/ce6376728fcda0e86df903cceaa65d7f918930ae/lib/test.go#L83
- https://github.com/bosnet/sebak/blob/bbed51cdeb85fb40a10d01c771e1a5889cd9bf99/lib/node_runner_etc_test.go#L30
- https://github.com/bosnet/sebak/blob/058b76afa0ace3e82a9b828becc4e6ceb8af184f/lib/node_runner_memory_network_test.go#L36
- https://github.com/bosnet/sebak/blob/c8fa7bd55d54f87f6d24f3f98bb973613cc20e8f/lib/node_runner_create_account_test.go#L178
- https://github.com/bosnet/sebak/blob/dd30a470730017079ee9be32e2d63fd3cc3a0038/lib/block_transaction_test.go#L144

PR #234, opened 2018-08-08
[WIP] Move contract to lib and store too.

anarcher:
- For fix import cycle, move `contract/` pkg to `lib/ `
- I will add test code too. 
- IMHO, unfortunately, with current one mono pkg style, we can not more separate `contract/`..


Geod24:
What leads to the cycle ?

anarcher:
@Geod24 lib/ import contract but. Contract/ has to use lib/block_account.  If we separate block_acount to lib/block/block_account and lib/amount too.

anarcher:
Sadly,  We don't have room to use `Amount` in `contract/`, because of  `amount` is in `lib/data.go`..
Do you have an idea for it? T_T

Geod24:
I tried to break down those dependencies a while ago, but block_account depends on operation as well, which depends on account. In the end, it's very hard to correctly split code with Go's package system because the level of isolation is either the directory, or nothing :/

anarcher:
> I tried to break down those dependencies a while ago, but block_account depends on operation as well, which depends on account. In the end, it's very hard to correctly split code with Go's package system because the level of isolation is either the directory, or nothing :/

I don't mind to blame Go package system. Because of many go projects which has many packages looks works well  [1],[2],[3],... 

Anyway,  If we break the mono pkg structure,  and if the block_account.go moves to `lib/block/block_account.go` , I think that it works well. (but we have to a lot of typing.. or surpporting refactoring tool?)

[1]:  https://github.com/kubernetes/kubernetes 
[2]: https://github.com/hashicorp/terraform
[3]: https://github.com/ethereum/go-ethereum

lib/contract/jsvm/executor_test.go
```diff
@@ -3,33 +3,48 @@ package jsvm
...
-	testStateClone.PutDeployCode(deployCode)
 
-	context := &context.Context{
-		StateStore: testStateStore,
-	}
-	api := api.NewAPI(context, testAddress, nil)
-	ex := NewOttoExecutor(context, api, deployCode)
+	ctx := test.NewMockContext(testAddress, nil)
+	api := test.NewMockAPI(ctx, testAddress)
+	api.SetHelloworldFunc(func(ctx context.Context, greeting string) (string, error) {
```
anarcher: `context.Context` and `api.API` are interface, So testing within `contract/jsvm` or another places in `contract`,  We can use  `contract/test/MockAPI` and `contract/test/MockContext`.  Their was not complete for now, but this test works well with this Mock struct.

lib/contract/test/api.go
```diff
@@ -0,0 +1,101 @@
...
+	"boscoin.io/sebak/lib/contract/storage"
+	"boscoin.io/sebak/lib/contract/value"
+)
+
+type (
+	HelloworldFunc     func(context.Context, string) (string, error)
+	GetStorageItemFunc func(context.Context, string) (*storage.StorageItem, error)
+	PutStorageItemFunc func(context.Context, string, *storage.StorageItem) error
+	CallContractFunc   func(context.Context, *payload.ExecCode) (*value.Value, error)
+)
```
anarcher: It seems looks ugly. but the pattern that the MockAPI use is simple. 
It looks like ugly DI.

lib/contract/test/context.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+func NewMockContext(addr string, dcFunc DeployCodeFunc) *MockContext {
+	ctx := &MockContext{
+		senderAddress:  addr,
+		deployCodeFunc: dcFunc,
+	}
+
+	return ctx
+}
```
anarcher: It can be changed like `MockAPI`.

anarcher:
Can I merge this PR? @kfangw

PR #235, opened 2018-08-09
move amount to common package

kfangw:
I want to restructure whole sebak packages
However, It is not easy because there is a big change coming.

Maybe... It is a start point I think.
`Amount` has no dependency to sebak package. But almost every package depends on the structure.

Could you guys accept this refactoring? 
It will be very helpful for the contract package to keep the modular design.

### DIFF
--- a/cmd/sebak/common/util.go
+++ b/cmd/sebak/common/util.go
@@ -7,7 +7,7 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"boscoin.io/sebak/lib"
+	"boscoin.io/sebak/lib/common"
 )
 
 /**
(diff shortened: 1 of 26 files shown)


kfangw:
@anarcher @spikeekips @Charleslee522 Could you review this?
maybe your proposal branch rebase or merge this change.

lib/common/_data.go
```diff
@@ -0,0 +1,15 @@
+package sebakcommon
```
spikeekips: I wish the code in this file, `_data.go` can be moved to `data.go`. Is there any reason to make new file?

Geod24: There was a file named `data.go` previously. I think the simplest solution would be to keep the previous file as is, and use `amount.go` instead of `data.go` for the new file. @kfangw ?

lib/common/data.go
```diff
@@ -1,6 +1,22 @@
...
-import "encoding/json"
+import (
+	"fmt"
+	"strconv"
+
+	"boscoin.io/sebak/lib/error"
+	"encoding/json"
+)
 
 func EncodeJSONValue(v interface{}) (b []byte, err error) {
```
Geod24: I don't think both files should be merged

cmd/sebak/common/util.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"boscoin.io/sebak/lib"
+	"boscoin.io/sebak/lib/common"
```
Geod24: Huge 👍

kfangw:
@spikeekips @Geod24 Adopt your guides
please review again

Issue #225, opened 2018-08-07
Validating transaction operations which has contract  from proposed ballot 


anarcher:
Each node has to validate the transaction from a ballot and `stateRoot` from this ballot.
So every node executes the contract codes in transaction's operation and makes a `stateRoot` as the result.  
And If it is ok, every node will change the states from this ballot's transactions and will make a block.

Issue #224, opened 2018-08-07
Propose new Ballot with stateRoot. 


anarcher:
- For validating to change states from a block's transactions, The proposer make a hash called `stateRoot` from changed states.  (We need to figure out or decide for a better approach of `stateRoot` in near future.)

PR #236, opened 2018-08-10
Add statedb in storage pkg

anarcher:
### Github Issue
- related: #224 #225 

### Background

- To `*LevelDBBackend` compatibility, The `StateDB`  is implemented by `DBBackend` interface.

```
type DBBackend interface {
	Has(string) (bool, error)
	Get(string, interface{}) error
	New(string, interface{}) error
	Set(string, interface{}) error
	Remove(string) error

	GetIterator(prefix string, reverse bool) (func() (IterItem, bool), func())

	News(...Item) error
	Sets(...Item) error

	Commit() error
	Discard() error
}
```

### Solution

- `StateDB` just keeps  `changedKeys` from  `Set` and `New` methods to make a hash for simple `stateRoot` for contract. 
- Another operations just pass to inner `*LevelDBBackend` in `StateDB`
- `*LevelDBBackend` added `put` method for using the `StateDB`.  This method is not public. so It just uses in `sebakstroage`. 

### Possible Drawbacks

- The interface can be changed by another need of using methods of `LevelDBBackend`. 
- Try to work together `FinishTransaction` and `FinshBallot` too. :-)

### DIFF
--- a/lib/storage/dbbackend.go
+++ b/lib/storage/dbbackend.go
@@ -1,0 +1,17 @@
+package sebakstorage
+
+type DBBackend interface {
+	Has(string) (bool, error)
+	Get(string, interface{}) error
+	New(string, interface{}) error
+	Set(string, interface{}) error
+	Remove(string) error
+
+	GetIterator(prefix string, reverse bool) (func() (IterItem, bool), func())
+
+	News(...Item) error
+	Sets(...Item) error
+
+	Commit() error
+	Discard() error
+}
--- a/lib/storage/leveldb.go
+++ b/lib/storage/leveldb.go
@@ -200,6 +200,21 @@ func (st *LevelDBBackend) Set(k string, v interface{}) (err error) {
 	return
 }
 
+func (st *LevelDBBackend) put(k string, v interface{}) (err error) {
+	var encoded []byte
+	serializable, ok := v.(sebakcommon.Serializable)
+	if ok {
+		encoded, err = serializable.Serialize()
+	} else {
+		encoded, err = sebakcommon.EncodeJSONValue(v)
+	}
+	if err != nil {
+		return
+	}
+	err = st.core.Put(st.makeKey(k), encoded, nil)
+	return
+}
+
 func (st *LevelDBBackend) Sets(vs ...Item) (err error) {
 	if len(vs) < 1 {
 		err = errors.New("empty values")
--- a/lib/storage/statedb.go
+++ b/lib/storage/statedb.go
@@ -4,6 +4,8 @@ import (
 	"sort"
 
 	"boscoin.io/sebak/lib/common"
+	"boscoin.io/sebak/lib/error"
+
 	"github.com/btcsuite/btcutil/base58"
 )
 
@@ -83,7 +85,9 @@ func (s *StateDB) MakeHash() ([]byte, error) {
 	hashes := make([][]byte, 0, len(ks))
 	for _, k := range ks {
 		bs, err := s.levelDB.GetRaw(k)
-		if err != nil {
+		if err == sebakerror.ErrorStorageRecordDoesNotExist {
+			bs = nil
+		} else if err != nil {
 			return nil, err
 		}
 		h := sebakcommon.MakeHash(bs)
--- a/lib/storage/statedb.go
+++ b/lib/storage/statedb.go
@@ -9,14 +9,14 @@ import (
 
 type StateDB struct {
 	levelDB     *LevelDBBackend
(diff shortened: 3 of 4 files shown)


spikeekips:
I guess, the reason why `changedkeys` has `bool` as a value is that if the value is `false` that key will be ignored in `MakeHash()`. Is there any code to set the specific key to `false`?

anarcher:
> I guess, the reason why `changedkeys` has bool as a value is that if the value is false that key will be ignored in MakeHash(). Is there any code to set the specific key to false?

It's a good point. At current `StateDB`, nowhere exists to use this bool value. Maybe  I thought that is room to ignore or not if the `StateDB` needs to specific handling key's status in near future.  With current `StateDB` , `map[string]struct{}`  is enough for now. Because   `map[string]struct{}` is used small memory footpoints than `map[string]bool`.

spikeekips:
> @anarcher : map[string]struct{} is enough for now.
Or `[]string` :)

anarcher:
> @anarcher : map[string]struct{} is enough for now.
> Or []string :)

I have already changed `map[string]struct{}` before this comment. `[]string` is another good option but I feel easier to use `map` instead of slice.  Even If we change for thread-safety when we need it, `map` -> `sync.Map` is an easier option for me. ㅠㅠ

anarcher:
> But it would be better to add testcases for Remove() and Discard() method.
> It's not mandatory :)

As you know, `Discard()` just pass and use  to `statedb.levelDB` for  discarding transactions. :-) 
Anyway, I will add more test code for it :-)

Charleslee522:
@anarcher Oh, sorry. I missed this. Then you don't have to add testcase for Discard method :)

anarcher:
Thanks, Your comments are always helpful! :-) @spikeekips , @Charleslee522

anarcher:
If there are needed more testing and features for it, I will make another PR for it. I will merge this PR for now.

anarcher:
Even though the PR was approved, but I don't feel good to use LevelDB's transaction in StateDB. 
Because LevelDB's tx make a One Global Lock and is expensive and doesn't allow multi transactions (even leveldb.DB (Not Tx struct) cannot do too, it makes sense) if what I know is correct. (Is it right?)

IMHO, It can cause a blocking between DB processing goroutines. For example, a block has many transactions  (+1000 or lower) and process their transactions, another DB based handling are blockings specialty if other routines don't share this one tx. And TXPool decides using DB backend for storing txs and Block processing things use their own tx. but for now what I know is that txpool doesn't use DB. (It is right?)

But, Without leveldb's tx , Implementing `GetIterator` in `StateDB` is not easier than this PR's StateDB. Maybe I make  a merger between memory objects and db objects for Iterator If StateDB has to support the `GetIterator `. Maybe that `StoreDB`  use both of leveldb's memdb and leveldb's filedb can be optional but Merging Two GetIterator is needed. 

In general,  IMHO, `LevelDB.Batch` is a better approach than using `LevelDB.Tx` directly too. Because of LevelDB.Batch decides using its own tx by batch size to prevent long time lock time if what I know is correct. (It needs checking the advantage is right)

Another approach for `StateDB` is https://github.com/bosnet/sebak/blob/contract_stateclone/lib/state_clone.go#L148

IMHO, This approach makes easier to handling state objects, but It breaks current DB handling codes in sebak. 

And I realized the design difference between the `stateclone` and `statedb`. IMHO, The `stateclone` is a [DataMapper] and `statedb` is a [ActiveRecord] (Caution: I am not a pattern believer)

[DataMapper] : https://martinfowler.com/eaaCatalog/dataMapper.html
[ActiveRecord] : https://martinfowler.com/eaaCatalog/activeRecord.html

Another reference about this issue may be this: https://gokit.io/faq/#persistence-mdash-how-should-i-work-with-databases-and-data-stores

@kfangw , @Charleslee522 , @spikeekips  What do you think about it? I need your wisdoms and opinions ㅠ_ㅠ

spikeekips:
> @anarcher: Because LevelDB's tx make a One Global Lock and is expensive and doesn't allow multi transactions (even leveldb.DB (Not Tx struct) cannot do too, it makes sense) if what I know is correct. (Is it right?)

I didn't find any sentences LevelDB does not support multiple transaction from one main db instance. It would be possible. In the current SEBAK, the `LevelDBBackend.OpenTransction()` also open multiple transaction from one `LevelDBBackend` instance.

spikeekips:
@anarcher Hmmm, that means, LevelDB does not support multiple `Opentransaction` from one main instance. The `st.New` should be blocked until `ts` is finished.

PR #240, opened 2018-08-13
fixed bugs from latest new-proposer

### LINKED ISSUE #216: The new ISAAC voting process(SIGN, ACCEPT)
### Voting Process
Originally the new model will have 2 voting process,

* `SIGN`: voting for received Ballot from proposer
* `ACCEPT`: voting for confirmed Ballot of each validator

In `ACCEPT` state, validator will compare the voting result of the other validators with his/her voting result.

In #215, `SIGN` and `ACCEPT` voting process is not yet implemented. Therefore, it will be completed in with this issue :)

spikeekips:
### Github Issue
resolve #216 

### Background
The latest commit, ab675b4428e57fafb10fd2645bc3654d8d654255 of new-proposer has some bugs.

* `cmd/sebak/cmd/run.go` does not have `SIGN` and `ACCEPT` threshold.
* `CheckNodeRunnerHandleACCEPTBallotValidateResult` is duplicated with `CheckNodeRunnerHandleBallotCheckResult`.
* `checker.FinishedVotingHole` in `checker.NodeRunner.Consensus().CloseConsensus()` 
* `checkTransactionSourceCheck` moved to `Transaction.Validate()`
* some test code fixed.

I tried to fix them. And further more, added documentations for functions in `node_runner_checker.go` and `node_runner_transaction_checker.go`.


Geod24:
"Fix bug" is not very descriptive...

spikeekips:
@Geod24 I listed some bugs in description. Thanks.

lib/transaction.go
```diff
@@ -108,11 +108,51 @@ func (tx Transaction) IsWellFormed(networkID []byte) (err error) {
 	return
 }
 
+// Validate checks,
+// * source account exists
+// * checkpoint is valid
+// * source has enough balance to pay
+// * and it's `Operations`
```
Charleslee522: 👍

cmd/sebak/cmd/run.go
```diff
@@ -44,19 +44,21 @@ var (
 	flagTLSCertFile         string = sebakcommon.GetENVValue("SEBAK_TLS_CERT", "sebak.crt")
 	flagTLSKeyFile          string = sebakcommon.GetENVValue("SEBAK_TLS_KEY", "sebak.key")
 	flagValidators          string = sebakcommon.GetENVValue("SEBAK_VALIDATORS", "")
-	flagThreshold           string = sebakcommon.GetENVValue("SEBAK_THRESHOLD", "66")
+	flagSIGNThreshold       string = sebakcommon.GetENVValue("SEBAK_SIGN_THRESHOLD", "66")
+	flagACCEPTThreshold     string = sebakcommon.GetENVValue("SEBAK_ACCEPT_THRESHOLD", "66")
```
Geod24: Isn't that a revert to the state in master ? Why do we have 2 different thresholds ?

lib/storage/dbbackend.go
```diff
@@ -1,17 +0,0 @@
...
-	Remove(string) error
-
-	GetIterator(prefix string, reverse bool) (func() (IterItem, bool), func())
-
-	News(...Item) error
-	Sets(...Item) error
-
-	Commit() error
-	Discard() error
-}
```
Geod24: Why do you remove those ? Looks unrelated

lib/transaction.go
```diff
@@ -108,11 +108,51 @@ func (tx Transaction) IsWellFormed(networkID []byte) (err error) {
...
+	totalAmount := tx.TotalAmount(true)
+
+	// check, have enough balance at checkpoint
+	if sebakcommon.MustAmountFromString(bac.Balance) < totalAmount {
+		err = sebakerror.ErrorTransactionExcessAbilityToPay
+		return
+	}
+
+	// check, have enough balance now
+	if sebakcommon.MustAmountFromString(ba.Balance) < totalAmount {
```
Geod24: This is the same as the previous check, it does not check for left balance

lib/transaction.go
```diff
@@ -108,11 +108,51 @@ func (tx Transaction) IsWellFormed(networkID []byte) (err error) {
...
+		err = sebakerror.ErrorTransactionExcessAbilityToPay
+		return
+	}
+
+	for _, op := range tx.B.Operations {
```
Geod24: I would say move this before the balance check, because if any operation is invalid `TotalAmount` does not make sense / could error out.

cmd/sebak/cmd/run.go
```diff
@@ -44,19 +44,21 @@ var (
...
-	storageConfig *sebakstorage.Config
-	validators    []*sebaknode.Validator
-	threshold     int
-	logLevel      logging.Lvl
-	log           logging.Logger
+	kp              *keypair.Full
+	nodeEndpoint    *sebakcommon.Endpoint
+	storageConfig   *sebakstorage.Config
+	validators      []*sebaknode.Validator
+	signThreshold   int
```
cmcm2222: Do we need 2 threshold paramater??

cmd/sebak/cmd/run.go
```diff
@@ -194,8 +197,11 @@ func parseFlagsNode() {
 		common.PrintFlagsError(nodeCmd, "--storage", err)
 	}
 
-	if threshold, err = strconv.Atoi(flagThreshold); err != nil {
-		common.PrintFlagsError(nodeCmd, "--threshold", err)
+	if signThreshold, err = strconv.Atoi(flagSIGNThreshold); err != nil {
```
cmcm2222: Same as previous question.

spikeekips: @cmcm2222 As I understand our new proposer model, the voting will be done in 2 state transition, `INIT` -> `SIGN` and `SIGN` -> `ACCEPT`.

lib/storage/dbbackend.go
spikeekips: @Geod24 Oops, it's mistake 👍

cmd/sebak/cmd/run.go
spikeekips: https://github.com/bosnet/sebak/pull/240#discussion_r209826426

lib/transaction.go
spikeekips: @Geod24 We can choose the proper position of this block. Anyway `Operation.Validate` is more expensive than checking `TotalAmount`.

cmd/sebak/cmd/run.go
Geod24: But why would you want a different threshold ? That sounds quite dangerous.

spikeekips:
> @Geod24 : The integration tests tells me this is not ready.

Yes. I can fix it.

cmd/sebak/cmd/run.go
spikeekips: @Geod24 Hmm, it would be better node operator to set the threshold, but if it can be matter, how about limit the minimum threshold like 66%.

spikeekips: @Geod24 So do you think the threshold option will be useless?

Geod24: At the moment, I don't see any reason for it to be configurable. In other words, [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

spikeekips: @Geod24 If you mean only for SEBAK, the current threshold option is not important, it can be set in inside variable. :)

spikeekips:
The integration test failed because it was too short until consensus. For more details, see 36be761 . This commit maybe need the more additional patches :)

Geod24:
Okay, but the test was waiting 5 seconds. How long does it take to reach consensus now ?

spikeekips:
> @Geod24 : Okay, but the test was waiting 5 seconds. How long does it take to reach consensus now ?

At this time, we can not convince how long does it takes, but in normal env I guess 5 ~10 seconds. The env of integration test does not reflect the real env, so `getAccountWithBalance()` will wait almost 10 seconds.

cmd/sebak/cmd/run.go
Geod24: My point is that we should just hardcode it to 67% for the time being and revise that when this kind of tweaks is needed (when we open the network).

spikeekips: @Geod24 Yes, I agree. It can be hardcoded. If I have a right to decision, I will leave the threshold related options, it will be useful to control or test.

spikeekips:
@Geod24 threshold part is ancient code, so we can decide to keep or remove at anytime later. Please review the rest of part.

spikeekips:
@Geod24 Yes, same threshold will be reasonable, I agree. We can choose later. At this time we will set the 2 threshold value for SIGN and ACCEPT in cmd options.

spikeekips:
@Geod24 Ping a316fbbfae7393bef7c462de0a3c9d6b94ac6297

lib/node_runner.go
```diff
@@ -367,27 +290,21 @@ func (nr *NodeRunner) handleBallotMessage(message sebaknetwork.Message) (err err
...
 		VotingHole:     VotingNOTYET,
 	}
-	err = sebakcommon.RunChecker(baseChecker, nr.handleBaseBallotCheckerDeferFunc)
+	err = sebakcommon.RunChecker(baseChecker, nr.handleMessageCheckerDeferFunc)
 	if err != nil {
 		if _, ok := err.(sebakcommon.CheckerErrorStop); !ok {
-			nr.log.Debug("failed to handle ballot", "error", err, "state", "base")
 			return
 		}
-		nr.log.Debug("stop handling ballot", "reason", err, "state", "base")
```
Geod24: Why are those removed ? And why were they `Debug` to begin with ?

lib/node_runner.go
```diff
@@ -325,16 +250,15 @@ func (nr *NodeRunner) handleMessage() {
 
 		if err != nil {
 			if _, ok := err.(sebakcommon.CheckerErrorStop); ok {
-				nr.log.Debug("consensus finished", "message", message.Head(50), "error", err)
 				continue
 			}
-			nr.log.Debug("failed to handle sebaknetwork.Message", "message", message.Head(50), "error", err)
+			nr.log.Debug("stop handling sebaknetwork.Message", "message", message.Head(50), "error", err)
```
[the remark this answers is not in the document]
spikeekips: @Geod24 Yes, you right. "stop" is not proper in this context.

lib/node_runner.go
```diff
@@ -325,16 +250,15 @@ func (nr *NodeRunner) handleMessage() {
...
 		}
 	}
 }
 
 func (nr *NodeRunner) handleMessageFromClient(message sebaknetwork.Message) (err error) {
-	nr.log.Debug("got message from client`", "message", message.Head(50))
+	nr.log.Debug("got message`", "message", message.Head(50))
```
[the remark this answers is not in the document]
spikeekips: @Geod24 I agree, we can also remove 'from client' things from `sebaknetwork.MessageType`. I think we can handle this issue in another issue.

lib/node_runner.go
```diff
@@ -441,7 +356,7 @@ func (nr *NodeRunner) InitRound() {
 		}
 	}
 
-	nr.log.Debug("all validators are checked for connectivity")
+	nr.log.Debug("connectivity checked")
```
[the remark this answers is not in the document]
spikeekips: @Geod24 Good. I will change it.

lib/node_runner.go
spikeekips: @Geod24 Oops, it's mistakes including `Debug`. I'll recover them.

lib/node_runner.go
```diff
@@ -167,146 +202,36 @@ func (nr *NodeRunner) ConnectValidators() {
...
+	nr.handleMessageCheckerDeferFunc = f
 }
 
 func (nr *NodeRunner) handleMessage() {
 	for message := range nr.network.ReceiveMessage() {
 		var err error
 
 		if message.IsEmpty() {
-			nr.log.Error("got empty message`")
+			nr.log.Debug("got empty message`")
```
Geod24: Could you explain why this was changed ?

spikeekips: @Geod24 Oops, it should be `Error`.

lib/node_runner.go
```diff
@@ -344,11 +268,10 @@ func (nr *NodeRunner) handleMessageFromClient(message sebaknetwork.Message) (err
...
 	}
 
-	if err = sebakcommon.RunChecker(checker, nr.handleMessageFromClientCheckerDeferFunc); err != nil {
-		if _, ok := err.(sebakcommon.CheckerErrorStop); ok {
-			return
+	if err = sebakcommon.RunChecker(checker, nr.handleMessageCheckerDeferFunc); err != nil {
+		if _, ok := err.(sebakcommon.CheckerErrorStop); !ok {
+			nr.log.Error("failed to handle message from client", "error", err)
 		}
-		nr.log.Error("failed to handle message from client", "error", err)
```
Geod24: So if it's another kind of error than `CheckerErrorStop` we are not logging it ?

spikeekips: @Geod24 This will `log.Error` except `sebakcommon.CheckerErrorStop`.

Geod24: Oh right I misread

spikeekips:
@Geod24 Thanks.

PR #241, opened 2018-08-13
Storing 'NIL' `Block` even when empty transactions

### LINKED ISSUE #237: If the transaction pool is empty, then the proposer proposes nil block
The `nil block` is a block with no transaction.

The `SEBAK` network must confirm the new block periodically(about 5 sec). Because the block confirmation is related with the freezing reward. The freezing reward must be stored constantly based on the block height.
Therefore even if the tx pool is empty, the `nil block` must be confirmed.

spikeekips:
### Github Issue
resolves #237 

### Background

This PR should be merged after #240 , so please review the comparison, https://github.com/bosnet/sebak/compare/c4be6c4fd7494e85d8ef74942a38f15d77d7b0f6...spikeekips:new-proposer-nil-block .


lib/node_runner_checker.go
```diff
@@ -412,27 +422,19 @@ func CheckNodeRunnerHandleACCEPTBallotStore(c sebakcommon.Checker, args ...inter
...
-				checker.Ballot,
-				checker.NodeRunner.Consensus().TransactionPool,
-			)
-			if err != nil {
-				return
-			}
-
-			checker.NodeRunner.Consensus().SetLatestConsensusedBlock(block)
-			checker.Log.Debug("ballot was stored", "block", block)
+		var block Block
```
cmcm2222: Looking good.

cmd/sebak/cmd/run.go
```diff
@@ -87,8 +87,11 @@ func init() {
 
 			parseFlagsNode()
 
-			runNode()
-			return
+			var exitCode int
+			if err = runNode(); err != nil {
+				exitCode = 1
+			}
+			os.Exit(exitCode)
```
Geod24:
```go
if err = runNode(); err != nil {
  os.Exit(1)
}
os.Exit(0)

File: cmd/sebak/cmd/run.go [hunk 3]
```diff
@@ -249,37 +252,38 @@ func parseFlagsNode() {
...
 	localNode.AddValidators(validators...)
 
 	// create network
 	nt, err := sebaknetwork.NewNetwork(nodeEndpoint)
 	if err != nil {
-		log.Crit("transport error", "error", err)
-
-		os.Exit(1)
+		log.Crit("failed to create Network", "error", err)
+		return err
```
Geod24: 👍

cmd/sebak/cmd/run.go
spikeekips: @Geod24 I want to open the possibility to set the another exit code like http://www.tldp.org/LDP/abs/html/exitcodes.html .

Geod24: Then the exit code should be returned from the function and used directly, no ?

spikeekips: @Geod24 The way to set the exit code, we can discuss. Checking error type, or `runNode()` return the exit code explicitly instead of error. Anything can be possible. :)

Geod24: You're right, it's not a big deal, let's leave it as is

Geod24:
This actually broke the integration test.
For coverage to work, we should always exit through the normal runtime path (which contains the cleanup functions), not hard-`exit(xxx)`.
Please, next time the tests are failing, make sure they are fixed before moving forward with more code as it means more things to fix later.

Issue #209, opened 2018-08-02
The limitation of the number of transactions in one proposed block


Charleslee522:
In the current version of the new ISAAC, there is no limitation of the number of transactions in one block.
The whole transactions which is gathered in 2 seconds(It can be changed) will be included into the block. So, I guess, the limitation is needed.

Charleslee522:
@cmcm2222 Exactly. From now, we have transaction pool and propose block with whole txs in the pool. To resolve this issue, picking top n txs in transaction pool is needed. I guess txs needs to be ordered by received time.

Charleslee522:
It had been implemented in https://github.com/bosnet/sebak/pull/215/files#diff-e33f7f460f666841210e11844c21d6abR21

spikeekips:
With this [`MaxTransactionsInBallot`](https://github.com/spikeekips/sebak/blob/new-proposer/lib/node_runner.go#L28), we need to check the number of transactions in the proposed ballot is over it.

Issue #201, opened 2018-08-01
Design documentation of the new ISAAC


Charleslee522:
Document for explaining design of the new ISAAC consensus protocol

cmcm2222:
I think yesterday you showed to us. right??

Charleslee522:
yes but we need to improve the document :)

cmcm2222:
@Charleslee522  Do you have any ideas for doc modifying?
Or If possible, you can give me simple flow diagram that wrote in hand. then I will do.

sconeman:
@cmcm2222 @Charleslee522 
@spikeekips has changed a lot. Maybe we should listen to his explanation.

Charleslee522:
@cmcm2222 @spikeekips @spikeekips @Geod24 @sconeman @kyehyukahn 
Please review this page https://github.com/bosnet/sebak/wiki/How-the-consensus-protocol-works and feel free to change it.

sconeman:
@Charleslee522 So Does ISAAC have timers by the ballot state?

Charleslee522:
@sconeman It should be, but not yet. Only TimeoutExpireRound(10 sec) is now.

sconeman:
@Charleslee522 In `ACCEPT` → `ALL-CONFIRM` chapter,
> 1. Each node checks the block is valid.
I think it is more good that this part is moved to `SIGN`  → `ALL-CONFIRM` chapter.
Because the description of `SIGN` is below.
> Checked that the proposed block is valid by itself

Charleslee522:
@sconeman You're right. Thanks :)

Charleslee522:
@cmcm2222 @spikeekips @spikeekips @Geod24 @sconeman @kyehyukahn 
I changed [design doc](https://github.com/bosnet/sebak/wiki/How-the-consensus-protocol-works), does it make sence?
@spikeekips How about voting without voting hole? I can assume the design that each node only votes when it agree with the proposed block.

Charleslee522:
@spikeekips 
@Geod24 has a question. Why does a block have only valid transactions? Let's talk about it in this issue by comment :)

spikeekips:
> @Charleslee522 : @spikeekips How about voting without voting hole? I can assume the design that each node only votes when it agree with the proposed block.

[...]

PR #243, opened 2018-08-13
Limit the maximum number of Transactions in one Ballot

spikeekips:
### Github Issue

#209 

### Background

The main issue is #209 and added to check `MaxOperationsInTransaction`, means it limits the maximum number of Operations in one Transaction.

This PR should be merged after #240, so you can review the comparison, https://github.com/bosnet/sebak/compare/c4be6c4fd7494e85d8ef74942a38f15d77d7b0f6...spikeekips:new-proposer-MaxTransactionsInBallot .


lib/error/errors.go
```diff
@@ -1,47 +1,49 @@
...
+	ErrorBlockOperationDoesNotExists          = NewError(135, "operation does not exists in block")
+	ErrorRoundVoteNotFound                    = NewError(136, "`RoundVote` not found")
+	ErrorBlockNotFound                        = NewError(137, "Block not found")
+	ErrorTransactionExcessAbilityToPay        = NewError(138, "Transaction requests over abiltiy to pay")
+	ErrorTransactionSameSource                = NewError(139, "Same transaction source found in ballot")
+	ErrorTransactionNotFound                  = NewError(140, "Transaction not found")
+	ErrorBallotFromUnknownValidator           = NewError(141, "ballot from unknown validator")
+	ErrorBallotAlreadyFinished                = NewError(142, "ballot already finished")
+	ErrorBallotAlreadyVoted                   = NewError(143, "ballot already voted")
+	ErrorBallotHasOverMaxTransactionsInBallot = NewError(144, "ballot has too many transactions")
```
Charleslee522: Like "too many operations in transaction" as follow, how about "too many transactions in a ballot"? It's just advisory not mendatory :)

lib/transaction_test.go
```diff
@@ -121,3 +119,34 @@ func TestTransactionIsValidCheckpoint(t *testing.T) {
 		return
 	}
 }
+
+func TestTransactionMaxOperationsInTransaction(t *testing.T) {
```
Charleslee522: 👍

lib/transaction_test.go
```diff
@@ -121,3 +120,34 @@ func TestTransactionIsValidCheckpoint(t *testing.T) {
...
+	var err error
+
+	// well-formed
```
cmcm2222: I always confused about this `well-formed` sign. 
If anyone could give me some understand this `well-formed` mean to be? 
it point that well marshaled? or message itself well formatted?

Geod24: Well formed means passing contextless validation (e.g. doesn't have too many transaction, has valid address, amount, signatures...)
Checking things like account balance and so on require context so they are not included.
This difference is important because a transaction which is `well-formed` can be invalid and then valid (or the other way around), but a transaction which is not `well-formed` can never be valid and thus can be discarded immediately.

spikeekips: @cmcm2222 Please see https://github.com/bosnet/sebak/issues/201#issuecomment-411670432 .

cmcm2222: @spikeekips @Geod24  thank you for kindness. :)

spikeekips:
@Geod24 Please continue review :)

lib/ballot_test.go
```diff
@@ -0,0 +1,39 @@
...
+	var txs []string
+	for i := 0; i < MaxTransactionsInBallot+1; i++ {
+		_, tx := TestMakeTransaction(networkID, 1)
+		txs = append(txs, tx.GetHash())
+	}
+
+	ballot = NewBallot(node, round, txs)
+	ballot.Sign(node.Keypair(), networkID)
+
+	if err := ballot.IsWellFormed(networkID); err == nil {
```
Geod24:
```go
err := ballot.IsWellFormed(networkID)
require.Nil(t, err)
```

lib/ballot_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	MaxTransactionsInBallot = 2
+
+	_, node := createNetMemoryNetwork()
+	round := Round{Number: 0, BlockHeight: 1, BlockHash: "hahaha", TotalTxs: 1}
+	_, tx := TestMakeTransaction(networkID, 1)
+
+	ballot := NewBallot(node, round, []string{tx.GetHash()})
+	ballot.Sign(node.Keypair(), networkID)
+	if err := ballot.IsWellFormed(networkID); err != nil {
```
Geod24: Ditto `require.Nil(t, err)`

lib/constant.go
```diff
@@ -1,9 +1,37 @@
...
 	// BaseFee is the default transaction fee, if fee is lower than BaseFee, the
 	// transaction will fail validation.
 	BaseFee sebakcommon.Amount = 10000
 )
+
+var (
+	// TimeoutProposeNewBallot works when the consensus process is finished for
+	// a proposed `Ballot`, proposer will wait for `TimeoutProposeNewBallot` and
+	// then it will propose new `Ballot`.
+	TimeoutProposeNewBallot time.Duration = time.Second * 2
```
Geod24: That's the maximum time right ?

lib/constant.go
```diff
@@ -1,9 +1,37 @@
...
+	TimeoutExpireRound time.Duration = time.Second * 10
+
+	// MaxTransactionsInBallot limits the maximum number of `Transaction`s in
+	// one proposed `Ballot`.
+	MaxTransactionsInBallot int = 1000
+
+	// MaxOperationsInTransaction limits the maximum number of `Operation`s in
+	// one `Transaction`.
+	MaxOperationsInTransaction int = 200
+)
```
Geod24: Those refactoring, while welcome, would have been better in a separate commit.
The point is that if we want to track down a problem (via `git bisect`) or fix one (`git revert`), we don't want to have to deal with unrelated / independent changes which are worthwhile.

spikeekips: @Geod24 OK, good point. We can handle `MaxOperationsInTransaction` in new issue.

lib/constant.go
spikeekips: @Geod24 not max. Proposer will wait to propose new ballot for `TimeoutProposeNewBallot` or `TimeoutProposeNewBallotFull`.

lib/ballot_test.go
spikeekips: @Geod24 ❤️

lib/error/errors.go
```diff
@@ -1,47 +1,49 @@
...
+	ErrorRoundVoteNotFound                    = NewError(136, "`RoundVote` not found")
+	ErrorBlockNotFound                        = NewError(137, "Block not found")
+	ErrorTransactionExcessAbilityToPay        = NewError(138, "Transaction requests over abiltiy to pay")
+	ErrorTransactionSameSource                = NewError(139, "Same transaction source found in ballot")
+	ErrorTransactionNotFound                  = NewError(140, "Transaction not found")
+	ErrorBallotFromUnknownValidator           = NewError(141, "ballot from unknown validator")
+	ErrorBallotAlreadyFinished                = NewError(142, "ballot already finished")
+	ErrorBallotAlreadyVoted                   = NewError(143, "ballot already voted")
+	ErrorBallotHasOverMaxTransactionsInBallot = NewError(144, "too many transactions in ballot")
+	ErrorTooManyOperations                    = NewError(145, "too many operations in transaction")
```
Geod24: Ditto refactoring: I would PR those changes to existing code to `master` as it doesn't depend on anything else and is good on its own.

lib/node_runner.go
```diff
@@ -231,7 +224,7 @@ func (nr *NodeRunner) handleMessage() {
 		var err error
 
 		if message.IsEmpty() {
-			nr.log.Error("got empty message`")
+			nr.log.Debug("got empty message`")
```
Geod24: Leftover ?

spikeekips: @Geod24 Sorry :(

lib/constant.go
Geod24: I thought this had to be adjusted to prevent nodes round timing drifting ?

spikeekips: @Geod24 This PR is based the previous design of ISAAC, so the new ISAAC concept will be different.

lib/error/errors.go
spikeekips: @Geod24 The last one is unused, I will remove it.

spikeekips:
@Geod24 This PR also getting old, I will merge it if you don't mind.

Issue #214, opened 2018-08-06
Some performance experiments


Charleslee522:
I did an experience with 
https://github.com/spikeekips/sebak/blob/new-proposer/lib/common/util_benchmark_test.go
The goal is,
1. to find out which is faster, `string slice` or `string map` when finding a string.
2. to determine which is faster, `string slice` or `string map` or `string map with hashing` when checking equality.

> $ go test ./lib/common/ -bench=.
```
goos: darwin
goarch: amd64
pkg: boscoin.io/sebak/lib/common
BenchmarkFoundInStringSlice500-8      	 2000000	       955 ns/op
BenchmarkFoundInStringMap500-8        	20000000	        72.2 ns/op
BenchmarkFoundInStringSlice1000-8     	 1000000	      1791 ns/op
BenchmarkFoundInStringMap1000-8       	20000000	        76.6 ns/op
BenchmarkFoundInStringSlice5000-8     	  200000	      9196 ns/op
BenchmarkFoundInStringMap5000-8       	20000000	        87.8 ns/op
BenchmarkEqualityOrderedSlice500-8    	 1000000	      1372 ns/op
BenchmarkEqualityMap500-8             	  100000	     14249 ns/op
BenchmarkEqualityMapWithHash500-8     	 1000000	      1747 ns/op
BenchmarkEqualityOrderedSlice1000-8   	  500000	      2723 ns/op
BenchmarkEqualityMap1000-8            	   50000	     29616 ns/op
BenchmarkEqualityMapWithHash1000-8    	 1000000	      1751 ns/op
BenchmarkEqualityOrderedSlice5000-8   	  100000	     14079 ns/op
BenchmarkEqualityMap5000-8            	   10000	    157504 ns/op
BenchmarkEqualityMapWithHash5000-8    	 1000000	      1764 ns/op
```

Result
> 1. to find out which is faster, `string slice` or `string map` when finding a string.

The string map is faster than string slice. As big o notation, string map is O(log n) and string slice is O(n). So I changed string slice to string map in https://github.com/spikeekips/sebak/blob/new-proposer/lib/ballot.go#L143

> 2. to determine which is faster, `string slice` or `string map` or `string map with hashing` when checking equality.

The `string map with hashing` is the fastest. As big o notation, `string map with hashing` is O(1) but, `string slice` and `string map` are O(n).

Charleslee522:
@spikeekips Please check when you review my commits.

Charleslee522:
[...]

PR #244, opened 2018-08-13
Only valid Transactions into TransactionPool

spikeekips:
### Background

Under our new proposer-based consensus model, the growing `TransactionPool` is the new issue. To keep clean the `TransactionPool`, the 1st method is that only valid transactions will be pushed into `TransactionPool`. The missing Transactions during consensus will be fetched from proposer(#212). 

This PR is also based on `new-proposer`, so after that this also can be merged.


lib/node_runner.go
```diff
@@ -548,15 +549,21 @@ func (nr *NodeRunner) proposeNewBallot(roundNumber uint64) error {
...
-			}
+	err := sebakcommon.RunChecker(transactionsChecker, sebakcommon.DefaultDeferFunc)
+	if err != nil {
+		return err
+	}
+
+	// remove invalid transactions
+	var invalidTransactions []string
+	for _, hash := range availableTransactions {
+		if _, found := sebakcommon.InStringArray(transactionsChecker.ValidTransactions, hash); found {
```
Charleslee522: This part is crucial, right? Because this function is called every block height. So we should focus on the performance.
I figure it would be better to use transactionsChecker.ValidTransactionsMap instead of transactionsChecker.ValidTransactions because of #214 :)

lib/node_runner_transaction_checker.go
```diff
@@ -15,13 +15,42 @@ type NodeRunnerHandleTransactionChecker struct {
...
+	return
+}
+
+func (checker *NodeRunnerHandleBallotTransactionChecker) ValidTransactions() []string {
+	return checker.validTransactions
+}
+
+func (checker *NodeRunnerHandleBallotTransactionChecker) InvalidTransactions() (invalids []string) {
+	for _, hash := range checker.Transactions {
+		if _, found := checker.ValidTransactionsMap[hash]; found {
```
Charleslee522: 👍

lib/node_runner.go
```diff
@@ -231,7 +231,7 @@ func (nr *NodeRunner) handleMessage() {
 		var err error
 
 		if message.IsEmpty() {
-			nr.log.Error("got empty message`")
+			nr.log.Debug("got empty message`")
```
Geod24: Leftover

spikeekips:
@Geod24 I added testing, please continue review :)

lib/common/net.go
```diff
@@ -7,15 +7,46 @@ import (
...
+	}
+	_, err := net.DialTimeout(
+		"tcp",
+		net.JoinHostPort("", strconv.FormatInt(int64(port), 10)),
+		100*time.Millisecond,
+	)
 	return err
 }
 
+func GetFreePort(excludes ...int) (port int) {
```
Geod24: That bit bothers me a lot:
- It's test utility, not production code, and should be in a test file (`lib/test.go`)
- It checks if a port is available by doing a TCP request to it (!!!)
- It's called in a loop 

A much saner approach would be to bind the servers to port `0` and then get the actual port number for the client to use. Even better would be not to go through the network stack for unittests, but we're unfortunately quite far away from that.

lib/node_runner_test.go
```diff
@@ -104,6 +113,105 @@ func createTestNodeRunnerWithReady(n int) []*NodeRunner {
 	return nodeRunners
 }
 
+func createTestNodeRunnersHTTP2Network(n int) (nodeRunners []*NodeRunner, rootKP *keypair.Full) {
```
Geod24: That kind of utility should probably be in `lib/test.go`

lib/node_runner_test.go
```diff
@@ -17,10 +19,18 @@ import (
...
 	kp      *keypair.Full
 	account *block.BlockAccount
+	tlsKey  *sebaknetwork.KeyGenerator
 )
 
 func init() {
 	kp, _ = keypair.Random()
+
+	dir, err := ioutil.TempDir("/tmp/", "sebak-test")
+	if err != nil {
```
Geod24: `require.Nil` ;)

lib/common/net.go
spikeekips: @Geod24 `lib/test.go` is not good place for these funcs, but there are some funcs, which are used in several packages. If you have good way for that, please let me know :)

I know using `0` port, we can replace it. Recently I think, we should test with `HTTP2Network` instead of `MemoryNetwork` because `MemoryNetwork` is used only for testing. Without binding 'real' port in `HTTP2Network`, we can consider the fake http.

lib/node_runner_test.go
spikeekips: @Geod24 This block is not in part of testing. :)

lib/node_runner_test.go
spikeekips: @Geod24 `createTestNodeRunner` exists in the same file, so it looks natural `createTestNodeRunnersHTTP2Network` in here. We can move to `lib/test.go` in next time. :)

spikeekips:
@Geod24 huffing and puffing :)

spikeekips:
@Geod24 I think this PR is getting old :), so if you don't mind, I will merge it.

PR #245, opened 2018-08-13
change statedb to use stateobject

anarcher:
- Add StateDB.BatchWrite (TODO: a Better name for it) it separates from Commit. Do we need a method of BatchWrite+Commit ?
- Doesn't support GetIterator in this PR...

@kfangw This approach works for the integration?


kfangw:
What I thinking of the stateObject is like this 👍 
But I want to apply the concept to the master branch.
I am trying to combine block_account and state object together which is based on ethereum's concept.
Please wait for my PR and pleeeeaaaaaase have your vacation :)
Let's talk about this on Thursday in depth.

PR #249, opened 2018-08-14
Make new-proposer compile

Geod24:
@spikeekips : This is the bare minimum required to make `new-proposer` compile.
I suggest we merge this and then discuss if we need a second threshold (or a configurable threshold at all) so that people can work on the branch. What do you think ?

### DIFF
--- a/cmd/sebak/cmd/run.go
+++ b/cmd/sebak/cmd/run.go
@@ -266,7 +266,7 @@ func runNode() {
 		os.Exit(1)
 	}
 
-	policy, _ := sebak.NewDefaultVotingThresholdPolicy(threshold)
+	policy, _ := sebak.NewDefaultVotingThresholdPolicy(threshold, threshold)
 	policy.SetValidators(len(localNode.GetValidators()) + 1) // including 'self'
 
 	isaac, err := sebak.NewISAAC([]byte(flagNetworkID), localNode, policy)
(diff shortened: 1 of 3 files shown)


Geod24:
Alright we need https://github.com/bosnet/sebak/pull/248/commits/36be761f2587679d8dde98377f396ce8fb3e1dce for tests to pass, which is why I recommended to put it in master (it makes sense on its own anyway).

spikeekips:
@Geod24 Except 36be761, we already started review for #240. I think we can wrap up ASAP including threshold(s) :) Anyway I will submit new PR for 36be761.

spikeekips:
#253 submitted.

Geod24:
@spikeekips : Except that there is some discussion happening in #240 
I don't think we should tie the outcome of those discussions to the ability of our whole team to work on new-proposer when we have such a trivial fix easily available.

spikeekips:
@Geod24 I don't understand why you make this PR instead of trying to finish #240. As you know, #240  is the first and main PR, `new-propser` branch is just branch. If we can finish #240, the `new-proposer-bug-fixes` will be merged, not `new-propser`.

Geod24:
@spikeekips : The point was not to block people on the outcome of #240 
Since the second threshold option is controversial, I suggested to fix the current bugs but keep the same behavior, and not block it on the additional code you want to introduce.

Geod24:
#240 was merged

PR #252, opened 2018-08-14
Statedb with trie

kfangw:
### Github Issue
#224 

### Background
As-is consensus protocol based on the `transactions` not `result of execution of transactions`
Hence, the transactions result of each node may diversity due to not intact or bug.
It raises lazy discovery of the problem. 

### Solution
Switch the target of consensus from `transactions` to `result of execution of transactions`
To achieve this, `world state` is needed.

Like ethereum, `world state` is a set of `state of an account`.
In the implementation, `world state` is a Merkle root of Merkle tree. Each state of an account is live in a leaf node of the tree. When the state is changed then the Merkle root also changed. And an account state is reachable by traversing the tree from the Merkle root.

### Possible Drawbacks
There are alternatives for the implementation of the `world state`
The first one is ethereum's Merkle Patricia Trie
The second one is tendermint's IAVL
The third one is a brand new implementation.

In the beginning, I write stateDB based on ethereum's MPT.
This may be changed to other implementation.


kfangw:
@spikeekips @anarcher 
Please review this change.

This change does not affect current behavior of SEBAK. Only the feature is added.
I will make another PR that change BlockAccount to StateDB. #266 
@anarcher Your improvement of the stateDB may be based on this branch which can be merged to master of sebak.

PR #253, opened 2018-08-14
Integration Tests: Use retry and sleep for balance check

spikeekips:
### Background

Recent PRs was failed to pass the integration test. It was caused that it does take more time as we expected(5 seconds) to finish the consensus of testing transaction.

> Don't complain this PR's name :)

### Solution
At first I increased the sleep time to 10 seconds and everything was good, but the matter was we can not convince the exact(or maximum) elapsed time to finish one transaction. This PR does not guess the maximum elapsed time, just checking the latest account information up to almost 10 seconds.

### DIFF
--- a/tests/utils.sh
+++ b/tests/utils.sh
@@ -29,11 +29,7 @@ function getAccountWithBalance () # u16 port, string addr and expected balance
     balance=0
     for i in $(seq 30)
     do
-	    balance=$(curl --insecure \
-                  --request GET \
-                  --header "Accept: application/json" \
-                  https://127.0.0.1:${1}/api/account/${2} \
-                  2>/dev/null | jq ".Balance" | sed 's/"//g')
+	    balance=$(getAccount $1 $2 | jq ".Balance" | sed 's/"//g')
 	    if [ "$balance" == "$3" ];then
 		    return 0
 	    fi
--- a/tests/utils.sh
+++ b/tests/utils.sh
@@ -19,3 +19,28 @@ function getAccount () # u16 port, string addr
                   https://127.0.0.1:${1}/api/account/${2} \
                   2>/dev/null)
 }
+
+function getAccountWithBalance () # u16 port, string addr and expected balance
+{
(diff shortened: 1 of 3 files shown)


tests/utils.sh
```diff
@@ -19,3 +19,24 @@ function getAccount () # u16 port, string addr
...
+    fi
+
+    balance=0
+    for i in $(seq 30)
+    do
+	    balance=$(getAccount $1 $2 | jq ".Balance" | sed 's/"//g')
+	    if [ "$balance" == "$3" ];then
+		    return 0
+	    fi
+	    sleep 0.3
```
Geod24: According to `man 1 sleep`:
screen shot 2018-08-16 at 09 14 34
So I'd rather have a `sleep 1` here.

Geod24:
Also, could you provide a more descriptive title for your commit / PR ?
Something that allows people watching this repository to know, from the title, what the PR is about.
For example "Implement a retry and sleep mechanism for integration tests' balance check" or "Make integration test balance check more fine-grained and reliable".

tests/utils.sh
spikeekips: @Geod24 As you said, it is not bottleneck of this test, but there are subtle difference as a result.

spikeekips:
> @Geod24 : Also, could you provide a more descriptive title for your commit / PR ?

I agree, but the title of PR is limited to explain everything, I think the current title will be good. If you don't mind, please suggest new title.

tests/utils.sh
Geod24: What differences ?

spikeekips: @Geod24 Finish fast?! :)

Geod24:
What about "Tests: Use retry and sleep for balance check" ?
Tells the reader what the change is about (tests' balance check) and what is (roughly) done. And it's under 50 chars.

spikeekips:
@Geod24 Deal. :)

PR #258, opened 2018-08-17
refactor:move block_account to block package

kfangw:
`block_account` can be live in another package not in the `sebak` package


kfangw:
@spikeekips @Charleslee522 Could you confirm this refactoring asap?
It is a very simple `move` refactoring.

lib/block/block_account.go
```diff
@@ -1,4 +1,4 @@
-package sebak
+package block
```
Geod24: I'd name this file just `block/account.go`, instead of `block/block_account.go`

PR #259, opened 2018-08-17
`BlockTransaction` should know `Block.Hash`

spikeekips:
### Background

The `BlockTransaction` was designed at the previous ISAAC, so under new ISAAC, based on 'new proposer', it will be natural that `BlockTransaction` has `Block.Hash` data. `BlockTransaction.Block` will have `Block.Hash` and it can be used for

* With single `Transaction` data, we can find which block has this `Transaction`
* With new `GetBlockTransactionsByBlock`, we can get (children) `Transaction`s by `Block.Hash`


Geod24:
Also, instead of storing a `Hash` in it, wouldn't it make much more sense to store a pointer to the block ?
It will reduce storage used and it doesn't matter much since the `Hash` cannot be serialized as part of the transaction.

spikeekips:
> @Geod24 : Why is it named `block_transaction` ? That doesn't make much sense to me, since the transaction makes up the block.

Originally, we don't have `Block` when `block_transaction` is named, so it just represents it will be saved in storage. If you don't like it, please make new name for `BlockTransaction` and `BlockOperation`, etc.

spikeekips:
> @Geod24 : Also, instead of storing a `Hash` in it, wouldn't it make much more sense to store a pointer to the block ?

This `Block` field of `BlockTransaction` will be used for indicating the `Block.Hash` of it when it is serialized and send thru API.

Geod24:
> This Block field of BlockTransaction will be used for indicating the Block.Hash of it when it is serialized and send thru API.

Can't we just send the height ? Much less wasteful.

spikeekips:
@Geod24 : Can't we just send the height ? Much less wasteful.

At current impl, `Block` can be find only by `Hash`.

Geod24:
> At current impl, `Block` can be find only by `Hash`.

Then I think we should implement this first. It's a very basic operation we will need absolutely everywhere.

lib/block_transaction.go
```diff
@@ -48,14 +51,15 @@ type BlockTransaction struct {
 	isSaved     bool
 }
 
-func NewBlockTransactionFromTransaction(tx Transaction, message []byte) BlockTransaction {
+func NewBlockTransactionFromTransaction(block Block, tx Transaction, message []byte) BlockTransaction {
```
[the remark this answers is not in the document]
Charleslee522: @spikeekips Yes, I think copying `Block` is heavy.
If you want to check that the hash is from `Block`, then you need to make sure that the block hash is actually in blockchain(db). Only checking `Block` type is not much helpful.
So I guess, using `string` and check hash in blockchain is best :)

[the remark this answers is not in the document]
spikeekips: @Charleslee522 Okay, I will.

lib/block_transaction_test.go
```diff
@@ -476,3 +517,102 @@ func TestMultipleBlockTransactionGetByAccount(t *testing.T) {
...
+		createdOrder0 = append(createdOrder0, tx.GetHash())
+		txHashes0 = append(txHashes0, tx.GetHash())
+	}
+
+	block0 := testMakeNewBlock(txHashes0)
+	for _, tx := range txs0 {
+		a, _ := tx.Serialize()
+		bt := NewBlockTransactionFromTransaction(block0.Hash, tx, a)
+		err := bt.Save(st)
+		if err != nil {
```
Geod24: `require`

lib/block_transaction_test.go
```diff
@@ -476,3 +517,102 @@ func TestMultipleBlockTransactionGetByAccount(t *testing.T) {
...
+		createdOrder1 = append(createdOrder1, tx.GetHash())
+		txHashes1 = append(txHashes1, tx.GetHash())
+	}
+
+	block1 := testMakeNewBlock(txHashes1)
+	for _, tx := range txs1 {
+		a, _ := tx.Serialize()
+		bt := NewBlockTransactionFromTransaction(block1.Hash, tx, a)
+		err := bt.Save(st)
+		if err != nil {
```
Geod24: `require`

lib/block_transaction_test.go
```diff
@@ -476,3 +517,102 @@ func TestMultipleBlockTransactionGetByAccount(t *testing.T) {
...
+			bo, hasNext := iterFunc()
+			if !hasNext {
+				break
+			}
+
+			saved = append(saved, bo)
+		}
+		closeFunc()
+
+		if len(saved) != len(createdOrder0) {
```
Geod24: `require`

lib/block_transaction_test.go
```diff
@@ -476,3 +517,102 @@ func TestMultipleBlockTransactionGetByAccount(t *testing.T) {
...
+			t.Error("fetched records insufficient")
+		}
+		for i, bt := range saved {
+			if bt.Hash != createdOrder0[i] {
```
Geod24: `require`

lib/block_transaction_test.go
```diff
@@ -476,3 +517,102 @@ func TestMultipleBlockTransactionGetByAccount(t *testing.T) {
...
+
+			saved = append(saved, bo)
+		}
+		closeFunc()
+
+		if len(saved) != len(createdOrder1) {
+			t.Error("fetched records insufficient")
+		}
+		for i, bt := range saved {
+			if bt.Hash != createdOrder1[i] {
```
Geod24: `require` for both

lib/block_transaction_test.go
```diff
@@ -476,3 +517,102 @@ func TestMultipleBlockTransactionGetByAccount(t *testing.T) {
...
+				t.Error("order mismatch")
+				return
+			}
+		}
+
```
Geod24: empty line

lib/block_transaction_test.go
```diff
@@ -476,3 +517,102 @@ func TestMultipleBlockTransactionGetByAccount(t *testing.T) {
...
+		if len(saved) != len(createdOrder0) {
+			t.Error("fetched records insufficient")
+		}
+		for i, bt := range saved {
+			if bt.Hash != createdOrder0[i] {
+				t.Error("order mismatch")
+				return
+			}
+		}
+
```
Geod24: empty

spikeekips:
> @Geod24 : Needs a rebase. Also I'm not sure named return value improve readability.,...

So, your suggestion?

Geod24:
Don't use named return value

spikeekips:
> @Geod24 : Don't use named return value

Why?

Geod24:
Because it makes code less readable

spikeekips:
@Geod24 Hmm, yeah, I can expect that kind of response of the named return value and the other developers have discussed this issue, https://stackoverflow.com/questions/15089726/why-should-return-parameters-be-named . Anyway, I think it is excellent feature of Golang and I have been enjoying to use it. We can open discussion for this issue in our dev team.

PR #260, opened 2018-08-17
Checking `Confirmed` time within a allowed time range

spikeekips:
### Background
The stored `Block` has field, `Confirmed`, this represents the confirmed and verified time by proposer. We can assume the approximate block confirmed time by `Block.Height` and also we can find it by `Block.Confirmed`.

If the proposer's system time is not synced with time server, so `Ballot.ProposerConfirmed()` is so far from synced time, `Block.Confirmed` time could be unreliable.

### Solution

This PR checks the `Ballot.ProposerConfirmed()` with node system time, so if the time diff is over `BallotConfirmedTimeAllowDuration`, the ballot will be not well-formed, `Ballot.IsWellformed()` returns error, `ErrorMessageHasIncorrectTime`


lib/node_runner_checker.go
```diff
@@ -320,8 +321,13 @@ func CheckNodeRunnerHandleINITBallotValidateTransactions(c sebakcommon.Checker,
 		return
 	}
 
-	if checker.Ballot.TransactionsLength() < 1 {
-		checker.VotingHole = VotingYES
+	confirmed, _ := time.Parse(sebakcommon.TIMEFORMAT_ISO8601, checker.Ballot.ProposerConfirmed())
```
Geod24: No error checking ?

lib/node_runner_checker.go
```diff
@@ -320,8 +321,13 @@ func CheckNodeRunnerHandleINITBallotValidateTransactions(c sebakcommon.Checker,
...
+	sub := time.Now().Sub(confirmed)
+	if sub < 0 {
+		sub *= -1
+	}
```
Geod24: I'd much rather have a proper `abs` check than relying on int wrapparound

lib/node_runner_checker.go
```diff
@@ -320,8 +321,13 @@ func CheckNodeRunnerHandleINITBallotValidateTransactions(c sebakcommon.Checker,
...
+	if sub > INITBallotProposerConfirmedTimeAllowDuration {
```
Geod24: Can you put a debug message ?

lib/node_runner_checker.go
spikeekips: The empty `Transaction`s will not be checked, see #237 .

spikeekips:
This PR needs testing code. I will add them ASAP. :)

lib/node_runner_checker.go
Geod24: I'm talking about `time.Parse` return value

spikeekips: In this state, checking INIT ballot does not check the time is well formatted. It is done by `Ballot.IsWellformed()`.

lib/ballot_test.go
```diff
@@ -0,0 +1,96 @@
...
+	node, _ := sebaknode.NewLocalNode(kp, endpoint, "")
+
+	round := Round{Number: 0, BlockHeight: 0, BlockHash: "", TotalTxs: 0}
+
+	{
+		ballot := NewBallot(node, round, []string{})
+		ballot.Sign(kp, networkID)
+
+		if err := ballot.IsWellFormed(networkID); err != nil {
+			t.Error(err)
```
Geod24: Use `require` ;)

lib/ballot_test.go
```diff
@@ -0,0 +1,96 @@
...
+			return
+		}
+	}
+
+	{ // bad `Ballot.B.Confirmed` time; too ahead
```
Geod24: I think those should be their own test instead of anon blocks

lib/common/util.go
```diff
@@ -15,10 +15,23 @@ import (
 
 	"github.com/btcsuite/btcutil/base58"
 	uuid "github.com/satori/go.uuid"
+	"github.com/stellar/go/keypair"
 )
 
+var (
+	TIMEFORMAT_ISO8601 string = "2006-01-02T15:04:05.000000000Z07:00"
```
Geod24: Why is the format an actual date ? That's so weird

spikeekips: @Geod24 date time format in Golang is weird, I agree. :)

lib/ballot_test.go
spikeekips: @Geod24 Separating different test also looks good, but the cases is limited and entire block is also short, this code can be acceptable, I think.

lib/common/time.go
```diff
@@ -0,0 +1,19 @@
+package sebakcommon
+
+import "time"
+
+var (
```
Geod24: `const` ?

lib/common/time_test.go
```diff
@@ -0,0 +1,40 @@
...
+	require.Equal(t, 90758840, parsed.Nanosecond())
+
+	_, offset := parsed.Zone()
+	require.Equal(t, 9*60*60, offset)
+}
+
+func TestParseISO8601Timezone(t *testing.T) {
+	now := time.Now()
+
+	location, err := time.LoadLocation("Europe/Amsterdam") // Wouter Bothoff lives in Amsterdam.
```
Geod24: Almost ;)
While I appreciate easter eggs for our core community members, I am concerned about putting someone's full name here, as targeting of individual has been an issue in the Bitcoin community.

Did he give you express consent? Otherwise I would recommend using his online alias.

spikeekips:
@cmcm2222 @soonkuk @Geod24 Please continue review :)

lib/ballot.go
```diff
@@ -80,6 +81,25 @@ func (b Ballot) IsWellFormed(networkID []byte) (err error) {
...
 
+	var confirmed, proposerConfirmed time.Time
+	if confirmed, err = sebakcommon.ParseISO8601(b.B.Confirmed); err != nil {
+		return
+	}
+	if proposerConfirmed, err = sebakcommon.ParseISO8601(b.ProposerConfirmed()); err != nil {
+		return
+	}
+
+	timeStart := time.Now().Add(time.Duration(-1) * BallotConfirmedTimeAllowDuration)
```
kfangw: I understand you intended duration time as timeEnd - timeStart = BallotConfirmedTimeAllowDuration.

However, the timeStart and TimeEnd variable based on different time.
Hence, `timeEnd - timeStart` is not `BallotConfirmedTimeAllowDuration` but `BallotConfirmedTimeAllowDuration + delta`

You'd better make it based on the same source. like this
```
timeNow := time.Now()
timeStart := timeNow.Add(time.Duration(-1) * BallotConfirmedTimeAllowDuration)
timeEnd := timeNow.Add(BallotConfirmedTimeAllowDuration)
```

PR #262, opened 2018-08-20
Merge sanity check

Geod24:
Merge of master (2d2460d6c9d96b9769dff6852a0a63e151c1426e) into `new-proposer`.


Geod24:
@Charleslee522 @spikeekips if one of you can review that this is sane, I'll merge it and we'll be up to date with master.

Geod24:
I wish go had the ability to have packages instead of a single level. `sebak.block` would be much nicer. Anyway I'll submit an additional PR to master. Would have been nice to note that in https://github.com/bosnet/sebak/pull/258

Charleslee522:
@Geod24 Are we going to merge `new-proposer` into `master` first? Without #241, #243, #244, #248, #256?
Then these will be merged into master after merging `new-proposer`, right?

Geod24:
@Charleslee522 : Last time we discussed, the plan was to work on `new-proposer` until it is ready. But in order to keep working on it without going through too much trouble, we need to merge `master` into `new-proposer` frequently. If we merge `new-proposer` into `master`, then there's no need for a separate branch anymore.

Charleslee522:
@Geod24 Thank you for your explanation. As a result, do you want to merge `new-proposer` now or do it later?

Geod24:
I don't think it's ready. There's still not enough tests to prove it's working. Do you think it is ?

PR #265, opened 2018-08-22
Use testify/require in most test

Geod24:
Makes the code more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and provide a better error message on test failure.


Geod24:
Updated with slightly more code

kfangw:
❤️

PR #269, opened 2018-08-23
Update Gopkg.lock to v0.5.0 format

Geod24:
```
This was previously done in #207 but the changes in #252 reversed to pre-v0.5.0 format.
```

@kfangw you should run `brew upgrade` ;)


kfangw:
@Geod24 ToT Sorry

PR #272, opened 2018-08-24
(WIP) Implement Issue #212 ' [API] Get transactions from proposer'

cmcm2222:
### Background
#212 

### Solution
From the start, This API can be implemented both way that use ` network package`  or `api_transaction.go`. 
I have to be choose which way is better to handle.


Gopkg.lock
```diff
@@ -245,32 +245,31 @@
...
-  digest = "1:c78454799563137757c640be87d6cf532e8c31015ac21eb441c694fffeadf277"
+  digest = "1:a0e12bc26f317c0e2d497baf767285e1790e526e8dd46553c5a67fcbc8692157"
   name = "golang.org/x/sys"
   packages = [
     "cpu",
     "unix",
   ]
   pruneopts = "UT"
-  revision = "78d5f264b493f125018180c204871ecf58a2dce1"
+  revision = "3b58ed4ad3395d483fc92d5d14123ce2c3581fec"
```
Geod24: I don't think that file is necessary either. Do you need to update it ?

lib/api_account.go
```diff
@@ -167,6 +168,7 @@ func GetAccountOperationsHandler(storage *sebakstorage.LevelDBBackend) http.Hand
 					count--
 					if !hasNext || count < 0 {
 						break
+
```
Geod24: Extra whitespace in this file

lib/api_transaction.go
```diff
@@ -137,3 +138,73 @@ func GetTransactionByHashHandler(storage *sebakstorage.LevelDBBackend) http.Hand
...
+
+func GetMissingTransactionsByHashHandler(storage *sebakstorage.LevelDBBackend /**nr *NodeRunner**/) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		vars := mux.Vars(r)
+		key := vars["hash"]
+		// var requestBody []byte
+		var err error
+
+		if r.Method != "POST" {
+			http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
```
Geod24: That's not the correct status though. Under no condition can a `POST` request to this endpoint be correct, so it should be a 400. This indicates "You are missing authentication".

cmcm2222: OK.

Gopkg.lock
cmcm2222: Nope. 
When I `dep ensure` it changed automatically.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "go.gopath": "/Users/youngjunko/sebak/"
+}
```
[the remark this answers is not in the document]
cmcm2222: oops. sorry.

lib/api_account.go
cmcm2222: OK.

anarcher:
I have a question about it. Which part of the handler is different from the `/transactions/{txid}` ?

anarcher:
@cmcm2222  Thanks for the kind explanation! Because a PR ( #279) changes the API handler base structure, After this PR is merged, continue this PR or merge this PR before that :-)

.gitignore
```diff
@@ -13,3 +13,4 @@ db
 *.orig
 .idea
 tests/*/coverage/
+.vscode/
```
kfangw: why still .vscode added?

lib/api_transaction.go
```diff
@@ -137,3 +138,73 @@ func GetTransactionByHashHandler(storage *sebakstorage.LevelDBBackend) http.Hand
 		}
 	}
 }
+
+const GetMissingTransactionByHashHandlerPattern = "/missingtxs"
+
+func GetMissingTransactionsByHashHandler(storage *sebakstorage.LevelDBBackend /**nr *NodeRunner**/) http.HandlerFunc {
```
kfangw: Is the API need the streaming feature?
If pagination is possible then, IMHO, the streaming is not needed
or, client-side can handle it

lib/missing_transaction.go
```diff
@@ -0,0 +1,37 @@
+package sebak
+
+import "encoding/json"
+
+type MissingTransactions struct {
```
kfangw: It is just a map of transactions.
do we need this structure?

lib/missing_transaction_test.go
```diff
@@ -0,0 +1,110 @@
+package sebak
```
kfangw: I don't get the purpose of these test code.
What do you intend by this test code?

lib/network/test_util.go
```diff
@@ -0,0 +1,138 @@
+package sebaknetwork
```
kfangw: IMHO, tests.go is convention for this kind of utils

lib/node_runner.go
```diff
@@ -17,7 +19,7 @@ import (
 	logging "github.com/inconshreveable/log15"
 
 	"boscoin.io/sebak/lib/common"
-	"boscoin.io/sebak/lib/network"
+	// "boscoin.io/sebak/lib/network"
```
kfangw: when you finish your PR then, please remove this kind of unnecessary changes

cmcm2222: @kfangw OK I will. Thanks.

cmcm2222:
@kfangw Design is simple. node that missed some of transactions request to proposer to get those in  node_runner_trasaction_checker.go `func CheckNodeRunnerHandleTransactionsGetMissingTransaction(c sebakcommon.Checker, args ...interface{})`. then Proposer send response to node in api_transaction.go `GetMissingTransactionsByHashHandler `.

kfangw:
@cmcm2222 
As my understanding, the flow is like this
1. Proposer propose ballot which contains transactions
2. A validator receive the ballot
3. If there is at least one transaction in the ballot which the node don't have it
3-1. Request to the proposer to get the transaction

So, the proposer should provide an API which returns the transaction.
The feature is already provided by `transactions` API.

Why you invent a new API?
Are there more features the proposer should provide?

anarcher:
@kfangw  Before A block is confirmed,  transactions of the block are still in transaction pool. So the transactions API need to read from tx pool and then DB backend, If I am right. (On the contrary, Client doesn't need to get transactions in txpool. Is it right? @cmcm2222 )


IMHO, If nodes use the API, I think the API belongs to `Node API` whether sync or async.

cmcm2222:
@anarcher @kfangw
 If I'm not wrong, `func GetTransactionByHashHandler' in api_transaction. go can get only one transaction. so when I implemented this with  @Charleslee522 @kyehyukahn  together. 
and we can assume that proposer's side can be finished consensus earlier than requested node. 
In case of this, proposer's tx pool doesn't have any txs used in that very round. 
so when requested node couldn't find txs in proposer'
s txpool, node should check txs in block. 
that's why this api appear. 

And @anacher mentioned on https://github.com/bosnet/sebak/pull/272#issuecomment-416171847
He recommended implement this API merging #279 after. I think his point sound rational.

.vscode/settings.json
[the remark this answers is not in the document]
Geod24: Ping

.gitignore
Geod24: That's the `gitignore` so it's fine, except it's missing a trailing newline

lib/missing_transaction.go
cmcm2222: @kfangw yep. this for just only `Get Transaction from proposer`. 
if you think that this is unnecessary, I will change it. 
what do you think of this?

lib/network/test_util.go
cmcm2222: ummm .. according to your comment , when I made some functions for test, I need to follow this convention. that is  you meant to ?

.gitignore
kfangw: i mean .vscode/ is here, but the vscode is on the commit.

kfangw:
I am wondering why there is still no logic for the feature yet.
Many of the change are looks like testing code without supporting the feature.

You need to think about two part.
1. request the missing transaction
2. respond the transaction

for the first one
how do you detect the transaction is missing?
where do you store the transaction received?
what trigger the logic?
...

for the second one.
where do you find the requested transaction?
the transaction(s) will be sent one by one? or bulk?
...

Please set up the requirements first. And design according to the requirements.
And share them with a proper document, sample code or in person.

lib/network/test_util.go
kfangw: I think that(test.go) is an implicit convention.

cmcm2222: Aha. I see. 
thanks I will divide to do that for test. 
And.. about `missing_transaction_test.go` , I also agree that if you think existing test code was meaningless. 
so you will tell me it is meaningless I will remove.

cmcm2222:
@kfangw thanks for your advice. 
Now I implement server part ( I mean proposer part ). 
But I can't add in `t.AddAPIhandlers` in `api.go`. 
cause get some transactions in proposer's transaction pool , I need approach node_runner object.

kfangw:
First of all, you'd better rebase your branch on the master. The API refactored to method from function.
Before do that squash your commits into one. `rebase -i HEAD~n` where n denote the number of the commits. It makes easy to rebase your branch on the master.

And the `new-proposer` branch also needed to be rebased on the master or merge it.

Let's talk about getting Tx Pool at API tomorrow.

cmcm2222:
This PR is moved to #306

Geod24:
Please rebase

Charleslee522:
Can this issue continue with this branch?

kfangw:
@cmcm2222 May I take over this issue from you?

cmcm2222:
@kfangw Ah. sorry for late reply. 
why not? you can take it.

kfangw:
@cmcm2222 yeap thanks I will make another PR. Please close this PR.

PR #279, opened 2018-08-27
Clean up network related code

### LINKED ISSUE #155: Support HTTP
Currently, sebak supports only HTTPS.
Because of that, a node must have cert files, and it is hard to handle for operating multiple nodes.
If sebak supports both HTTP and HTTPS, SREs can register certificates on load balancers and operate sebak node with HTTP. If so, SREs can automate updating certificates easily.

spikeekips:
### Github Issue

resolves #155 

### Background

The first idea was simple, using HTTP2. Through the several features was added, especially http handlers, the code is getting bigger and bigger and additional struct also added.

---
Not included in this PR, but I think that the `MemoryNetwork` seems to be useless, so we need to remove it and create some kind of fake http2.

### Solution

* handlers became `http.HandlerFunc` from `sebaknetwork.HandlerFunc`
* `Network` will have only `AddHandler(pattern string, handler http.HandlerFunc) (router *mux.Route)`, so `AddAPIHandlers` was removed
* handlers in `lib/network` was moved to `lib/`, so they don't need context any more to access `LocalNode`
* Handlers became struct methods of `NetworkHandlerNode` and `NetworkHandlerAPI`, these struct will have `sebaknode.LocalNode` and `sebakstorage.LevelDBBackend` as fields
* `Endpoint` will have `Port() string`
* `MessageFromClient  MessageType = "message"` was removed and instead of that, `TransactionMessage MessageType = "transaction"` was added. 
* `MessageBroker` became simple, `ResponseMessage` and `ReceiveMessage` to `Response` and `Receive`
```
type MessageBroker interface {
	Response(io.Writer, []byte) error
	Receive(Message)
}
```
* `MessageBroker.Response` will get `io.Writer` instead of `*HTTP2Network`
* If node's endpoint is set `http`, the `HTTP2Network` will listen without 'TLS' #155 
* The 'TLS' is only available for `https`


cmd/sebak/cmd/run.go
```diff
@@ -41,8 +41,8 @@ var (
 		fmt.Sprintf("%s://%s:%d", defaultNetwork, defaultHost, defaultPort),
 	)
 	flagStorageConfigString string
-	flagTLSCertFile         string = sebakcommon.GetENVValue("SEBAK_TLS_CERT", "sebak.crt")
-	flagTLSKeyFile          string = sebakcommon.GetENVValue("SEBAK_TLS_KEY", "sebak.key")
+	flagTLSCertFile         string = sebakcommon.GetENVValue("SEBAK_TLS_CERT", "")
+	flagTLSKeyFile          string = sebakcommon.GetENVValue("SEBAK_TLS_KEY", "")
```
Geod24: Why ? It's useful to have

cmd/sebak/cmd/run.go
```diff
@@ -164,11 +164,22 @@ func parseFlagsNode() {
...
+		}
+	}
+	if len(flagTLSKeyFile) > 0 {
+		if _, err = os.Stat(flagTLSKeyFile); os.IsNotExist(err) {
+			common.PrintFlagsError(nodeCmd, "--tls-key", err)
+		}
 	}
-	if _, err = os.Stat(flagTLSKeyFile); os.IsNotExist(err) {
-		common.PrintFlagsError(nodeCmd, "--tls-key", err)
+	if strings.ToLower(nodeEndpoint.Scheme) == "https" && (len(flagTLSCertFile) < 1 || len(flagTLSKeyFile) < 1) {
```
Geod24: I think that whole block of code should be wrapped in the `https` check.
If the node does not expose an `https` endpoint, even if someone provides the flags, we ignore it.

Side question: Can an HTTPS node communicate with an HTTP node ?

lib/api.go
```diff
@@ -1,25 +1,27 @@
...
 
-func AddAPIHandlers(s *sebakstorage.LevelDBBackend) func(ctx context.Context, t *sebaknetwork.HTTP2Network) {
-	fn := func(ctx context.Context, t *sebaknetwork.HTTP2Network) {
-		t.AddAPIHandler(GetAccountHandlerPattern, GetAccountHandler(s)).Methods("GET")
-		t.AddAPIHandler(GetAccountTransactionsHandlerPattern, GetAccountTransactionsHandler(s)).Methods("GET")
-		t.AddAPIHandler(GetAccountOperationsHandlerPattern, GetAccountOperationsHandler(s)).Methods("GET")
-		t.AddAPIHandler(GetTransactionByHashHandlerPattern, GetTransactionByHashHandler(s)).Methods("GET")
-	}
-	return fn
+type NetworkHandlerNode struct {
```
Geod24: There's a lot of trailing whitespace here

lib/api.go
```diff
@@ -1,25 +1,27 @@
...
+	localNode *sebaknode.LocalNode
+	network   sebaknetwork.Network
+}
```
Geod24: Those 2 lines also have loads of trailing whitespace

lib/api.go
```diff
@@ -1,25 +1,27 @@
...
+
+type NetworkHandlerAPI struct {
```
Geod24: Make it three lines

Geod24:
That's a lot of code for a single PR...

lib/api_account.go
```diff
@@ -4,204 +4,192 @@ import (
...
-			return
-		} else if !found {
-			http.Error(w, http.StatusText(http.StatusNotFound), http.StatusNotFound)
-			return
-		}
-
-		var err error
-		var ba *block.BlockAccount
+func (api NetworkHandlerAPI) GetAccountHandler(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
```
kfangw: The handler function looks like changed a lot, however, actually, it is not changed a lot.
Could you make an intermediate commit to make it look good?

kfangw: like this
```
func (api NetworkHandlerAPI) GetAccountHandler(w http.ResponseWriter, r *http.Request) {
    {
        vars := mux.Vars(r)
        address := vars["address"]
        ...
    }
}
```

kfangw: or, just confirm that the logic is not changed.

lib/node_runner.go
```diff
@@ -27,13 +26,12 @@ type NodeRunner struct {
 	connectionManager *sebaknetwork.ConnectionManager
 	storage           *sebakstorage.LevelDBBackend
 
-	handleMessageFromClientCheckerFuncs []sebakcommon.CheckerFunc
-	handleBallotCheckerFuncs            []sebakcommon.CheckerFunc
+	handleTransactionCheckerFuncs []sebakcommon.CheckerFunc
```
kfangw: is the renaming related to the issue?

kfangw:
btw, is the integration_tests failed due to the change?

cmd/sebak/cmd/run.go
Geod24: I think those should be two separate PR.
In addition, the logic here seems to be that TLS is used based on the content of those variables.
Since it reads from environment variable, that is problematic.
The choice to use TLS should be based solely on the scheme of the endpoint.
So we can use `http` even if a user provide a value in those variables.

cmd/sebak/cmd/run.go
Geod24: @spikeekips : I was suggesting this:
```go
if strings.ToLower(nodeEndpoint.Scheme) == "https" {
	 if len(flagTLSCertFile) < 1 || len(flagTLSKeyFile) < 1 {
		common.PrintFlagsError(
			nodeCmd,
			"--tls-cert, --tls-key",
			errors.New("HTTPS needs `--tls-cert` and `--tls-key`"),
		)

	if _, err = os.Stat(flagTLSCertFile); os.IsNotExist(err) {
		common.PrintFlagsError(nodeCmd, "--tls-cert", err)
	}

	if _, err = os.Stat(flagTLSKeyFile); os.IsNotExist(err) {
		common.PrintFlagsError(nodeCmd, "--tls-key", err)
	}
}
```

The point is, if we have the flags provided but we don't use https, just ignore them.

cmd/sebak/cmd/run.go
spikeekips: @Geod24 Yes, right. HTTP and HTTPS is determined by endpoint scheme.

lib/api.go
spikeekips: @Geod24 What does it mean?

spikeekips:
> @Geod24 : That's a lot of code for a single PR! Any chance you can break it down a bit ?

I agree. This PR handles several issues. #155 is very long issues and the others can be handled at the same time because these are related with each other. I think the entire changes are limited and not too much. I hope your understanding.

lib/api_account.go
spikeekips: @kfangw Wow great, I did expect or think this kind of way. Thanks, @kfangw .

lib/node_runner.go
spikeekips: @kfangw A little bit, or deeply. The `MessageFromClient` was removed from `sebaknetwork.MessageType`, and 'message from client' was also too, because the message like transaction can be received from node. So 'message from client' expression was replaced by 'transaction'.

lib/api.go
Geod24: I was fooled by Github, ignore this and subsequent whitespace comments 😊

lib/api_account.go
```diff
@@ -4,23 +4,21 @@ import (
...
 )
 
 const GetAccountHandlerPattern = "/account/{address}"
 
-func GetAccountHandler(storage *sebakstorage.LevelDBBackend) http.HandlerFunc {
-
-	return func(w http.ResponseWriter, r *http.Request) {
-
+func (api NetworkHandlerAPI) GetAccountHandler(w http.ResponseWriter, r *http.Request) {
+	{
```
Geod24: Why the extra `{` ?

cmcm2222:
LGTM.

Geod24:
> This PR handles several issues.

That's the root of the problem. Unless the issue is trivial, I don't see any reason to make one giant PR for multiple issues.

> #155 is very long issues and the others can be handled at the same time because these are related with each other.

While they are related, they are not dependent. It's very possible to have a PR with refactorings, one with #155 and other PRs for other issues.
One recurrent problem with PRs addressing multiple issues is that very often, the reasoning becomes circular.
The design of feature A will be affected by the design of feature B, while taking both of them in isolation would have yielded a better solution for both features.

> I think the entire changes are limited and not too much. I hope your understanding.

It's over 1k LoC of diff. If that isn't already a lot, addressing multiple concerns at once makes it practically impossible to correctly review.

kfangw:
@spikeekips Could you merge this If there is no issue?
I want to work api issues based on this change. until then, I will work on the latest commit of this.

Charleslee522:
@spikeekips @kfangw oh, I'm reviewing now :)

lib/api_transaction.go
```diff
@@ -3,19 +3,18 @@ package sebak
...
 	"github.com/gorilla/mux"
 )
 
 const GetTransactionsHandlerPattern = "/transactions"
 
-func GetTransactionsHandler(storage *sebakstorage.LevelDBBackend) http.HandlerFunc {
-
-	return func(w http.ResponseWriter, r *http.Request) {
+func (api NetworkHandlerAPI) GetTransactionsHandler(w http.ResponseWriter, r *http.Request) {
+	{
```
Charleslee522: Is `{` leftover?

lib/api_transaction.go
```diff
@@ -72,10 +71,8 @@ func GetTransactionsHandler(storage *sebakstorage.LevelDBBackend) http.HandlerFu
 
 const GetTransactionByHashHandlerPattern = "/transactions/{txid}"
 
-func GetTransactionByHashHandler(storage *sebakstorage.LevelDBBackend) http.HandlerFunc {
-
-	return func(w http.ResponseWriter, r *http.Request) {
-
+func (api NetworkHandlerAPI) GetTransactionByHashHandler(w http.ResponseWriter, r *http.Request) {
+	{
```
Charleslee522: Is `{` leftover?

lib/network/http2_network.go
```diff
@@ -27,8 +27,20 @@ var (
 	UrlPathPrefixAPI  = fmt.Sprintf("/%s", RouterNameAPI)
 )
 
+type HTTP2MessageBroker struct {
+	network *HTTP2Network
```
Charleslee522: 👍

lib/network/http2_network_config_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"boscoin.io/sebak/lib/common"
+)
+
+func TestHTTP2NetworkConfigHTTPSAndTLS(t *testing.T) {
```
Charleslee522: 👍

lib/api_message.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"boscoin.io/sebak/lib/network"
+)
+
+func (api NetworkHandlerNode) MessageHandler(w http.ResponseWriter, r *http.Request) {
```
Charleslee522: The `MessageHandler`, `BallotHandler`, `NodeInfoHandler` and `ConnectHandler` have same prifix(UrlPathPrefixNode).
So I think defining these four handlers in lib/api_node.go is better for readability.

lib/api_account.go
spikeekips: @Geod24 https://github.com/bosnet/sebak/pull/279#discussion_r212880191

spikeekips:
As @kfangw pointed, the integration test was failed. I'am finding the reason.

cmd/sebak/cmd/run.go
Geod24: So can you revert this ?

docker/node3.env
```diff
@@ -9,3 +9,6 @@ SEBAK_ENDPOINT=https://127.0.0.1:2823
 SEBAK_VALIDATORS=https://127.0.0.1:2821?address=GDIRF4UWPACXPPI4GW7CMTACTCNDIKJEHZK44RITZB4TD3YUM6CCVNGJ&alias=node1 https://127.0.0.1:2822?address=GAYGELM74WJMKSLDN5YP2VAMP64WC4IXIGICUNK2SCVIT7KPTLY7M3MW&alias=node2
 # node1's public key
 SEBAK_GENESIS_BLOCK=GDIRF4UWPACXPPI4GW7CMTACTCNDIKJEHZK44RITZB4TD3YUM6CCVNGJ
+# TLS
+SEBAK_TLS_CERT=sebak.crt
+SEBAK_TLS_KEY=sebak.key
```
Geod24: Those become unnecessary after the revert

lib/api_account.go
Geod24: https://github.com/bosnet/sebak/pull/279/files#r213527182

cmd/sebak/cmd/run.go
spikeekips: @Geod24 In this PR, endpoint determine the HTTP and HTTPS.

Geod24: Exactly. So you don't need to change `flagTLSKeyFile` because it will only be used when endpoint set HTTPS.

spikeekips: @Geod24 I bothered you. :) As you pointed, the code will be reverted. Thanks.

PR #280, opened 2018-08-27
Use stateDB for contract execute/deploy

kfangw:
Contract deploy/execute by using stateDB


kfangw:
@anarcher Please review
You can follow the change easily from `contract/deploy_test`, `contract/executor_test` and `jsvm/executor_test`

lib/contract/api/api.go
```diff
@@ -1,15 +1,69 @@
...
+
+// Get this contract's balance
+func (a *API) GetBalance() (balance *value.Value, err error) {
+	balanceString := a.ctx.StateDB().GetBalance(a.contractAddress)
+	amount := sebakcommon.MustAmountFromString(balanceString)
+	balance, err = value.ToValue(uint64(amount))
+	return balance, nil
+}
+
+// Call another contract
```
anarcher: Why do you comment on this method? :-)

lib/contract/init/init.go
```diff
@@ -0,0 +1,7 @@
+package init
+
+import "boscoin.io/sebak/lib/contract/native/execfunc"
+
+func init() {
+	execfunc.Init()
+}
```
anarcher: Is not good `exexfunc.init` ?

lib/contract/api/api.go
kfangw: I am not sure that we need call another contract from a contract.
now, it may cause circular import. 
I will handle it when we need that feature.

lib/contract/init/init.go
kfangw: good point i will change. 
It was influenced by the other project.

PR #282, opened 2018-08-28
Drop EOL Go 1.9; Add Go 1.11 and use it as default

Geod24:
It was released a few days ago.
The docker images where dropped.
We depend on the CircleCI image which might be supported a bit longer, but let's update sooner than later.


Charleslee522:
Please specify reviewers.

Issue #284, opened 2018-08-28
Update to a new version of go-observable


Geod24:
When https://github.com/GianlucaGuarini/go-observable/pull/5 is merged, we need to upgrade to that version to get rid of one of our race condition.

Geod24:
FYI @kfangw

kfangw:
@Geod24 thx

PR #285, opened 2018-08-28
api: Improve test code, small refactor for DRYness

Geod24:
While I was solving the race issue I came across this.
The bug is actually in `go-observable`, not ours (or rather, we use a pattern that didn't seem to be supported by `go-observable`) so this just refactors stuff.


Geod24:
See https://github.com/bosnet/sebak/issues/284 for more details

Issue #274, opened 2018-08-24
Write API document


outsideris:
For client development, public API document is needed as Swagger or Postman format.

It is best if we make API document automatically from source.

anarcher:
- https://github.com/growler/go-slate

anarcher:
We can close it? @kfangw @outsideris  Maybe we can fill more information of it. But I would like to care with another new issue and PR.. :->

anarcher:
- Add github page :  https://bosnet.github.io/sebak/api/index.html
- Thanks! @outsideris !!

anarcher:
I will close it now, If we have another issues or needs, we will make other issues for it. :->

PR #314, opened 2018-09-03
Using API blueprint for API document

outsideris:
### Github Issue
Related to #274

### Background
It's another approach for API document which is different with #300 . 
[API blueprint](https://apiblueprint.org/) is API document format.

### Pros
* API blueprint is compatible with markdown. So, it's human-readable in GitHub before generating HTML.
* It doesn't disrupt our source code. API document needs much text. I dislike too many comments above the source code.

### Cons
* We should care about API document. Because it is not generated from our source. We should write them. Otherwise, it will be outdated quickly.
* Default theme is not so good even though we can customize it.
* There is no "Try it out" feature.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1,0 +1,8 @@
+FORMAT: 1A
+HOST: https://localhost:12345/api
+
+# Sebak Client API
+SEBAK, the next BOScoin Tokennet with ISAAC consensus protocol 
+
+<!-- partial(v1/accounts.md) -->
+<!-- partial(v1/models.md) -->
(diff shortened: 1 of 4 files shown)


outsideris:
Additionally, we can prevent our document is outdated if we use [dredd](https://dredd.readthedocs.io/en/latest/).

Dredd is API test framework. It will run test based on API blueprint document: 



Dredd uses our parameter example and response. So, if we change our API but, document doesn't changed, test will be fail.

However, we should write our document strictly to use dredd. Because all example on the document should be real test data.

anarcher:
I like this approach of api blueprint.  And `github.com/swaggo/swag` doesn't support multiple descriptions not yet..

anarcher:
Can we use the `dredd` with CI?

cmcm2222:
Sorry for late review. 
Thanks for kind explanation. @outsideris 
LGTM

Issue #118, opened 2018-06-21
API: Response format application/hal+json


kfangw:
http://stateless.co/hal_specification.html

Geod24:
That makes sense if we provide a RESTful API. Is that the intent ?

kfangw:
yeap. I intend to build a RESTful APIs.
But, after build a response as it is, I will change the response later.

outsideris:
What is better than just plain JSON.
It seems similar to HATEOAS.

spikeekips:
RFC is https://tools.ietf.org/html/rfc4627 .

spikeekips:
Good job @kfangw

PR #319, opened 2018-09-03
API response (HAL+JSON)

kfangw:
### Github Issue
#118 

### Background
To enhance user(client-side) experience, HAL+JSON response is needed
Spec is here https://tools.ietf.org/html/rfc4627

### Solution
To make it easy, a library is adopted https://github.com/nvellon/hal
By using the library, the only thing needed is to implement `Mapper` interface
```
type Mapper interface {
	GetMap() Entry
}
```

### Possible Drawbacks
* prev/next for pagination is not yet implemented.
* The API response is still returned as-is form. I need to consider the clients.


kfangw:
@spikeekips @anarcher @outsideris Please review this change.

@spikeekips Please double-check the keys of a response map such as `account_id`, `checkpoint`, etc..

anarcher:
What about their responses has their own package? e.g. `/api/resources` ?

anarcher:
The Resoures can be a implemantion of `common.Serializable`?  For Render, common interface for rendering can be useful. IMHO.  :-)

lib/api_resource.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+type APIResourceOperation struct {
+	hash    string
+	txHash  string
+	funder  string //Source Account
+	account string //Target Account
+	otype   string
+	amount  string
+}
```
anarcher: IMHO, How about `resource.Operation` ? 
And  the `resource.Operation` has `block.Operation` like below:

```
type  Operation sturct {
      op *block.Operation
...
}
func (o Operation) GetMap() hal.Entry {
    return hal.Entry{
       "id" : o.op.hash,
      ...
   }
}
```
It makes senes? :-)

kfangw: I thought about that. But, I decided that the Resources does not know the blocks.

kfangw:
I will update commit for your recommendation. Thx ~! @anarcher

spikeekips:
@kfangw Like`funder` And `account` we’d better to use the existing terms, I think. 

- funder -> source
- account -> target

lib/api_resource.go
anarcher: But Copy values can cost. humm. And binding blocks and resources can make handlers itself to complex or verbose IMHO. 

``` 
resource.NewAccount(block.BlockAccount) *resource.Account 
api.NewAPIResourceAccunt(block.BlockAccount) *api.ResourceAccount
```

kfangw: It may possible when the restructuring is done.
Let me change~

Issue #19, opened 2018-06-01
Support 'HTTP-Problem' in API response


spikeekips:
The detailed specification of 'HTTP-Problem' can be found in https://tools.ietf.org/rfc/rfc7807.txt .

Geod24:
Given that:
- the APIs are still being developed
- this has little impact on the core of our product
- the ROI is pretty low

I moved this to the `MainNet` milestone

spikeekips:
Good job @kfangw

PR #320, opened 2018-09-04
http problem

kfangw:
### Github Issue
#19 

### Background
Support Problem Details for HTTP APIs (https://tools.ietf.org/rfc/rfc7807.txt)

### Solution
Fill the fields of the Problem struct

```
type Problem struct {
	Type string `json:"type"`
	Title string `json:"title"`
	Status int `json:"status,omitempty"`
	Detail string `json:"detail,omitempty"`
	Instance string `json:"instance,omitempty"`
}
```

### Possible Drawbacks
need to prepare documents for the `Instance` field

### DIFF
--- a/lib/error/http_problem.go
+++ b/lib/error/http_problem.go
@@ -1,12 +1,14 @@
-package sebakerror
+package errors
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 )
 
 const (
-	HttpProblemDefaultType = "about:blank" // It should be URI
+	HttpProblemDefaultType     = "about:blank" // It should be URI
+	HttpProblemErrorTypePrefix = "https://boscoin.io/sebak/error/"
 )
 
 type problem struct {
@@ -50,6 +52,10 @@ func NewDetailedStatusProblem(status int, detail string) problem {
 	p := NewStatusProblem(status)
 	p.Detail = detail
 	return p
+}
+
+func NewErrorProblem(err *Error) problem {
+	return problem{Type: fmt.Sprintf("%s%d", HttpProblemErrorTypePrefix, err.Code), Title: err.Message}
 }
 
 func Problem(w http.ResponseWriter, detail string, status int) {
--- a/lib/error/http_problem.go
+++ b/lib/error/http_problem.go
@@ -52,6 +52,10 @@ func NewDetailedStatusProblem(status int, detail string) problem {
 	return p
 }
 
+func Problem(w http.ResponseWriter, detail string, status int) {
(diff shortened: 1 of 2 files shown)


kfangw:
@spikeekips @anarcher please review 

And I have a question.
The location of the problem(s) is now `sebakerror` is it suitable?

spikeekips:
@kfangw: It will be good if you can not find better place:)

lib/error/http_problem.go
```diff
@@ -0,0 +1,139 @@
...
+	p.Instance = instance
+	return p
+}
+
+func (p Problem) SetDetail(detail string) Problem {
+	p.Detail = detail
+	return p
+}
+
+func (p Problem) Serialize() ([]byte, error) {
```
spikeekips: It just suggestion, please feel free. ;) `http.Error` receives writer and error code. How about design `StatusProblem` to be similar 
https://golang.org/pkg/net/http/#Error . It will be zero learning :)

kfangw:
Type and Title for Problem are mandatory
Status is optional but, it should has the same meaning as http status
Detail is optional too. And it can be filled in anywhere case by case
Instance is optional too. 

Hence there is only 3 New function
```func NewProblem(problemType string, title string) problem```
```func NewStatusProblem(status int) problem ```
```func NewDetailedStatusProblem(status int, detail string) problem ```
And the problem package is not exported

**Comment by anarcher** [comment]
If I have a `sebakerror` and I want to return the problem of it. What do I do for it? :-) 
And within event stream, How to handle error with the problem? :-) 
If we need a glue between error and problem, What do you think like below:

```
httputils.WriteJSONError(w,err) // uses http.Error for writing the error 
sebakerror.HTTPError // is an error with extra information (http.StatusCode and problem)
```
(If we return sebakerror, default values fill the problem fields makes sense? )
But EventStream has to have own writeJSONError :-) 

```
   HTTP/1.1 403 Forbidden
   Content-Type: application/problem+json
   Content-Language: en

   {
    "type": "https://example.com/probs/out-of-credit",
    "title": "You do not have enough credit.",
    "detail": "Your current balance is 30, but that costs 50.",
    "instance": "/account/12345/msgs/abc",
    "balance": 30,
    "accounts": ["/account/12345",
                 "/account/67890"]
   }
```

In problem+json,  type and instance have to have a link?

spikeekips:
@anarcher In the most of the cases of event stream+hal, the items of _embeded will be streamed, so if error occurred during streaming, it is natural to put the problem+json and no more items, I think.

kfangw:
I wrote one more constructor ```func NewErrorProblem(err *Error) problem```
Is it works as the glue between error and problem?
@anarcher

PR #324, opened 2018-09-05
Merge new-proposer into master

Geod24:
Merge `new-proposer` into `master`

PR #325, opened 2018-09-05
received data may not ordered. do not check the sequence

kfangw:
### Github Issue
Bug
### Background
Test failed randomly

### Solution
The received data sequence is not guaranteed due to go-observable on function is concurrent
Hence, remove the sequence check

### DIFF
--- a/lib/api_test.go
+++ b/lib/api_test.go
@@ -218,10 +218,8 @@ func TestGetAccountTransactionsHandler(t *testing.T) {
 
 	require.Equal(t, len(btmap), len(receivedBts), "length is not same")
 
-	i := 0
 	for _, bt := range receivedBts {
 		require.Equal(t, bt.Hash, btmap[bt.Hash].Hash, "hash is not same")
-		i++
 	}
 }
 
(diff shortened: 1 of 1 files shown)


kfangw:
@spikeekips This is what you told a few days ago.
please confirm ^^

Geod24:
This bug was actually the reason I did https://github.com/bosnet/sebak/pull/275
AFAIK there's no more race in this part of the code?

lib/api_test.go
```diff
@@ -214,11 +216,11 @@ func TestGetAccountTransactionsHandler(t *testing.T) {
 	var receivedBts []BlockTransaction
 	json.Unmarshal(readByte, &receivedBts)
 
-	require.Equal(t, len(bts), len(receivedBts), "length is not same")
+	require.Equal(t, len(btmap), len(receivedBts), "length is not same")
 
 	i := 0
```
spikeekips: `i` seems to be useless.

kfangw: thx i missed it

Geod24:
> The received data sequence is not guaranteed due to go-observable on function is concurrent

`go-observable` does not start any goroutine. 

> The reason why this bug occurred is not a race condition.
> Just wrong written testcase.

Can you link to a test that failed since the fixes to `go-observable` / the previously mentioned PR ?
I haven't seen any, so I don't see why that change is useful.
AFAICS it's just avoid testing the order, when the order is already guaranteed (and should be tested).

Geod24:
Also, since this was hastily merged, it conflicts with https://github.com/bosnet/sebak/pull/324 and I have to redo it...

kfangw:
@Geod24 The order is not guaranteed. And it is not needed to be guaranteed.
It is just a bug fix for wrong written testcase.

PR #327, opened 2018-09-06
Api test fix

kfangw:
https://github.com/bosnet/sebak/pull/325


kfangw:
It was confirmed at #325

Issue #286, opened 2018-08-28
Refactoring API structure


kfangw:
Several weeks ago, @anarcher @zzim2x mentioned restructuring API in person.

Description will be filled soon.

kfangw:
Could you share your code snippet here? @zzim2x

anarcher:
Ping @zzim2x

zzim2x:
I've not found the code. It might seem to be on my other laptop.

spikeekips:
Can we close this issue?

anarcher:
Yes, I will close it.   :-)

PR #336, opened 2018-09-06
Using eventstream with APIHandlers

### LINKED ISSUE #299: API: Refactoring API Eventstream
- With API Handler struct, Design `Streamer` or `EventStream` struct to work with `http.Flusher`
- related to: #286

### LINKED ISSUE #310: Fix client API content-type
I didn't check all API endpoints, but when I checked `/api/account/{address}`, its `content-type` is `text/plain` not `application/json`.

```
HTTP/2 200
content-type: text/plain; charset=utf-8
content-length: 259
date: Fri, 31 Aug 2018 08:59:30 GMT

{"Address":"GDVSXU343JMRBXGW3F5WLRMH6L6HFZ6IYMVMFSDUDJPNTXUGNOXC2R5Y","Balance":"1000000000000","Checkpoint":"5q1dAkXXSBNnmYZU7XWQmGGev-5q1dAkXXSBNnmYZU7XWQmGGev","CodeHash":null,"RootHash":"0x0000000000000000000000000000000000000000000000000000000000000000"}
```

I think it should be `application/json` as same as the response body.

anarcher:
### Github Issue

Related  #286  
Closes  #299 
Fix #310

### Background

Using EventStream in APIHandlers
For writing a response, add `httputils` pkg. For now, this pkg has a little funcs for HTTP handling
```
   +IsEventStream(r *http.Request) : bool
   +WriteJSON(w http.ResponseWriter, code int, v interface{}) : error
```

### Solution

```
    if httputils.IsEventStream(r) {
        event := fmt.Sprintf("address-%s", address)
        es := NewDefaultEventStream(w, r)
        es.Render(blk)
        es.Run(observer.BlockAccountObserver, event)
        return
    }

    if err := httputils.WriteJSON(w, 200, blk); err != nil {
        http.Error(w, "Error reading request body", http.StatusInternalServerError)
    }
```

### Possible Drawbacks

- We will consider with #118, #19 and this PR

### DIFF
--- a/lib/httputils/json.go
+++ b/lib/httputils/json.go
@@ -1,0 +1,42 @@
+package httputils
+
+import (
+	"encoding/json"
+	"net/http"
+
+	common "boscoin.io/sebak/lib/common"
+)
+
+func WriteJSON(w http.ResponseWriter, code int, v interface{}) error {
+	w.Header().Set("content-type", "application/json")
+	w.WriteHeader(code)
+
+	if s, ok := v.(common.Serializable); ok {
+		return writeSerializable(w, s)
+	}
+
+	bs, err := json.Marshal(v)
+	if err != nil {
+		return err
+	}
+
+	if _, err := w.Write(bs); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func writeSerializable(w http.ResponseWriter, s common.Serializable) error {
+
+	bs, err := s.Serialize()
+	if err != nil {
+		return err
+	}
+
+	if _, err := w.Write(bs); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/lib/httputils/json.go
+++ b/lib/httputils/json.go
@@ -3,33 +3,14 @@ package httputils
 import (
 	"encoding/json"
 	"net/http"
-
-	common "boscoin.io/sebak/lib/common"
 )
 
+// WriteJSON writes the value v to the http response as json encoding
 func WriteJSON(w http.ResponseWriter, code int, v interface{}) error {
 	w.Header().Set("content-type", "application/json")
 	w.WriteHeader(code)
 
-	if s, ok := v.(common.Serializable); ok {
-		return writeSerializable(w, s)
-	}
-
 	bs, err := json.Marshal(v)
-	if err != nil {
-		return err
-	}
-
-	if _, err := w.Write(bs); err != nil {
-		return err
-	}
-
-	return nil
-}
-
-func writeSerializable(w http.ResponseWriter, s common.Serializable) error {
-
-	bs, err := s.Serialize()
 	if err != nil {
 		return err
 	}
--- a/lib/api.go
+++ b/lib/api.go
@@ -13,6 +13,13 @@ import (
 
 const maxNumberOfExistingData = 10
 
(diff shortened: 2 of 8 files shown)


Charleslee522:
@anarcher Can you rebase? kk

lib/httputils/json.go
```diff
@@ -0,0 +1,42 @@
+package httputils
+
+import (
+	"encoding/json"
+	"net/http"
+
+	common "boscoin.io/sebak/lib/common"
+)
+
+func WriteJSON(w http.ResponseWriter, code int, v interface{}) error {
```
Geod24: Please add documentation. Also would a `MustWriteJSON` be useful ?

lib/httputils/json.go
```diff
@@ -0,0 +1,42 @@
...
+	w.Header().Set("content-type", "application/json")
+	w.WriteHeader(code)
+
+	if s, ok := v.(common.Serializable); ok {
```
Geod24: Doesn't `json.Marshal` do that itself ?

lib/httputils/json.go
anarcher: If This func doesn't return an error, I will use the name `MustWriteJSON` (maybe).

Geod24: I mean, add another function, `MustWriteJSON`, which wraps this with a `panic`.

lib/httputils/json.go
anarcher: It's a good question. :-)  Sometimes, `common.Serializable` is confusing for me. I need to think about it. And I have a question. The `common.Serializable` exists Only for DB Storage layer?

As you mentions, Just doing `json.Marhsal` make sense in API layer. Thanks :-)

lib/httputils/json.go
anarcher: And If someone needs to `MustWriteJSON` and make it, I am OK to make it :-)

Geod24: Ah it was just for quick & dirty tests. But I guess we can wait until the need arises.

lib/httputils/json.go
Geod24: Oh wait, we have our own `Serializable` interface ? Why aren't we using `pkg/encoding`'s [`TextMarshaller`](https://golang.org/pkg/encoding/#TextMarshaler) which has the exact same definition and is recognized by the std lib ?

anarcher: First my thought,  `WriteJSON`  takes  `json.Marshaler` and handles it. But  From the result, I will change the `RenderSerializeFunc` to `RenderJSONFunc` too. ㅠㅠ  :-> 
I don't think the `TextMarshaller` is suitable for api rendering layer. IMHO.

anarcher:
First my thought,  `WriteJSON`  takes  `json.Marshaler` and handles it. But  from the result, I will change the `RenderSerializeFunc` to `RenderJSONFunc` of `EventStream` too. ㅠㅠ  :-> 
I don't think the `TextMarshaller` is suitable for api rendering layer. IMHO.
And   We maybe need another struct (Resource) for supporing `json+hal`. I feel like to `WriteJSON(v interface{})` is not bad.

Issue #122, opened 2018-06-21
API: Paging support


kfangw:
If there is a lots of data to be sent to client, it may be a burden to network.

We need an idea to support paging for regular API call and streaming call both.

Geod24:
Something like https://developer.github.com/v3/guides/traversing-with-pagination/ ?

kfangw:
yes, but i am not sure for streaming.
I will attach details when I am clear for this issue.

kfangw:
#118 the resource is expressed in the JSON+HAL

kfangw:
cursor, ordering is needed

anarcher:
How to represent the `cursor` in HTTP Request/Response? 
- GET querystring  such like`?cursor={cursor}&limit={limit}` 
-  cursor =  `{key_prefix}-{index_item}-{block_height}-{sequence_id}-{uuid}` ?  We can need some kind of `paging_token`?

anarcher:
What do you think about it? 

```
type Paging struct {
     Cursor  string `json:cursor`
     Limit uint64 `json:limit`
     Order string `json:order`
  
     err error // maybe?
}

//r := *http.Request 
p := NewPaging(r) // validate inputs
if p.Error() != nil { // or p.Validate() error 
     // return error
}

bks := GetBlockByXXX(p.Cursor,p.Limit,p.Order)
```

anarcher:
I think this issue is closed :-) I will close it!!!

PR #140, opened 2018-06-28
Replace checkpoints with sequence ID

Geod24:
```
Next checkpoint do not play a role in the consensus. They take up space and are solely debugging information.
Their original role was to provide a stronger rule w.r.t balance verification,
however such rules proved to be impossible to implement without compromising
other aspects of the design (namely, it made issuing transactions very complicated).
```

PR #337, opened 2018-09-06
Page support

kfangw:
### Github Issue
#122 

### Background
For the client API pagination needed to be supported
To implement this feature, cursor and limit required

### Solution

* iterator 

IteratorOptions struct is newly introduced
```
type IteratorOptions struct {
	Reverse bool
	Cursor  []byte
	Limit   uint64
}
```
The GetIterator function get the IteratorOptions as argument
```func (st *LevelDBBackend) GetIterator(prefix string, option *IteratorOptions) (func() (IterItem, bool), func())```

The wrapping function for each model is changed like this

first returned function (next function) return one more value which denote the key
the key can be used for cursor
```
func LoadBlockTransactionsInsideIterator(
	st *sebakstorage.LevelDBBackend,
	iterFunc func() (sebakstorage.IterItem, bool),
	closeFunc func(),
) (
	func() (BlockTransaction, bool, []byte),
	func(),
) 
```

* reverse index key management

The reverse index for transactions, blocks, and operations is like this
{key_prefix}-{index_item}-{block_height}-{sequence_id}-{uuid}

Hence, the ordering follows this priority

block height
sequence number
created time
All of the nodes have the same order of block height and sequence number
And created time may diverse

Consequently, the cursor can be used for any nodes when the cursor consists block height and sequence number.

### Possible Drawbacks
reverse index key may too long

### DIFF
--- a/lib/block/account.go
+++ b/lib/block/account.go
@@ -106,30 +106,30 @@ func GetBlockAccount(st *storage.LevelDBBackend, address string) (b *BlockAccoun
 	return
 }
 
-func GetBlockAccountAddressesByCreated(st *storage.LevelDBBackend, reverse bool) (func() (string, bool), func()) {
-	iterFunc, closeFunc := st.GetIterator(BlockAccountPrefixCreated, reverse)
+func GetBlockAccountAddressesByCreated(st *storage.LevelDBBackend, iteratorOptions *storage.IteratorOptions) (func() (string, bool, []byte), func()) {
+	iterFunc, closeFunc := st.GetIterator(BlockAccountPrefixCreated, iteratorOptions)
 
-	return (func() (string, bool) {
+	return (func() (string, bool, []byte) {
 			item, hasNext := iterFunc()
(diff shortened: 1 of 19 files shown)


kfangw:
based on #140, something needed to be changed. please wait for a while.

kfangw:
The reverse index for transactions, blocks, and operations is like this
```{key_prefix}-{index_item}-{block_height}-{sequence_id}-{uuid}```

Hence, the ordering follows this priority
1. block height
2. sequence number
3. created time

All of the nodes have the same order of block height and sequence number
And created time may diverse

Consequently, the ```cursor``` can be used for any nodes when the cursor consists block height and sequence number.

@anarcher @spikeekips please review ~^^

kfangw:
The prefix can be shortened by #347

spikeekips:
If possible, maximum number of `IteratorOption.Limit` should be checked for each `GetBlockXXX`.

Issue #100, opened 2018-06-19
Discuss: Handling `HTTP2LogOutput` 


anarcher:
Currently,  `HTTP2LogOuput(io.Writer)` make all access log of  HTTP handlers, and the log format is different than SEBAK Log format. And Especially `/connect` log make verbose.   

```
127.0.0.1 - - [19/Jun/2018:11:10:43 +0900] "POST /connect HTTP/2.0" 200 127 "" "v-GBMD.LBZL"
127.0.0.1 - - [19/Jun/2018:11:10:43 +0900] "POST /connect HTTP/2.0" 200 127 "" "v-GCZU.7SRT"
127.0.0.1 - - [19/Jun/2018:11:10:44 +0900] "POST /connect HTTP/2.0" 200 127 "" "v-GBMD.LBZL"
127.0.0.1 - - [19/Jun/2018:11:10:44 +0900] "POST /connect HTTP/2.0" 200 127 "" "v-GCZU.7SRT"
127.0.0.1 - - [19/Jun/2018:11:10:45 +0900] "POST /connect HTTP/2.0" 200 127 "" "v-GBMD.LBZL"
```
Verbose logs can be decreased by #99 :-) 

Discussion: 
- Replace `goriila.CombinedLoggingHandler`  to Log15-compatibility CustomLoggingHandler 
- Ignore logging `/connect` access log (by option or loglevel?) 
    - We can change to log only when first failure of connection. (copy from #99)

PR #345, opened 2018-09-09
Refactoring log part and Support JSON log output

spikeekips:
### Github Issue

related to #100 

### Background

Logging part easily can be crowed and unmanaged, and for our incoming E2E testing, we need to clean up the logging part.

### Solution
Clean up and refactoring logging

#### Log format can be set with `--log-format=json`

By default, `terminal` will be used, this is the existing log format. With
`--log-format=json`, we can get the jsonized log message by one line like this,
```
{"lvl":"dbug","module":"sebak","msg":"NodeRunner started","node":"GCPQ.7DUR","t":"2018-09-09T17:31:04.745872455+09:00"}
...
{"caller":"run.go:385","error":"http: Server closed","lvl":"crit","module":"main","msg":"failed to start node","t":"2018-09-09T19:23:16.483210507+09:00"}
```

#### `--log-output` -> `--log`

To set the log output file, the option name became shorter, `--log`. The environmen variable name also is changed from `SEBAK_LOG_OUTPUT` to `SEBAK_LOG`

#### HTTP log can be set with `--http-log` and `--http-error-log`

In the previous impl, we can set a http log thru endpoint like this,

```
$ sebak node run ... --endpoint "https://localhost:12345?HTTPLog=/tmp/http.log"
```
and http error log can not be set, it was marked as 'TODO'.

Now with `--http-log` and `--http-error-log` we can set them. By default, these will be streamed thru `os.Stdout`.

#### `network.Message` is renamed to `common.NetworkMessage`

To filter `network.Message` by `common.formatJSONValue()`, it is moved to new place, `MessageType` also too.

#### 'json' log output

Using json log output, we can simply parse the log output for various usages, especially incoming E2E testing env.  For doing so, the existing logging part was changed to make the output to be more json.

##### `common.JsonFormatEx` instead of using `logging.JsonFormatEx`

For support json log output, `logging.JsonFormatEx` is simple choice, but to filter some additional data type, `logging.JsonFormatEx` use internal functions.

* `logging.JsonFormatEx`
```
{
    "ballot": "{\n  \"H\": {\n    \"hash\": \"3P6h9nwLfxNVqHJRCKC6fM71SLXmRvwwqz8Uo6Lbmoq5\",\n    \"signature\": \"2DKoYhyaveJMJRXJ6FkBBakhCfU1MHStTLGNdEQu4JPaok2WpkfoN8b2HgTXdQhNSDG41y9JVAPZeyh88ocSU5kN\",\n    \"proposer-signature\": \"61r95e3R6Gyczi3bFdC7CVfwxxH9tS7SKRDppnn8QBLQxbmBAEKB1kutrKrC52YPjXtqtehnmFDx1BGmqcE2qcWL\"\n  },\n  \"B\": {\n    \"confirmed\": \"2018-09-09T19:39:26.971192805+09:00\",\n    \"proposed\": {\n      \"confirmed\": \"2018-09-09T19:39:26.943698948+09:00\",\n      \"proposer\": \"GBIDQUJSXBBNZERPP4YX3CCEOELLTW6GHWUO6HN3WTYJB5KBJJEUOL3M\",\n      \"round\": {\n        \"number\": 0,\n        \"block-height\": 1,\n        \"block-hash\": \"4t5RMTjYRs1N3dAQ3965vjsxyRPdnU5C8iGk2a9SKJnk\",\n        \"total-txs\": 0\n      },\n      \"transactions\": null\n    },\n    \"source\": \"GBIDQUJSXBBNZERPP4YX3CCEOELLTW6GHWUO6HN3WTYJB5KBJJEUOL3M\",\n    \"state\": 1,\n    \"vote\": \"YES\",\n    \"reason\": null\n  }\n}",
    "lvl": "dbug",
    "module": "sebak",
    "msg": "new ballot created",
    "node": "GBID.JJEU",
    "t": "2018-09-09T19:39:27.016154213+09:00"
}
```

* `common.JsonFormatEx`
```
{
    "ballot": {
        "H": {
            "hash": "Bpro7ddcsfxTbvB9464N4jg3qNMEqe6nnebbHDsSwuUN",
            "signature": "3VzNMdMcf2tWdfvQBJ1tAGaJRcG3rhJEDSbu7C7P2xHbAvmgwMHYxsz6KT5K5TmzkqmieZUyQNxKxTJQyitpdxaa",
            "proposer-signature": "3q7MY8SQJt4ZFH3Zq4Eig4xEUeBe3bH7tx76KQdyuLftVRbjQHCegfBNZzsyvmzx2gHCAuLg7vM9JRKwE5wKHUsE"
        },
        "B": {
            "confirmed": "2018-09-09T19:37:35.604998124+09:00",
            "proposed": {
                "confirmed": "2018-09-09T19:37:35.575004956+09:00",
                "proposer": "GAZ6SRZGJYZDRHLV7RXZWICWOAK4GFAROU66HPZVLD6NFX53RQQU3HML",
                "round": {
                    "number": 0,
                    "block-height": 1,
                    "block-hash": "6MPJUtT1WEsUN1agK637id3weUKsitFvgoCETkzN3e7W",
                    "total-txs": 0
                },
                "transactions": null
            },
            "source": "GAZ6SRZGJYZDRHLV7RXZWICWOAK4GFAROU66HPZVLD6NFX53RQQU3HML",
            "state": 1,
            "vote": "YES",
            "reason": null
        }
    },
    "lvl": "dbug",
    "module": "sebak",
    "msg": "new ballot created",
    "node": "GAZ6.RQQU",
    "t": "2018-09-09T19:37:35.636310053+09:00"
}
```

> These json log message is prettified one. In logs, it will be printed in one line.

With customized `common.JsonFormatEx`, the json output will be merged with json object instead of string.

### For making json log output to be more jsonized

* For `struct`, if you want to make it merged into json log, implement `MarshalJSON()` or `Serialize()`. Without them, it will just be printed using `json.Marshal()`. If it has `String()` method, it will be used.
* We have used the `Serialize()` to make it jsonized. The `json` is default output, so I suggest we will `json.Marshal()` directly instead of implementing `Serialize()`. We can add json tags in struc fields and if more sophisticated control of fields and it's values we should implement `MarshalJSON()`.


Gopkg.lock
```diff
@@ -211,50 +195,6 @@
   revision = "792786c7400a136282c1664665ae0a8db921c6c2"
   version = "v1.0.0"
 
-[[projects]]
-  digest = "1:a22debd993d18b9203c985810f33f52e7583724800ee8d15e7d362e7175639d1"
-  name = "github.com/prometheus/client_golang"
```
Geod24: Looks like this is removing the prometheus dependency ?

spikeekips: @Geod24 Yes, it is not used at this time. Did I miss something?

spikeekips:
> @Charleslee522: I figure, there are two issues in this pr, as you wrote in summary(JSON log and moving Message to common package).

If these are not related, I will. `common.NetworkMessage` and json logging is related. To filter the `network.Message` in `common.formatJSONValue()`, there are import circulation issue.

Gopkg.lock
Geod24: We did. https://github.com/bosnet/sebak/commit/f776bf0066f3e8191e18cf14c28321d60538f45c#diff-bc5b5719c421f802a296dffd1525b2f4L180
This line shouldn't have been removed.

spikeekips: @Geod24 Hmm, it was removed. I see. We can recover it later, right?

Geod24: I think we should recover it now. Doesn't make sense to remove this dependency to re-add it in a few commits.

Geod24: I'll raise a PR.

spikeekips: @Geod24 This PR should not keep the useless one and the unused packages in Gopkg also should not be in here.Anyway it's not matter to keep the old one.

PR #347, opened 2018-09-10
refactor prefix to make it short

kfangw:
### Background
Make the key short


kfangw:
When we have ```block``` package for block, block_transaction, and block_operation
then, the prefix can be moved to block package.

but now, common may be ok.

@spikeekips because of #337, I am worried about the long index key.

kfangw:
rebased

PR #349, opened 2018-09-10
Add LevelDB.Walk to support cursor

anarcher:
### Github Issue

Related to:  #122 

### Background

Instead of Iterator, Use Walk such like https://golang.org/pkg/path/filepath/#Walk
I am not sure this approach is suitable for leveldbBackend so it is a pr for a question which is suitable or not. 

```
    var walkedKeys []string
    // prefix,cursor,reverse,walkFunc
    err := st.Walk("test-", "test-1", false, func(k, v []byte) (bool, error) { 
        cnt++
        walkedKeys = append(walkedKeys, string(k))
        return true, nil
    })
```

### Solution

Because of using walkFunc via callbackFunc,Inside Walk func  close leveldb.Release().

### Possible Drawbacks

```
func GetBlockAccountCheckpointByAddr(...) ([]*BlockAccountcheckporint,error) {
...
}
```

- `filepath.WalkFunc` just returns error and `filepath.SkipError` handles it. (https://golang.org/pkg/path/filepath/#pkg-variables)
- WalkFunc can use `IterItem` instead `k,v []byte` too.

### DIFF
--- a/lib/storage/leveldb.go
+++ b/lib/storage/leveldb.go
@@ -328,10 +328,33 @@ func (st *LevelDBBackend) GetIterator(prefix string, option *IteratorOptions) (f
 		})
 }
 
-type WalkFunc func(key, value []byte) (bool, error)
+type (
+	WalkFunc   func(key, value []byte) (bool, error)
+	WalkOption struct {
+		Cursor  string
+		Limit   uint64
+		Reverse bool
+	}
+)
 
-//TODO(anarcher): walkOption? := { prefix,cursor,reverse }
-func (st *LevelDBBackend) Walk(prefix, cursor string, reverse bool, walkFunc WalkFunc) error {
+func NewWalkOption(cursor string, limit uint64, reverse bool) *WalkOption {
+	o := &WalkOption{
+		Cursor:  cursor,
+		Limit:   limit,
+		Reverse: reverse,
+	}
+	return o
+}
+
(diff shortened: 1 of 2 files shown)


kfangw:
Looks good.
But could you explain a better point against as-is iterator for my understanding?
I got one better point which is easy to handle error

Geod24:
I like the concept! However when submitting those PRs, it would be very nice if the usage was demonstrated. It would make grasping the implications much easier for the reviewers.

anarcher:
@spikeekips @Geod24  I do NOT intend to change `LevelDB.Iter` to  `LevelDB.Walk` for now.  #337  works well. I suggest that We consider both of `Iter` or `Walk` when we have to time to add or change `GetXXX`. :->

@kfangw Another advantage of it is that we don't `closeFunc` call when we use it. 

Can I merge it? or more example is needed?

spikeekips:
@anarcher The reason I mentioned #337, is that there are `IteratorOption` in #337, so this 'Walk' also support `IteratorOption` too.

anarcher:
@spikeekips  Great idea!

How about it like below: 
```
func Walk(walkOption *WalkOption,walkFunc WalkFunc) (next bool,err error) 
```

`IteratorOption` is good but `walk` can have own `walkOpiton` (It's just naming taste... haha)

spikeekips:
@anarcher I don't think, `Walk` is not same with `Iterator`. Yes, `Walk` should have it's own options :)

PR #359, opened 2018-09-11
MemoryNetwork: Remove global state

Geod24:
That will make the code simpler and more reliable. Hopefully in the future we can parallelize our tests.


Geod24:
Rebased

lib/block/test.go
```diff
@@ -29,17 +27,6 @@ func init() {
 	kp, _ = keypair.Random()
 }
 
-func createNetMemoryNetwork() (*network.MemoryNetwork, *node.LocalNode) {
```
kfangw: Thx ^^

PR #364, opened 2018-09-11
Refactor, clean up code, and fix some race conditions

Geod24:
So, after this we're down to 5 races left 🎉 
The early commit simplify the code, removing the `defer` fixes a race condition and simplifying `HTTP2Client` prevents another one from crippling back in.


lib/network/connection_manager.go
```diff
@@ -69,7 +69,10 @@ func (c *ConnectionManager) GetConnection(address string) (client NetworkClient)
 }
 
 func (c *ConnectionManager) Start() {
-	go c.connectValidators()
+	c.log.Debug("starting to connect to validators", "validators", c.validators)
+	for _, v := range c.validators {
+		go c.connectingValidator(v)
```
Charleslee522: Is this for performance or for fixing racecondition? :)

Geod24: That's part of the cleaning up code :)
And it should (slighly) improve performance as well :)

lib/common/http2_client.go
```diff
@@ -17,58 +16,43 @@ type HTTP2Client struct {
...
+			CheckRedirect: func(req *http.Request, via []*http.Request) error {
+				return http.ErrUseLastResponse // NOTE prevent redirect
+			},
 		},
 	}
 
 	return
 }
 
-func (c *HTTP2Client) Close() {
```
spikeekips: Is there reason to remove `Close()`? We can use it to close connection from client side by force in keep-alive state

Geod24: I removed it because it is not used anywhere.
Since it is not used anywhere, we cannot make sure any implementation we provide correctly works. If we need it in the future, we can trivially re-implement it.

spikeekips: @Geod24 I see, it is not used but we don't need the existing code like this method, closing connection is important for keep-alive. Please don't remove it.

Geod24: I'll remove this commit then

spikeekips:
@Geod24 You look so busy, I lost cgance to approve this PR :)

Issue #126, opened 2018-06-25
add `race_detector` into go test


Charleslee522:
-race option should be added in running all go tests
https://golang.org/doc/articles/race_detector.html

When this option is added(`SEBAK_LOG_HANDLER=null go test ./... -race`), many tests are failed like below.
Solving these errors seems to be important in the long run.
```
Read at 0x00c4205f0098 by goroutine 30:
   boscoin.io/sebak/vendor/github.com/GianlucaGuarini/go-observable.(*Observable).removeEvent()
       /Users/charles/go/src/boscoin.io/sebak/vendor/github.com/GianlucaGuarini/go-observable/helpers.go:41 +0x13c
   boscoin.io/sebak/vendor/github.com/GianlucaGuarini/go-observable.(*Observable).Off()
       /Users/charles/go/src/boscoin.io/sebak/vendor/github.com/GianlucaGuarini/go-observable/observable.go:74 +0x14a
   boscoin.io/sebak/lib.streaming()
       /Users/charles/go/src/boscoin.io/sebak/lib/api.go:58 +0x5b2
   boscoin.io/sebak/lib.GetAccountHandler.func1()
       /Users/charles/go/src/boscoin.io/sebak/lib/api_account.go:52 +0x3d6
   net/http.HandlerFunc.ServeHTTP()
       /usr/local/opt/go/libexec/src/net/http/server.go:1947 +0x51
   boscoin.io/sebak/vendor/github.com/gorilla/mux.(*Router).ServeHTTP()
       /Users/charles/go/src/boscoin.io/sebak/vendor/github.com/gorilla/mux/mux.go:162 +0x176
   net/http.serverHandler.ServeHTTP()
       /usr/local/opt/go/libexec/src/net/http/server.go:2694 +0xb9
   net/http.(*conn).serve()
       /usr/local/opt/go/libexec/src/net/http/server.go:1830 +0x7dc
```

```
Previous write at 0x00c4205f009a by goroutine 6:
   boscoin.io/sebak/vendor/github.com/GianlucaGuarini/go-observable.(*Observable).dispatchEvent()
       /Users/charles/go/src/boscoin.io/sebak/vendor/github.com/GianlucaGuarini/go-observable/helpers.go:91 +0x251
   boscoin.io/sebak/vendor/github.com/GianlucaGuarini/go-observable.(*Observable).Trigger()
[...]

PR #366, opened 2018-09-11
Get rid of race and enable CI checks

Geod24:
The title says it all. Fixes https://github.com/bosnet/sebak/issues/126

### DIFF
--- a/lib/common/util.go
+++ b/lib/common/util.go
@@ -6,8 +6,6 @@ import (
 	"io"
 	"net/url"
 	"os"
-	"sync"
-	"sync/atomic"
 
 	"encoding/binary"
 	uuid "github.com/satori/go.uuid"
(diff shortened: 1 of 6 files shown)


lib/network/connection_manager.go
```diff
@@ -117,6 +117,13 @@ func (c *ConnectionManager) AllValidators() []string {
 //   the number of validators which are currently connected
 //
 func (c *ConnectionManager) CountConnected() int {
+	c.RLock()
+	defer c.RUnlock()
+	return c.countConnectedUnlocked()
+}
+
+// Ditto
```
spikeekips: Please remove this comment, I think the method name explains well.

lib/common/util.go
```diff
@@ -28,29 +26,6 @@ func GetUniqueIDFromDate() string {
 	return NowISO8601()
 }
 
-type SafeLock struct {
```
spikeekips: Originally I admit to use `SafeLock` was so naive :)

lib/network/connection_manager.go
Geod24: Done

PR #372, opened 2018-09-12
Improve HTTP2Client code

Geod24:
- Remove some unused code
- Cleanup the control flow

### DIFF
--- a/lib/common/http2_client.go
+++ b/lib/common/http2_client.go
@@ -2,7 +2,6 @@ package common
 
 import (
 	"bytes"
-	"context"
 	"crypto/tls"
 	"io"
 	"net"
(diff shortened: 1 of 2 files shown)


lib/common/http2_client.go
```diff
@@ -2,71 +2,58 @@ package common
...
 
 	return
 }
 
 func (c *HTTP2Client) Close() {
-	if c.conn == nil {
-		return
-	}
-	c.conn.Close()
+	c.transport.CloseIdleConnections()
```
spikeekips: When I wrote this code, I want to `Close` client connection at anytime even in use, not idle. Please reflect this requirement.

Geod24: The problem with this class is that a `Client` does not map to a single connection.
See [the documentation](https://golang.org/pkg/net/http/#Client):

> The Client's Transport typically has internal state (cached TCP connections), so Clients should be reused instead of created as needed. Clients are safe for concurrent use by multiple goroutines.

The `Client` should not mess with the connection, it's the `Transport` role.
As for cancelling a request, it should be done with `Context.Cancel` on the `Request`.
So nothing the `Client` should know about.

spikeekips: @Geod24 Okay, I also already know the `Request.WithContext()` way, but I wanted to close in `net.Con`. Anyway you wrote this code to consider context way, it will be good.

PR #378, opened 2018-09-13
Fix MSON models and apply aglio for api rendering

anarcher:
(aglio -i ./API.md -o api.html)

### Solution

Using https://github.com/danielgtaylor/aglio


anarcher:
I will merge it~ @kfangw :->

PR #382, opened 2018-09-13
testcode revises

kfangw:
testcode revised


kfangw:
@anarcher there was a data race T^T...
It is prevented by lock/unlock... 
https://github.com/bosnet/sebak/pull/382/commits/464f7a14fbbdb4ded38215dbdc6d8edaee616b57

Issue #356, opened 2018-09-10
Operation Type Implementation


soonkuk:
##Background##
We defined operation type for the transfer feature(payment, createAccount) until now.
And the methods which is defined in OperationBody interface are Validate(), IsWellFormed(), TargetAddress(), GetAmount().
There is a need to define more operation type increasingly.
But the methods of OperationBody interface isn't general for other operation type.(e.g. operation_voting_result, operation_proposal ...). However OperationBody implementation for voting result should have GetAmount() method to satisfy interface.
It will be good if OperationBody struct doesn't need to implement unnecessary methods repeatedly.

##Solution##
type OperationBody interface {
	Validate(*storage.LevelDBBackend) error
	IsWellFormed([]byte) error
	TargetAddress() string
	GetAmount() common.Amount
}

type OperationBodyImpl struct {
}

func (opb OperationBodyImpl) Validate(*storage.LevelDBBackend) error {
	// To Do Default return
}

func (opb OperationBodyImpl) IsWellFormed([]byte) error {
	// To Do Default return
}

func (opb OperationBodyImpl) TargetAddress() string {
	// To Do Default return
}

func (opb OperationBodyImpl) GetAmount() common.Amount {
	// To Do Default return
}

type OperationBodyVotingResult struct {
	OperationBodyImpl
	Target string        `json:"target"`
	Amount common.Amount `json:"amount"`
}

func (opv OperationBodyVotingResult) Validate(*storage.LevelDBBackend) error {
	// Implementation
}

func (opv OperationBodyVotingResult) IsWellFormed([]byte) error {
	// Implementation
}

Implement default struct about interface.
The default struct define the methods which should be generated.
Each operation type inherit default struct and define the method only needed for it.

kfangw:
It makes sense. 
Could you make a PR for it?
We can discuss further in the PR

soonkuk:
Ok, then I'll make a PR.

Issue #311, opened 2018-09-03
Congress Voting: Messages and API


soonkuk:
Proposal about congress voting and voting result should be stored in the blockchain.
We should implement the transaction operation type about proposal and voting result and define operation data.

kfangw:
May I take this Issue? @soonkuk

kfangw:
https://blockchainos.atlassian.net/wiki/spaces/BOS/pages/556302380/Congress+Voting+Messages+and+API

PR #389, opened 2018-09-13
Organize operation type (#356)

soonkuk:
### Background
We defined operation type for the transfer feature(`payment`, `createAccount`) until now.
And the methods which is defined in `OperationBody` interface are `Validate()`, `IsWellFormed()`, `TargetAddress()`, `GetAmount()`.
There is a need to define more operation type increasingly.
But the methods of `OperationBody` interface isn't general for other operation type.(e.g. `operation_voting_result`, `operation_proposal` ...). However `OperationBody` implementation for voting result should have GetAmount() method to satisfy interface.
It will be good if `OperationBody` struct doesn't need to implement unnecessary methods repeatedly and use current `OperationBody` interface continually.

### Solution
Implement default struct about interface.
The default struct define the methods which should be generated.
Each operation type inherit default struct and define the method only needed for it.

### DIFF
--- a/lib/transaction/operation.go
+++ b/lib/transaction/operation.go
@@ -15,6 +15,9 @@ type OperationType string
 const (
 	OperationCreateAccount OperationType = "create-account"
 	OperationPayment                     = "payment"
+	OperationVotingResult                = "voting-result"
+	OperationProposal                    = "proposal"
+	OperationMembership                  = "membership"
 )
 
 type Operation struct {
@@ -126,4 +129,21 @@ type OperationBody interface {
 	IsWellFormed([]byte) error
 	TargetAddress() string
 	GetAmount() common.Amount
+}
+
+type OperationBodyImpl struct {
+}
+
+func (ob OperationBodyImpl) IsWellFormed(networkID []byte) (err error) {
+	err = nil
+	return
+}
+
+func (ob OperationBodyImpl) TargetAddress() string {
+	return "none"
+}
+
+func (ob OperationBodyImpl) GetAmount() (a common.Amount) {
+	a = 0
+	return
 }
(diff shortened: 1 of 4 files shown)


Geod24:
I'm not sure I see why we need `BodyType` ?
Since `Type` already allows us to know what the BodyType is.

soonkuk:
I think all operations can be categorized to two type currently.
One is operations about remittance and the other is operations about storing data in the blockchain.
Operations about remittance is create account, payment, freeze ...
and operations about storing data is voting result, proposal ...
The main difference of this operation is largely whether they have a target.
According to this large type, the operation is treated differently in library.
And also there is need to categorize the detailed type of operation like as create-accout, payment ...
So I make the `type` and `BodyType` for operation struct.

Geod24:
> One is operations about remittance and the other is operations about storing data in the blockchain.

But remittance do store data in the blockchain.
Even if we go by your definition, those operation do not overlap. An operation is either one, or the other category. Which is why they can be part of a single enumeration, and the second field is not needed.

kfangw:
@soonkuk you'd better use `rebase -i HEAD~x` to amend your commits

soonkuk:
I removed bodyType and modified operationBody implementation.

lib/transaction/operation.go
```diff
@@ -127,3 +130,20 @@ type OperationBody interface {
 	TargetAddress() string
 	GetAmount() common.Amount
 }
+
+type OperationBodyImpl struct {
```
kfangw: Not necessary

kfangw:
@soonkuk How is going on this PR?
I will add 2 new operations for Congress voting #311

soonkuk:
I talked with @kfangw and realized that there is no reason to implement this.
So I will close this PR.

kfangw:
> I talked with @kfangw and realized that there is no reason to implement this.
> So I will close this PR.

@soonkuk Please check  #445 and this commit https://github.com/bosnet/sebak/pull/445/commits/9d144f85e3ecc4307a1407dfade0517384a2328a

PR #387, opened 2018-09-13
Implement frozen accounts

Geod24:
Fixes #307 

Missing some tests, so I might address #386 sooner than later.
The only thing missing is the actual 2 weeks delay, which is part of #309

PR #405, opened 2018-09-17
Do not panic on wrong operation body

Geod24:
When sending an invalid Operation (e.g. with no target),
this function would panic on the type assertion.

### DIFF
--- a/lib/transaction/operation.go
+++ b/lib/transaction/operation.go
@@ -66,28 +66,33 @@ func NewOperationFromBytes(b []byte) (op Operation, err error) {
 	return
 }
 
-func NewOperationFromInterface(oj OperationFromJSON) (op Operation, err error) {
+func NewOperationFromInterface(oj OperationFromJSON) (Operation, error) {
+	var op Operation
 	op.H = oj.H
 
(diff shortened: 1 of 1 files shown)


lib/transaction/operation.go
```diff
@@ -66,25 +66,28 @@ func NewOperationFromBytes(b []byte) (op Operation, err error) {
...
-		op.B = NewOperationBodyCreateAccount(body["target"].(string), amount)
 	case OperationPayment:
-		var amount common.Amount
-		amount, err = common.AmountFromString(fmt.Sprintf("%v", body["amount"]))
-		if err != nil {
-			return
+		if amount, err := common.AmountFromString(fmt.Sprintf("%v", body["amount"])); err != nil {
+			return Operation{}, err
+		} else if target, is_target := body["target"].(string); !is_target {
+			return Operation{}, err
```
kfangw: `err` may `nil`. Is what you intended?

Geod24: Good catch, it's a mistake! Fixed

Geod24:
This is blocking https://github.com/bosnet/sebak/pull/387

PR #408, opened 2018-09-17
API: Get Block data for Nodes

### LINKED ISSUE #301: New API: Get `Block`s and get `Header`s of `Block`
(related to #298)

We need two new API, getting `Block`s and `Header`s of `Block` for **SYNC**.

> In this issue, it is not matter what is proper the structure of `Header(block_header.go)` for 'SYNC'. It will be handled in #298 . This issue will make APIs for #298 .

### `/node/blocks`

Request and response the `Block(block.go)`

#### Options
| name | type | description |
| --- | --- | --- |
| `height-range` | int-int | set range of `Block.Height`, ex) `9-348`(from `9` to `348` |
| `hash` | `Hash` string | set `Block.Hash`, this can be set multiple |
| `full` | boolean | if set `1`, the response item will have all available data of `Block`, `Transaction`s and it's `Operation`s too. Default is `0`, will response the only serialized `Block` only |

Examples,

* `/node/blocks?height-range=9-348`
* `/node/blocks?height-range=9-348&full=1`
* `/node/blocks?hash=AAA&hash=BBB&hash=CCC`
* `/node/blocks/?height-range=9-348&hash=DDD`

### `/node/blocks/headers`

Request and response the `Header(block_header.go)`.

#### Options
| name | type | description |
| --- | --- | --- |
| `height-range` | int-int | set range of `Block.Height`, ex) `9-348`(from `9` to `348` |
| `hash` | `Hash` string | set `Block.Hash`, this can be set multiple |

This is examples,

* `/node/blocks/headers?height-range=9-348`
* `/node/blocks/headers?hash=AAA&hash=BBB&hash=CCC`
* `/node/blocks/headers?height-range=9-348&hash=DDD`

#### Response
JSON serialized string of `Header(block_header.go)`.

> These API can handle long list of items, so we need the paging feature(#122 )

spikeekips:
### Github Issue

resolves #301 

## Major Changes

* `runner.GetBlocksHandler()` added. #301
* `IteratorOptions` -> `ListOptions`
    - `ListOptions` was changed to interface, `DefaultListOptions` was newly created.
    - `DefaultListOptions` supports method chaining.
    - `parseQueryString` was removed, instead of it, `storage.NewDefaultListOptionsFromQuery` was newly added
* When `Block` is saved, new key will be stored by `GetBlockKeyPrefixHeight()`
* Testing, `block/block_test.go` was added for `block/block.go`
* `storage/leveldb.go` was refactored to return `errors.ErrorStorageCoreError` for native leveldb error.

### DIFF
--- a/lib/storage/options.go
+++ b/lib/storage/options.go
@@ -1,0 +1,115 @@
+package storage
+
+import (
+	"net/url"
+	"strconv"
+
+	"boscoin.io/sebak/lib/common"
+)
+
+var DefaultMaxLimitListOptions uint64 = 100
+
+type ListOptions interface {
+	Reverse() bool
+	SetReverse(bool) ListOptions
+	Cursor() []byte
+	SetCursor([]byte) ListOptions
+	Limit() uint64
+	SetLimit(uint64) ListOptions
+	Template() string
+	URLValues() url.Values
+}
+
+type DefaultListOptions struct {
+	reverse bool
+	cursor  []byte
+	limit   uint64
+}
+
+func NewDefaultListOptions(reverse bool, cursor []byte, limit uint64) *DefaultListOptions {
+	return &DefaultListOptions{
+		reverse: reverse,
+		cursor:  cursor,
+		limit:   limit,
+	}
+}
+
+// NewDefaultListOptionsFromQuery makes ListOptions from url.Query.
+func NewDefaultListOptionsFromQuery(v url.Values) (options *DefaultListOptions, err error) {
+	var reverse bool
+	var cursor []byte
+	var limit uint64 = DefaultMaxLimitListOptions
+
+	r := v.Get("reverse")
+	if len(r) > 0 {
+		if reverse, err = common.ParseBoolQueryString(r); err != nil {
+			return
+		}
+	}
+
+	r = v.Get("cursor")
+	if len(r) > 0 {
+		cursor = []byte(r)
+	}
+
+	r = v.Get("limit")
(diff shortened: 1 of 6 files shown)


lib/storage/options.go
```diff
@@ -0,0 +1,115 @@
...
+import (
+	"net/url"
+	"strconv"
+
+	"boscoin.io/sebak/lib/common"
+)
+
+var DefaultMaxLimitListOptions uint64 = 100
+
+type ListOptions interface {
```
kfangw: ❤️

lib/block/block.go
```diff
@@ -120,6 +131,11 @@ func (b Block) Save(st *storage.LevelDBBackend) (err error) {
 	if err = st.New(b.NewBlockKeyConfirmed(), b.Hash); err != nil {
 		return
 	}
+	if err = st.New(GetBlockKeyPrefixHeight(b.Height), b.Hash); err != nil {
+		return
+	}
+
+	observer.BlockObserver.Trigger(EventBlockPrefix, b)
```
kfangw: It triggers but nowhere listens to this event. why?

lib/network/api/stream.go
```diff
@@ -100,6 +100,11 @@ func (s *EventStream) Render(args ...interface{}) {
 		bs = payload
 	}
 
+	if !s.rendered {
```
anarcher: Why `s.rendered` needs?

PR #411, opened 2018-09-17
Change doc README for api doc

anarcher:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,5 +1,43 @@
 # API documents with [API blueprint](https://apiblueprint.org/)
 
+- https://apiblueprint.org/documentation/tutorial.html
+
+Install [aglio](https://github.com/danielgtaylor/aglio):
+
+```
+brew install node
+npm config set prefix '~/.npm-global'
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -1,5 +1,43 @@
...
 
+- https://apiblueprint.org/documentation/tutorial.html
+
+Install [aglio](https://github.com/danielgtaylor/aglio):
+
+```
+brew install node
+npm config set prefix '~/.npm-global'
+export PATH=~/.npm-global/bin:$PATH
+npm install -g aglio
```
outsideris: I think global installation is not good for non-nodejs developers.

There are two options.

1. Intall locally `npm install aglio`. and then you can run `./node_modules/.bin/aglio`.
2. Just use `npx aglio` if there is npx command in nodejs. It run `aglio` instantly with installation.

kfangw: `npx aglio` is better!

anarcher: I will change it with `npx aglio`

PR #413, opened 2018-09-17
Cleanup: Remove some Node function, refactor `handleMessage`

Geod24:
Pretty straightforward stuff. I think ultimately we could merge `Validator` and `LocalNode` into `Node`, if we take the Keypair out of `LocalNode`, but I'll leave that for another day.


kfangw:
defer to spike and charles.

PR #421, opened 2018-09-18
Get The Missing Transactions From Proposer

### LINKED ISSUE #212: API: Get transactions from proposer

spikeekips:
## Github Issue

resolves #212 

## Background

Under the current impl of ISAAC, the incoming `Transaction`s will be
broadcasted, but not verified that the target node really received them. To fill
these missing `Transaction`s, the node will request the unknown `Transaction`s
of the `Ballot` of proposer node. This PR is for it.

### `GET|POST /node/transactions`

* Basically the response will be same with `/node/blocks`. For more details #301 . 

#### Options

* `GET`

| name | type | description |
| -- | -- | -- |
| hash | `Hash` string | `Transaction` hash |

* `POST`

Set the list of hash like this,
```
[
  "AAA",
  "BBB",
  "CCC",
  "DDD"
]
```

Document type must be `application/json`. If list of hash is found in query string and post data, they will be merged :)

### Response

```
<NodeItemDataType> <json marshaled data>
```

In this api, there will be 2 `NodeItemDataType`, "transaction"(`NodeItemTransaction`) and "error"(`NodeItemError`).

```
transaction {"T":"transaction","H":{"version":"","created":"2018-09-18T09:52:53.702626239+09:00","hash":"FTJAPEKHNybgFWoBeoMCBuS47Y1Hb85c99eBQQ1ADA99","signature":"4hwYT5uns6QWPajTfXjmMQxbUHx6EuapUxcc83SDDVtd4nJj4JToZuAhR8khAjWPMCdRMC8RFi8oy86bvGzjBd9S"},"B":{"source":"GAPRM34PEEMMC5W4WFLUMGS3WSI2D3HRUSTBSIGHD2EBN7T35EBHTTGO","fee":"10000","sequenceID":0,"operations":[{"H":{"type":"payment"},"B":{"target":"GCTOG5UDXYWJHCR3VYMX3OGSZ4EWNJ56FOZG2U3HTKGQ3WRX3VHM65WI","amount":"495"}},{"H":{"type":"payment"},"B":{"target":"GBT7J5ZDR6COPS6IRMDC5MNZKNSKIP5UASEBG375OQ3JILB52YKVWB5X","amount":"466"}}]}}
```

To parse the response item, `runner.UnmarshalNodeItemResponse` can explain.

### DIFF
--- a/lib/error/base.go
+++ b/lib/error/base.go
@@ -5,6 +5,7 @@ import "encoding/json"
 type Error struct {
 	Code    uint   `json:"code"`
 	Message string `json:"message"`
+	Data    map[string]interface{}
 }
 
 func (o *Error) Serialize() (b []byte, err error) {
@@ -17,6 +18,28 @@ func (o *Error) Error() string {
 	return string(b)
 }
 
+// SetData sets `Error.Data`
+func (o *Error) SetData(k string, v interface{}) *Error {
+	o.Data[k] = v
+
+	return o
+}
+
+func (o *Error) Clone() *Error {
+	data := map[string]interface{}{}
+	if o.Data != nil {
+		for k, v := range o.Data {
+			data[k] = v
+		}
+	}
+
+	return &Error{
+		Code:    o.Code,
+		Message: o.Message,
+		Data:    data,
+	}
+}
+
 func NewError(code uint, message string) *Error {
-	return &Error{Code: code, Message: message}
+	return &Error{Code: code, Message: message, Data: map[string]interface{}{}}
 }
--- a/lib/node/runner/api_block.go
+++ b/lib/node/runner/api_block.go
@@ -8,18 +8,20 @@ import (
 
 	"boscoin.io/sebak/lib/block"
 	"boscoin.io/sebak/lib/error"
+	"boscoin.io/sebak/lib/transaction"
 )
 
 const GetBlocksPattern = "/blocks"
 
-type GetBlocksDataType string
+type NodeItemDataType string
 
 const (
-	GetBlocksDataTypeBlock       GetBlocksDataType = "block"
-	GetBlocksDataTypeHeader      GetBlocksDataType = "header"
-	GetBlocksDataTypeTransaction GetBlocksDataType = "transaction"
-	GetBlocksDataTypeOperation   GetBlocksDataType = "operation"
-	GetBlocksDataTypeError       GetBlocksDataType = "error"
+	NodeItemBlock            NodeItemDataType = "block"
+	NodeItemBlockHeader      NodeItemDataType = "block-header"
+	NodeItemBlockTransaction NodeItemDataType = "block-transaction"
+	NodeItemBlockOperation   NodeItemDataType = "block-operation"
+	NodeItemTransaction      NodeItemDataType = "transaction"
+	NodeItemError            NodeItemDataType = "error"
 )
 
 func (nh NetworkHandlerNode) GetBlocksHandler(w http.ResponseWriter, r *http.Request) {
@@ -108,13 +110,13 @@ func (nh NetworkHandlerNode) GetBlocksHandler(w http.ResponseWriter, r *http.Req
 	w.Header().Set("X-SEBAK-RESULT-COUNT", string(len(bs)))
 
 	for _, b := range bs {
-		var itemType GetBlocksDataType
+		var itemType NodeItemDataType
 		if options.Mode == GetBlocksOptionsModeHeader {
-			itemType = GetBlocksDataTypeHeader
-			renderGetBlocksItem(w, itemType, b.(*block.Block).Header)
+			itemType = NodeItemBlockHeader
+			nh.renderNodeItem(w, itemType, b.(*block.Block).Header)
 		} else {
-			itemType = GetBlocksDataTypeBlock
-			renderGetBlocksItem(w, itemType, b.(*block.Block))
+			itemType = NodeItemBlock
+			nh.renderNodeItem(w, itemType, b.(*block.Block))
 		}
 
 		if options.Mode == GetBlocksOptionsModeFull {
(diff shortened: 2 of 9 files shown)


lib/node/runner/api_transaction.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"boscoin.io/sebak/lib/block"
+	"boscoin.io/sebak/lib/common"
+	"boscoin.io/sebak/lib/error"
+)
+
+const GetTransactionPattern string = "/transactions"
+
+func (nh NetworkHandlerNode) GetNodeTransactionsHandler(w http.ResponseWriter, r *http.Request) {
+	hashes := r.URL.Query()["hash"]
```
kfangw: As you know client API support url like this
`/transactions/{hash}`
At this point, you use a query instead. Is there any reason for that?
I know that it is not a big deal because the pattern does not needed to follow REST, just curious.

lib/error/base.go
```diff
@@ -5,6 +5,7 @@ import "encoding/json"
 type Error struct {
 	Code    uint   `json:"code"`
 	Message string `json:"message"`
+	Data    map[string]interface{}
```
kfangw: Is it only for the API?

lib/node/runner/api_transaction.go
```diff
@@ -0,0 +1,91 @@
...
+			unknown[hash] = true
+			continue
+		} else {
+			nh.renderNodeItem(w, NodeItemTransaction, tx)
+		}
+	}
+
+	// check in `BlockTransaction`
+	for _, hash := range hashes {
+		if _, found := unknown[hash]; !found {
```
kfangw: for this purpose, values of the map `unknown` are meaningless.
IMHO, empty struct is better. 
`unknown := map[string]struct{}`
`unknown[hash] = struct{}{}`

lib/node/runner/api_transaction.go
```diff
@@ -0,0 +1,91 @@
...
+			continue
+		}
+
+		if exists, err := block.ExistBlockTransaction(nh.storage, hash); !exists || err != nil {
```
kfangw: How about change if blocks like this to enhance readability?
```
		if exists, err := block.ExistBlockTransaction(nh.storage, hash); err != nil {
			nh.renderNodeItem(w, NodeItemError, err)
			return
		} else if !exists {
			err := errors.ErrorTransactionNotFound.Clone().SetData("hash", hash)
			nh.renderNodeItem(w, NodeItemError, err)
			continue
		} 
```

AFAIK, if conditions share the same basic block. Hence, a variable declared in if clause can be referred in else if clause

lib/node/runner/api_transaction.go
```diff
@@ -0,0 +1,91 @@
...
+			nh.renderNodeItem(w, NodeItemError, err)
+			return
+		}
+
+		btx, err := block.GetBlockTransaction(nh.storage, hash)
+		if err != nil {
+			nh.renderNodeItem(w, NodeItemError, err)
+			return
+		}
+		fmt.Println(">>", string(btx.Message))
```
kfangw: please remove the debugging message

lib/node/runner/api_transaction.go
spikeekips: @kfangw In proposed `Ballot` there can be up to `1000`(`MaxTransactionsInBallot`) transactions, so in theory the node should request up to `1000` transactions from proposer. To request huge list of `hash`, `/transactions/{hash}` is not acceptable, I think.

lib/error/base.go
spikeekips: @kfangw At this time. :)

lib/node/runner/api_transaction.go
spikeekips: @kfangw 👍

lib/node/runner/api_transaction.go
kfangw: Agreed.

lib/node/runner/api_transaction.go
spikeekips: @kfangw It looks like matter of choice, but I will apply your comment. :)

lib/node/runner/api_transaction.go
spikeekips: @kfangw :)

anarcher:
```
Response
<NodeItemDataType> <json marshaled data>
```
It means that the response's `Content-type: application/json` is not?

lib/node/runner/api_block.go
```diff
@@ -144,17 +146,7 @@ func (nh NetworkHandlerNode) GetBlocksHandler(w http.ResponseWriter, r *http.Req
...
-	}
-
-	w.Write(append([]byte(itemType+" "), append(s, '\n')...))
-}
-
-func UnmarshalGetBlocksHandlerItem(d []byte) (itemType GetBlocksDataType, b interface{}, err error) {
+func UnmarshalNodeItemResponse(d []byte) (itemType NodeItemDataType, b interface{}, err error) {
 	sc := bufio.NewScanner(bytes.NewReader(d))
 	sc.Split(bufio.ScanWords)
 	sc.Scan()
```
anarcher: Maybe We would check `sc.Err()` :->

spikeekips: @anarcher 👍

lib/node/runner/api_block.go
```diff
@@ -108,13 +110,13 @@ func (nh NetworkHandlerNode) GetBlocksHandler(w http.ResponseWriter, r *http.Req
 	w.Header().Set("X-SEBAK-RESULT-COUNT", string(len(bs)))
 
 	for _, b := range bs {
```
anarcher: I have a question.  Why is the type of `bs` not `block.Block` but `interface{}`?   :->

spikeekips: @anarcher Yes, `bs` is `[]interface{}`, because `bs` handles `block.Block` and `block.Header`.

anarcher: Humm,  Current Three calls about  get block from storage looks like to get `block.Block`..  Is it right?  https://github.com/bosnet/sebak/pull/421/files#diff-e805ed6b5e4b972f13e0939538ffb002R60
- https://github.com/bosnet/sebak/pull/421/files#diff-e805ed6b5e4b972f13e0939538ffb002R86
- https://github.com/bosnet/sebak/pull/421/files#diff-e805ed6b5e4b972f13e0939538ffb002R97

And render too.
``` 
nh.renderNodeItem(w, itemType, b.(*block.Block).Header)
nh.renderNodeItem(w, itemType, b.(*block.Block))
```

spikeekips: @anarcher Haha, you pointed I missed. You right.

PR #423, opened 2018-09-18
Limit the maximum number of Operations in one Transaction

### LINKED ISSUE #418: Limit the maximum number of Operations in one Transaction
(similar to the #243 )

Like capacity of maximum number of `Transaction`s in one `Ballot`, the number of `Operation`s in one `Transaction` should be limited.`

spikeekips:
### Github Issue
resolves #418 

### Solution

`common.MaxOperationsInTransaction` defined, it's value is set to `1000` by default.


lib/transaction/transaction.go
```diff
@@ -84,6 +84,7 @@ func NewTransaction(source string, sequenceID uint64, ops ...Operation) (tx Tran
 }
 
 var TransactionWellFormedCheckerFuncs = []common.CheckerFunc{
+	CheckTransactionMaxOperations,
```
Charleslee522: It seems like confirming that it is full.
How about this `CheckTransactionOverOperationLimit`?

spikeekips:
@Charleslee522 I will check, thanks :)

lib/common/constant.go
```diff
@@ -18,4 +18,7 @@ var (
 	// MaxTransactionsInBallot limits the maximum number of `Transaction`s in
 	// one proposed `Ballot`.
 	MaxTransactionsInBallot int = 1000
+	// MaxOperationsInTransaction limits the maximum number of `Operation`s in
+	// one `Transaction`.
+	MaxOperationsInTransaction int = 1000
```
Geod24: That should be `const`

lib/transaction/transaction_test.go
```diff
@@ -87,3 +88,18 @@ func TestIsWellFormedTransactionWithInvalidSignature(t *testing.T) {
...
 }
+
+func TestIsWellFormedTransactionMaxOperationsInTransaction(t *testing.T) {
+	var err error
+
+	MaxOperationsInTransaction := common.MaxOperationsInTransaction
+	defer func() {
+		common.MaxOperationsInTransaction = MaxOperationsInTransaction
+	}()
+	common.MaxOperationsInTransaction = 2
```
Geod24: We should not modify global state like this... It breaks abstraction big time and prevents the tests from running in parallel.

Either:
- We make 1001 operations to test
- We make it a parameter to the function

lib/common/constant.go
spikeekips: @Geod24 `const`, it makes sense, but I want to open to change `MaxOperationsInTransaction` value. I don't know which is better, `var` or `const`.

lib/transaction/transaction_test.go
spikeekips: @Geod24 Yes, you right. In parallel env, it makes problem.

lib/transaction/transaction_test.go
```diff
@@ -87,3 +88,12 @@ func TestIsWellFormedTransactionWithInvalidSignature(t *testing.T) {
...
 	require.NotNil(t, err)
 }
+
+func TestIsWellFormedTransactionMaxOperationsInTransaction(t *testing.T) {
+	var err error
+
+	_, tx := TestMakeTransaction(networkID, common.MaxOperationsInTransaction+1)
+	err = tx.IsWellFormed(networkID)
+	require.NotNil(t, err)
+	require.Equal(t, errors.ErrorTransactionHasOverMaxOperations.Code, err.(*errors.Error).Code)
```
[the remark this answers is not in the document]
spikeekips: @Geod24 You mean, the `tx` really has the operations, over limit?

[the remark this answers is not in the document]
Geod24: I mean that a test which tests a failure should also test a success.
Because if you test that something fails, you cannot be sure why it fails unless you make it pass afterwards.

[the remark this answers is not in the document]
spikeekips: @Geod24 There are already tests for valid transaction in `transaction_test.go`.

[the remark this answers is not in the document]
Geod24: The point is not to test the valid transaction, it's to make sure that this test fails for the *right* reason.

lib/transaction/transaction_test.go
```diff
@@ -87,3 +88,20 @@ func TestIsWellFormedTransactionWithInvalidSignature(t *testing.T) {
...
 }
+
+func TestIsWellFormedTransactionMaxOperationsInTransaction(t *testing.T) {
+	var err error
+
+	{ // over common.MaxOperationsInTransaction
+		_, tx := TestMakeTransaction(networkID, common.MaxOperationsInTransaction+1)
+		err = tx.IsWellFormed(networkID)
+		require.NotNil(t, err)
+		require.Equal(t, errors.ErrorTransactionHasOverMaxOperations.Code, err.(*errors.Error).Code)
```
Geod24: Actually this could just be:
```
err = tx.IsWellFormed(networkID)
require.Equal(t, errors.ErrorTransactionHasOverMaxOperations, err)
```

No ?

spikeekips: @Geod24 Yes, it can be :)

Issue #394, opened 2018-09-14
API: query parameter for `type`


kfangw:
as-is query parameters are limit, cursor, and reverse.

To get a specific operation such as the proposal, type query is needed

Things to do are

* Make a reverse index for the type

* Query parameter `type` support for API

PR #427, opened 2018-09-18
filter type for operation API

kfangw:
### Github Issue
#394 

### Background
#356 #311 
Operation API support query for `type`

### Solution
There are 2 alternatives

1. reverse index for operation
Write Heavy, Read Light
Make one more set for operation index with `type-xxx` prefix

2. filter out not matched
Write Light, Read Heavy
No index added
Filter out not matched object at the handler

IMO,
This usage is rarely called.
Hence, I adopted the second option

### Possible Drawbacks
Or... Is it really needed?

### DIFF
--- a/lib/network/api/operation.go
+++ b/lib/network/api/operation.go
@@ -12,7 +12,7 @@ import (
 	"boscoin.io/sebak/lib/network/api/resource"
 	"boscoin.io/sebak/lib/network/httputils"
 	"boscoin.io/sebak/lib/storage"
-	"boscoin.io/sebak/lib/transaction"
+	"boscoin.io/sebak/lib/transaction/operation"
 	"strings"
 )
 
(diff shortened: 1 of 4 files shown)


spikeekips:
@kfangw I think this feature will be very useful :)

spikeekips:
@kfangw Filtering by operation type && account, right? I expected filtering only by operation type, so the api path will be `/api/operations/{type}`, or `/api/operations/?type={type}` like this.

kfangw:
@spikeekips Filtering by one condition is easy to achieve.
But, for 2+ conditions, we need to consider the alternatives what I mentioned above(issue description).

When the API support only one condition like ```/api/operations/?type={type}```,
Filtering again according to account should be done by the client.

What I concern is the performance for the second option >2. filter out not matched
The filtering is done at the application level, not the database level.

Any good idea for filtering with 2+ conditions? @spikeekips

spikeekips:
@kfangw I don't intend 2 type condition :) I just mean,

* `/api/operations/?type=create-account`: all operations, it's type is `create-account`
* `/api/operations/?type=proposal`
* `/api/operations/?type=voting-result`

This kind of stuffs.

kfangw:
@spikeekips 
There haven't existed ```/api/operations/``` API
I thought that without related account or transaction, there are no usages for getting operations
But, what you pointed make sense.
Pass all operations with one condition and shift filtering work on the client.

I will revise this PR.

lib/network/api/operation.go
```diff
@@ -12,12 +12,16 @@ import (
...
 	"boscoin.io/sebak/lib/storage"
+	"boscoin.io/sebak/lib/transaction"
 )
 
 func (api NetworkHandlerAPI) GetOperationsByAccountHandler(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	address := vars["id"]
 	options, err := storage.NewDefaultListOptionsFromQuery(r.URL.Query())
+
+	oType := transaction.OperationType(r.URL.Query().Get("type"))
```
anarcher: Do We need to check a wrong `type` such as `aaa`?

Issue #412, opened 2018-09-17
Hash malleability issue because of JSON serialization


Geod24:
Today @soonkuk approached me with an issue regarding [a WIP commit](https://github.com/soonkuk/sebak/commit/5b1c22ae32c2f581651ef4566fb8af415e1a46c1)

The integration tests were failing for no apparent reason. From the code it was obvious this came from [adding an empty struct to the operation body](https://github.com/soonkuk/sebak/commit/5b1c22ae32c2f581651ef4566fb8af415e1a46c1#diff-64dcdb85b4f0936552aff214b5b9cbcaR12) which changed the hash.

Indeed, the transaction that sebak would generate was:
```json
{
  "T": "transaction",
  "H": {
    "version": "",
    "created": "2018-09-17T02:28:45.183923554Z",
    "hash": "GEgcpwUHVrSM5McEsccfSdnM3r4MALX3SbfShcHdmAQK",
    "signature": "3yXsyrCnarBz1azSd1ymhPjmnRjDz8Mo4oydoJEaWPsMz4rVmLmvYwfM315z7hy5fep9sd3s2JGHjobAiAN9mdjm"
  },
  "B": {
    "source": "GDIRF4UWPACXPPI4GW7CMTACTCNDIKJEHZK44RITZB4TD3YUM6CCVNGJ",
    "fee": "10000",
    "sequenceID": 0,
    "operations": [
      {
        "H": {
          "type": "create-account"
        },
        "B": {
          "target": "GDTEPFWEITKFHSUO44NQABY2XHRBBH2UBVGJ2ZJPDREIOL2F6RAEBJE4",
          "amount": "10000000000"
        }
      }
    ]
  }
}
```

while the transaction recorded in the test suite was:
```json
{
  "T": "transaction",
  "H": {
    "version": "",
    "created": "2018-01-01T00:00:00.000000000Z",
    "hash": "B9n71T2RwHPFY7uyZKZh9tXCdRCHAngakoZQXw5ZExfQ",
    "signature": "5zdcFUaYAXZ7ZNyKWzfw2b9GENS65mFtgXC4KVcXhJWAARGAHLa1faewwdCEMuQVgaHQojd6r86svXP8xn4zKvCT"
  },
  "B": {
    "source": "GDIRF4UWPACXPPI4GW7CMTACTCNDIKJEHZK44RITZB4TD3YUM6CCVNGJ",
    "fee": "10000",
    "sequenceID": 0,
    "operations": [
      {
[...]

PR #428, opened 2018-09-18
Transaction: Do not rely on sender's hash

Geod24:
```
It does not make sense to include the tx hash in the query,
because we have to validate it anyway.
Instead, just omit it and calculate it on deserialization,
saving both bandwidth and CPU time.
```

Found when working on #387 and #412 
It will also make tests easier to update :)

### DIFF
--- a/lib/node/runner/api_block.go
+++ b/lib/node/runner/api_block.go
@@ -179,9 +179,7 @@ func UnmarshalNodeItemResponse(d []byte) (itemType NodeItemDataType, b interface
 		err = unmarshal(&t)
 		b = t
 	case NodeItemTransaction:
-		var t transaction.Transaction
-		err = unmarshal(&t)
-		b = t
+		b, err = transaction.NewTransactionFromJSON(d[len(sc.Bytes())+1:])
 	case NodeItemError:
 		var t errors.Error
 		err = unmarshal(&t)
--- a/lib/node/runner/checker_message_test.go
+++ b/lib/node/runner/checker_message_test.go
@@ -71,7 +71,7 @@ func TestMessageChecker(t *testing.T) {
 	require.Equal(t, err, errors.ErrorNewButKnownMessage)
 }
 
-func TestMessageCheckerWithInvalidMessage(t *testing.T) {
+func TestMessageCheckerWithInvalidHash(t *testing.T) {
 	_, invalidTx := transaction.TestMakeTransaction(networkID, 1)
 	invalidTx.H.Hash = "wrong hash"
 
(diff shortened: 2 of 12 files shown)


Geod24:
> I understand the purpose of this PR, but I still don't convince `Hash` can be removed.

Can you elaborate ?

Geod24:
@spikeekips : The question is why ? What are the advantages of setting it client-side for the node ?

spikeekips:
@Geod24 You should also explain, the strong reason without `Hash`. The current impl with `Hash` works.

Geod24:
@spikeekips : That's describe in the PR. But I'll be a little more verbose here:

Having `Hash` in the query does not provide any advantage for the node, because the node is going to calculate this hash anyway based on the body. Then it is going to compare this against the value the client sent, and reject the transaction if the client sent the wrong value.
However, the client sending the wrong value does not mean anything wrong *as long as the signature is correct*.

So having an incorrect `Hash` sent to us will not have any negative impact on the consensus, and as this PR shows, we could just replace the value because the *signature* is what really matters.
However, this validation has some downsides:
- It's extra code / work, for no benefit
- It's extra bandwidth
- It makes writing tests harder, because in addition to the signature, the hash also needs to change

I hope this makes the matter clearer.

spikeekips:
@Geod24 @kfangw I also understand what this PR trying to do, but I am not convinced yet. 

> @Geod24 : 
> It's extra code / work, for no benefit
> It's extra bandwidth
> It makes writing tests harder, because in addition to the signature, the hash also needs to change

In the node side, the generating `Hash` must be done regardless of the existence of `Hash` in incoming Transaction. The matter is client side. I also agree the current impl needs extra bandwidth comparing to this PR. You mean, the `Transaction.Hash` is useless, but we check the well-formed of Transaction by `Hash` directly. With `Hash` we can check it additionally before *calculating* `Hash` again in node side.

Geod24:
> With `Hash` we can check it additionally *before* calculating `Hash` again in node side.

How do you check `Hash` is correct *before* calculating `Hash` ? You have to compute `Hash` node-side to do that check...

> The matter is client side.

You can see it this way: `Hash` is already checked, because if `Hash` was incorrect, the `Signature` validation would not pass.

spikeekips:
@Geod24 With `Transaction.Hash` we can verify the signature is correctly signed without generate it again and then, we also should check the `Hash` is correctly made.

Geod24:
@spikeekips : But what is the point of doing the signature check before the hash check ?

spikeekips:
> @Geod24 : But what is the point of doing the signature check before the hash check ?

I mean, using `Hash` of incoming transaction, we can verify `Signature` with it without generating `Hash` from `TransactionBody`.

PR #432, opened 2018-09-18
Fix and improve sequence ID

Geod24:
There was a slight bug which led to valid transactions being rejected.
This'll fix it.


Charleslee522:
Which part had the bug?

Geod24:
The first commit fixes the bug (that's why the tests are modified)

anarcher:
LGTM! :-)

PR #441, opened 2018-09-19
Fix #386: Add unittests for ValidateOp / ValidateTx

### LINKED ISSUE #386: Write unittests for ValidateOp / ValidateTx
Those two methods are some of the most critical to sebak, as they are core part of the consensus rules.
They should be properly documented and tested, but currently no error case is tested:
https://codecov.io/gh/bosnet/sebak/src/master/lib/node/runner/checker_ballot_transaction.go#L148

Geod24:
Better reviewed commit-by-commit.

### DIFF
--- a/lib/node/runner/checker_ballot_transaction.go
+++ b/lib/node/runner/checker_ballot_transaction.go
@@ -141,11 +141,24 @@ func BallotTransactionsSourceCheck(c common.Checker, args ...interface{}) (err e
 	return
 }
 
-// Validate checks,
-// * source account exists
-// * sequenceID is valid
-// * source has enough balance to pay
-// * and it's `Operations`
+//
+// Validate the entirety of a transaction
+//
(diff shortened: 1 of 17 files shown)


lib/node/runner/checker_ballot_transaction.go
```diff
@@ -141,11 +141,24 @@ func BallotTransactionsSourceCheck(c common.Checker, args ...interface{}) (err e
...
-// * source has enough balance to pay
-// * and it's `Operations`
+//
+// Validate the entirety of a transaction
+//
+// This function is critical for consensus, as it defines the validation rules for a transaction.
+// It is executed whenever a transaction is received.
+//
+// As it is run on every single transaction, it should be as fast as possible,
+// thus any new code should carefuly consider when it takes place.
```
Charleslee522: carefuly -> carefully

Issue #439, opened 2018-09-19
Hash for BlockOperation


kfangw:
https://github.com/bosnet/sebak/blob/ae5a58e519743cde45ea0cf4efcb972f3ff27347/lib/block/operation.go#L37-L55

Actually, Hash of BlockOperation is not ```Hash``` 
It is combined string by a hash of operation and a hash of the transaction
`return fmt.Sprintf("%s-%s", op.MakeHashString(), tx.GetHash())`
Currently, Operation Hash is used as a database key.

Like `Block` consist of hashes of the transactions,
`BlockTransaction` should consist of hashes of the operations

According to the current implementation, 
hashing transaction needs whole operation data but stored operations in the `BlockTransaction` is the hash of operation.
It means that the `BlockTransaction` is not self-provable.

kfangw:
How about adding the sequence number to `operation` to prevent hash collision?
If then, we can make the unique hash for every operation.

kfangw:
@anarcher @spikeekips @Geod24 I want to listen to your opinion

spikeekips:
@kfangw As I mentioned, `Transaction.Hash` looks like redundant, but it can be possible to check signature without calculating `Hash` in node-side.

kfangw:
@spikeekips It is the totally different issue what we discussed in person.

spikeekips:
@kfangw At this time, we can simply verify the BlockTransaction with original message, that is, checking Hash is matched, but it's not what we want.

kfangw:
fixed by #445

PR #445, opened 2018-09-20
Operation refactor to support diverse operation types

kfangw:
### Github Issue
#311 #439 

### Background
* BlockOperation Hash means TxHash + OpHash.
* Needs diverse Operations

### Solution
* Refactor: Use json.Unmarshaler interface
* BlockOperation store `body` instead of `amount, target`. Because these fields are not mandatory for all operations. Some operation has no amount or target
* Introduce OpHash for BlockOperation
* Introduce `OperationBodyPayable` interface for `OperationBodyCreateAccount` and `OperationBodyPayment`
* Add dummy OperationBody for congress voting


kfangw:
please review Commit by Commit.
@spikeekips @Geod24 @Charleslee522 

I will make another PR for Congress voting on top of this PR.
There might be a lot of concerns for Congress voting operations such as validation.

kfangw:
> Everything looks good 👍 One thing I suggest that instead of Body []byte in BlockOperation, how about using Body transaction.OperationBody. It will be automatically stored as serialized into storage and also without unmarshaling can be used.

I am planning to refactor packages like this
The `block` package does not depend on any other packages except `common` and `storage`.
IMO, the `block` package is a model(or schema) of database.
Hence, it should not import other packages. But other packages use it as a data entity.
If it is accepted, packages are more `loose coupling high cohesion`

But now, It is convenience if the BlockOperation knows `transaction.OperationBody`.
So I will patch it according to your comment
The refactoring will be done after several weeks later.

@spikeekips what do you think about this?

spikeekips:
@kfangw 👍

Charleslee522:
@kfangw Okay. I'll check test case in next pr. Please check integration test fail.

lib/node/runner/api_block.go
```diff
@@ -179,7 +179,9 @@ func UnmarshalNodeItemResponse(d []byte) (itemType NodeItemDataType, b interface
 		err = unmarshal(&t)
 		b = t
 	case NodeItemTransaction:
-		b, err = transaction.NewTransactionFromJSON(d[len(sc.Bytes())+1:])
+		var t transaction.Transaction
+		err = unmarshal(&t)
+		b = t
```
Geod24: That looks like a mistake

kfangw: It is intended.

Geod24: Oh I see what you did! Yeah that makes sense then! 👍

PR #446, opened 2018-09-20
Checking `BaseReserve` for new created account

### LINKED ISSUE #242: Checking Base reserve amount in validation process
Each account must be created with over base reserve amount, so `create-account` operation must be checked it has over given base reserve amount. For more details, https://www.stellar.org/developers/guides/concepts/fees.html will be helpful.

spikeekips:
### Github Issue

resolves #242 

### Background

Current tokennet has limit for new created account, it must be received up to base reserve, `0.1`BOS. If not, the new account will not be created.

### Solution

The `Amount` is set in `common.BaseReserve` like `common.BaseFee`. In the 2 parts, the base reserve will be checked,

1. `OperationCreateAccount.IsWellformed()`
1. `BlockAccount.Save()`


lib/common/constant.go
```diff
@@ -6,6 +6,10 @@ const (
 	// BaseFee is the default transaction fee, if fee is lower than BaseFee, the
 	// transaction will fail validation.
 	BaseFee Amount = 10000
+
+	// BaseReserve is minimum amount of balance for new account. By default, it
+	// is `0.1`BOS.
```
Geod24: Missing space

spikeekips:
> @soonkuk: I think checking base reserve is needed also before withdraw.

`BaseReserve` only applied when account is created. The balance of existing account can be below of `BaseReserve`.

spikeekips:
> @Geod24 : I don't understand why we need a BaseReserve then ?

`BaseReserve` is minimum amount of balance for **newly created account**. In BOScoin, new account must have at least `BaseReserve` amount of balance.

soonkuk:
> > `BaseReserve` only applied when account is created. The balance of existing account can be below of `BaseReserve`.
> 
> I don't understand why we need a `BaseReserve` then ?

I guess that that protect making account

spikeekips:
@soonkuk You closed accidentally :)

soonkuk:
> @soonkuk You closed accidentally :)

I guess the purpose of base reserve is to prevent making account unrestrictedly.

spikeekips:
@soonkuk `BaseReserve` is BOScoin policy :)

Geod24:
@spikeekips I understand the implementation, I'm asking for the rationale behind it.
I understand why Stellar has a minimum balance, but why do we do that only during account creation ?

spikeekips:
@Geod24 I understand why Stellar has a minimum balance, but why do we do that only during account creation?

Yes, our policy is different with stellar. Why? I think ours is reasonable rather than stellar.

Geod24:
> Yes, our policy is different with stellar. Why? I think ours is reasonable rather than stellar.

Can you explain why ?

spikeekips:
> @Geod24 : Can you explain why ?

Before I comment :), I'm not the original writer of our whitepaper. After creating account, the restriction  to keep the minimum balance seems to be useless. I have a balance and I can withdraw. Simple equation.

Geod24:
> Before I comment :), I'm not the original writer of our whitepaper. After creating account, the restriction to keep the minimum balance seems to be useless. I have a balance and I can withdraw. Simple equation.

Is it mentioned in the whitepaper ? I tried to grep for it in the WP 1.0 but didn't find anything.

I don't see this restriction as useless, as it would be easy for someone to make a DOS attack by just creating a lot of account with balance 1. It would cost a bit of fee, but it's not impossible.

Anyways I'm not opposed to this, just found it odd.

spikeekips:
@Geod24 not exactly in WP, but in current Tokennet. This is not the golden rule. :)

Geod24:
@spikeekips : Doesn't the current tokennet also have a minimum balance for account ?

spikeekips:
@Geod24 Current tokennet has same rule for stellar and we wanted to apply this rule, but as we know it’s stellar. Anyway our policy is that BaseReserve is for new creative ed account and it is ‘0.1 BOS’.

lib/block/account.go
```diff
@@ -52,7 +53,11 @@ func (b *BlockAccount) Save(st *storage.LevelDBBackend) (err error) {
 	if exists {
 		err = st.Set(key, b)
 	} else {
-		// TODO consider to use, [`Transaction`](https://godoc.org/github.com/syndtr/goleveldb/leveldb#DB.OpenTransaction)
+		if b.Balance < common.BaseReserve {
+			err = errors.ErrorInsufficientAmountNewAccount
+			return
+		}
```
Geod24: Could you move this in the consensus rules please ? The `Save` method shouldn't be the one doing this check.

Geod24:
> Anyway our policy is that BaseReserve is for new creative ed account and it is ‘0.1 BOS’.

Okay 👍

lib/block/account.go
kfangw: I have the same opinion as @Geod24.
This check is redundant and already checked in the `IsWellFormed` method.
`BlockAccount` shouldn't care about this kind of rules.

spikeekips: @kfangw Hmm, I want to make restricting of `BaseReserve` to be more strictly, but I will accept your opinion.

Geod24: @spikeekips : It's not about being strict, but about separation of concerns (SoC).
`Save`'s responsibility is to write the data out. It should only check for things that prevents it from doing its job, like read-only fs, `nil` key, etc...

spikeekips: @Geod24 What you mean 'Save's responsibility? If we have something to check in `BlockAccount`, we can do it in `Save()`.

Geod24: @spikeekips : No, if it gets to `Save` that means `BlockAccount` is already is an invalid state.
By responsibility I mean "role", "duty", etc... Or, in other words, what is the scope of this function ? Why don't we put all the consensus rules in `Save`, for example ?

spikeekips: @Geod24 Whether putting the some checking routine in `Save()` or not, it's just matter of choice, what is the explicit rule can be exist? What is the 'role' or 'duty'? Who made these things  for our `Save()`?

Geod24: @spikeekips : One simple rule is about side effect. By isolating side effects as much as possible, code is more robust and testable. And `Save` function is a nest of side effects (two write to the storage and triggers the observer).

PR #450, opened 2018-09-26
`--publish` new `node run` option, can set the publish endpoint

### LINKED ISSUE #400: Support to set public endpoint in `sebak node run --publish`
Currently `sebak node run` support to set network binding with `--endpoint` option. If node are running in internal subnet and it has different network endpoint for outside of the subnet.
```
-- subnet: 192.168.0.xxxx
node0(192.168.0.1) node1(192.168.0.2)

-- outside subnet: 49.171.139.xxx
node2(49.171.139.165)
```
`node0` can connect to `node2, but `node2` can not to `node0`.

The requirements are,
* rename the current `--endpoint` to `--bind`
  - if the `host` part is `localhost`, network will be bind to `localhost`
  - if the `host` part is not `localhost`, network will be bind globally
* create new option, `--publish`
  - this will be used for other nodes and clients
    - `--validators` of other nodes
    - node info
    - HAL resource response
  - If does not set `--publish` option, the `--bind` will be used instead

spikeekips:
### Github Issue

resolves #400 

### Background

The previous `--endpoint` option can set the bind address of network and also be the node's endpoint. This endpoint URL is published thru `/node/` to the other nodes. The problem is `--endpoint` is just used for binding, not real URL address to connect.

### Solution

The changes are simple.

* `--endpoint` option was removed
* `--bind` and `--publish` options were added
* `--bind` set the binding URL. The default is `defaultBindURL   string      = "https://0.0.0.0:12345"`
* `--publish` set the published URL address of node
* `--publish` is not set, `/node/` will use `http.Request` information; check out `common.RequestURLFromRequest()`

### Possible Drawbacks

Please update your `$ sebak node run` options :)


lib/node/runner/api_node.go
```diff
@@ -13,22 +14,35 @@ import (
 	"boscoin.io/sebak/lib/storage"
 )
 
+const (
+	NodeInfoHandlerPattern string = "/"
+	ConnectHandlerPattern  string = "/connect"
+	MessageHandlerPattern  string = "/message"
+	BallotHandlerPattern   string = "/ballot"
```
Charleslee522: 👍

PR #453, opened 2018-09-26
Genesis block should have valid transaction

spikeekips:
### Background

We can generate the genesis block by `block.MakeGenesisBlock`. Currently impl just create empty block and it uses the hash of genesis accoun and nothing. In ISAAC block created by Ballot, but genesis block can not have Ballot :)

### Solution

So, we can use Transction for genesis block to resolve this uncomfortable structural issue. Using Transaction in genesis block, we also solve another problem, that is, we can not find the genesis account from genesis block.  In this commit, I made the reason of genesis block :)

The genesis block has some noticeable things,

`MakeGenesisBlock` makes genesis block from genesis account and transaction.  This Transaction is different from other normal Transaction;

* must have only one `Operation`, `OperationCreateAccount`
* `Transaction.B.Source` is same with `OperationCreateAccount.Target`
* `Transaction.B.Fee` is 0
* `OperationCreateAccount.Amount` is same with balance of genesis account
* `OperationCreateAccount.Target` is genesis account

`block.TestMakeGenesisBlockFindGenesisAccount` can explain how to find genesis account.

### DIFF
--- a/lib/block/block.go
+++ b/lib/block/block.go
@@ -13,6 +13,7 @@ import (
 	"boscoin.io/sebak/lib/consensus/round"
 	"boscoin.io/sebak/lib/error"
 	"boscoin.io/sebak/lib/storage"
+	"boscoin.io/sebak/lib/transaction"
 )
 
 const (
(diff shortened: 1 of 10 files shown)


lib/block/block.go
```diff
@@ -41,26 +42,74 @@ func (bck Block) String() string {
...
+		Source:     account.Address,
+		Fee:        0,
+		SequenceID: account.SequenceID,
+		Operations: []transaction.Operation{op},
+	}
+
+	tx := transaction.Transaction{
+		T: "transaction",
+		H: transaction.TransactionHeader{
+			Created: common.GenesisBlockConfirmedTime,
```
Charleslee522: Why is `Created` time constant?

lib/block/block.go
```diff
@@ -41,26 +42,74 @@ func (bck Block) String() string {
...
 
-	b := NewBlock(
-		proposer,
-		round,
+	blk = NewBlock(
+		"",
+		round.Round{}, // empty round
 		transactions,
-		confirmed,
+		common.GenesisBlockConfirmedTime,
```
Charleslee522: Why is the block confirmed time constant?

lib/block/block.go
spikeekips: @Charleslee522 Every node will have different `Transaction.Hash` for genesis block, so I set it to `constant`.

lib/block/block.go
spikeekips: @Charleslee522 Genesis block will be created in every node separately, so we need to set it to be same.

lib/block/block.go
Charleslee522: Then is this field required? I think it is useless if it is constant.

spikeekips: @Charleslee522 This `Transaction` also will be stored in `BlockTransaction` so we need to keep it same between nodes.

Charleslee522: @spikeekips Then I think it is better to be set by conf like network-id.
It is weird that the first tx and block time is too old.

Geod24: [Nothing wrong with hardcoding the time, or the whole block](https://github.com/bitcoin/bitcoin/blob/4dac24db23d121ec21080f04e210227c4ff5d281/src/chainparams.cpp#L41-L57)

lib/block/operation.go
```diff
@@ -138,6 +138,10 @@ func (bo BlockOperation) NewBlockOperationSourceKey() string {
 	)
 }
 
+func (bo BlockOperation) LoadBody() (body transaction.OperationBody, err error) {
```
spikeekips: @kfangw Please review this code. With using `BlockOperation.Body`, I could not find the way to access to original `OperationBody`, so I wrote new function, `transaction.UnmarshalOperationBodyJSON`(below), which is load `OperationBody` from `[]byte`.

kfangw: @spikeekips Looks good.
But please mind #456 in advance.
After boscon, let's do refactoring like that

spikeekips: @kfangw Oops, I missed. I’ll ask your advice tomorrow.

PR #454, opened 2018-09-26
Set `BlockTransaction.Confirmed` by `Block.Confirmed`

spikeekips:
### Background

The previous `block.BlockTransaction.Confirmed` set to the saving time, so it will be different between nodes

### Solution
So it is set by`block.Block.Confirmed`, instead of saving time.

### DIFF
--- a/lib/block/transaction.go
+++ b/lib/block/transaction.go
@@ -42,7 +42,7 @@ type BlockTransaction struct {
 	blockHeight uint64
 }
 
-func NewBlockTransactionFromTransaction(blockHash string, blockHeight uint64, tx transaction.Transaction, message []byte) BlockTransaction {
+func NewBlockTransactionFromTransaction(blockHash string, blockHeight uint64, confirmed string, tx transaction.Transaction, message []byte) BlockTransaction {
 	var opHashes []string
 	for _, op := range tx.B.Operations {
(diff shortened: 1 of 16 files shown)


lib/block/transaction.go
```diff
@@ -79,10 +80,8 @@ func (bt BlockTransaction) NewBlockTransactionKeySource() string {
 
 func (bt BlockTransaction) NewBlockTransactionKeyConfirmed() string {
 	return fmt.Sprintf(
-		"%s%s%s%s",
+		"%s%s",
 		GetBlockTransactionKeyPrefixConfirmed(bt.Confirmed),
-		common.EncodeUint64ToByteSlice(bt.blockHeight),
```
kfangw: Do you think the confirmed time is enough to order block transactions? @spikeekips

spikeekips: @kfangw Yes, I think so. The following `common.GetUniqueIDFromUUID()` supports to make sequential unique string in same machine, so `NewBlockTransactionKeyConfirmed()` will be sorted by creation.

lib/network/api/transaction_test.go
```diff
@@ -7,13 +7,14 @@ import (
 	"io/ioutil"
 	"testing"
```
Charleslee522: How about removing this line?

PR #457, opened 2018-09-27
Operation CongressVoting

### LINKED ISSUE #311: Congress Voting: Messages and API

kfangw:
### Github Issue
Resolves #311 

### Background
To store proposal/result of congress voting to blockchain, new operations needed

### Solution
New 2 operations

### Possible Drawbacks
Operations are just stored in the blockchain with its transaction but no affect to account.

### DIFF
--- a/lib/transaction/operation.go
+++ b/lib/transaction/operation.go
@@ -14,8 +14,8 @@ type OperationType string
 const (
 	OperationCreateAccount        OperationType = "create-account"
 	OperationPayment                            = "payment"
-	OperationCongressVoting                     = "cv"
-	OperationCongressVotingResult               = "cvr"
+	OperationCongressVoting                     = "congress-voting"
+	OperationCongressVotingResult               = "congress-voting-result"
 )
 
 type Operation struct {
--- a/lib/transaction/operation.go
+++ b/lib/transaction/operation.go
@@ -112,13 +112,13 @@ func UnmarshalOperationBodyJSON(t OperationType, b []byte) (body OperationBody, 
 		body = ob
 	case OperationCongressVoting:
 		var ob OperationBodyCongressVoting
-		if err = json.Unmarshal(envelop, &body); err != nil {
+		if err = json.Unmarshal(b, &ob); err != nil {
 			return
 		}
 		body = ob
 	case OperationCongressVotingResult:
 		var ob OperationBodyCongressVotingResult
-		if err = json.Unmarshal(envelop, &body); err != nil {
+		if err = json.Unmarshal(b, &ob); err != nil {
 			return
 		}
 		body = ob
--- a/lib/transaction/operation.go
+++ b/lib/transaction/operation.go
@@ -12,8 +12,10 @@ import (
 type OperationType string
 
 const (
-	OperationCreateAccount OperationType = "create-account"
-	OperationPayment                     = "payment"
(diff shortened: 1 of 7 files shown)


kfangw:
To make an integration test case, it is needed to support the client module for API. #471 
This may be revised when I found a problem during the test by the client module.

lib/transaction/operation.go
```diff
@@ -12,8 +12,10 @@ import (
 type OperationType string
 
 const (
-	OperationCreateAccount OperationType = "create-account"
-	OperationPayment                     = "payment"
+	OperationCreateAccount        OperationType = "create-account"
+	OperationPayment                            = "payment"
+	OperationCongressVoting                     = "cv"
+	OperationCongressVotingResult               = "cvr"
```
spikeekips: @kfangw How about more expressive name like `congress-voting`, `congress-voting-result`? They needs more spaces, but expressive :)

lib/transaction/operation_congress_voting.go
```diff
@@ -0,0 +1,33 @@
+package transaction
+
+import (
+	"boscoin.io/sebak/lib/error"
+	"encoding/json"
+	"time"
+)
+
+type OperationBodyCongressVoting struct {
+	Contract string
```
spikeekips: @kfangw Just suggestion, `Contract []byte`, how about this?

lib/transaction/operation_congress_voting.go
```diff
@@ -0,0 +1,33 @@
...
+	Voting   struct {
+		Start string
+		End   string
```
spikeekips: @kfangw Starting and ending should be `Block.Height`. :)

lib/transaction/operation_congress_voting_result.go
```diff
@@ -0,0 +1,70 @@
...
+
+	var tobeHash = ""
+	for _, u := range o.BallotStamps.Urls {
+		if _, err := url.Parse(u); err != nil {
+			return errors.ErrorInvalidOperation
+		}
+		tobeHash += u
+	}
+
+	hash := common.MakeHash([]byte(tobeHash))
```
spikeekips: @kfangw When I introduced the first thought of voting result things, the `Hash` is the file checksum of voting result, the file is the downloaded from `Urls`.

lib/transaction/operation_congress_voting_result.go
```diff
@@ -0,0 +1,70 @@
...
+	tobeHash = ""
+	for _, u := range o.Voters.Urls {
+		if _, err := url.Parse(u); err != nil {
+			return errors.ErrorInvalidOperation
+		}
+		tobeHash += u
+	}
+
+	hash = common.MakeHash([]byte(tobeHash))
+	if string(hash) != o.Voters.Hash {
```
spikeekips: @kfangw This is also same with upper comment.

lib/transaction/operation_congress_voting_result.go
kfangw: Then, the `Hash` is not verifiable, right?
@spikeekips

spikeekips: @kfangw Yeah, I thought so. I can be verifiable by manual.

PR #462, opened 2018-09-28
utils for running nodes locally

kfangw:
run_build_image.sh - build docker image and remove dangling one(builder)
run_start_containers.sh - start containers by using the image
run_stop_containers.sh - stop and rm the containers

Dockerfile revised
- To improve the performance of building an image, Copy source code should be located bottom as possible. And apk add should be located on top. It may increase CI spead


kfangw:
@soonkuk It may help for your testing on a local machine.

spikeekips:
@kfangw It will be better the instruction doc to use these scripts.

Dockerfile
```diff
@@ -2,33 +2,31 @@
...
-COPY ./ /go/src/boscoin.io/sebak
-WORKDIR /go/src/boscoin.io/sebak
-
 # You probably don't need to change this
 ARG BUILD_MODE="install"
 ARG BUILD_ARGS=''
 ARG BUILD_PKG="./..."
 
-RUN apk add --no-cache git openssh gcc musl-dev linux-headers
+RUN apk add --no-cache git openssh gcc musl-dev linux-headers ca-certificates
```
Geod24: Why that change ?

run_build_image.sh
```diff
@@ -0,0 +1,16 @@
...
+IMAGE_TAG="sebak:test"
+
+docker build -q -t ${IMAGE_TAG} \
+    --build-arg BUILD_MODE='test' \
+    --build-arg BUILD_PKG='./cmd/sebak' \
+    --build-arg BUILD_ARGS='-tags integration -c -o /go/bin/sebak' \
+    ${ROOT_DIR}
+
+docker rmi $(docker images --filter dangling=true -q --no-trunc) || true
+docker images
```
Geod24: That is highly workflow dependent and should not be here IMO

Dockerfile
kfangw: As described in the above, it is related to building a docker image.
> To improve the performance of building an image, Copy source code should be located bottom as possible. And apk add should be located on top. It may increase CI spead

run_build_image.sh
kfangw: The reason why I added this statement is that
the Dockerfile is based on two `From`. Hence, `builder` image remains as unnamed dangling.
But, I understand what you saying.

kfangw:
@Geod24 @spikeekips thanks for your comments.
What I intended for this PR is making a util for easy testing on a local machine.
As @Geod24  said, I know that this is not needed to be on this repo.
I will take more time to consider to support what I want to.

Dockerfile
Geod24: Yes the `COPY/WORKDIR` move makes sense, but my comment is on adding `ca-certificates` here.
The first image is used for building, the second for running.
`ca-certificates` is used for running sebak, not for building it.

spikeekips:
@kfangw This PR can be closed.

kfangw:
please see #486

PR #465, opened 2018-09-28
Remove unused abstraction

Geod24:
This DBBackend abstraction is completely unused, as shown by the diff.
This probably comes from the fact that LevelDB already provides
a good enough abstraction storage-wise: It's a storage that can
go either to disk or memory, and we don't need more (storage wise).


kfangw:
It is used by `contract` branch.
But for the master branch, it is useless.

Geod24:
@kfangw : Anything preventing you from switching to `LevelDBBackend` ?

kfangw:
> @kfangw : Anything preventing you from switching to `LevelDBBackend` ?

I and @anarcher had been tried to introduce stateDB to SEBAK.
There is still unused code at `statedb` package.
And with the `statedb` package, at the `contract` branch, there was a reason why use the interface.
But I am not clear now, it was weeks ago.

As I mentioned above, It is useless for the master branch.
If there is a need, then the code will be reverted.
You can go as-is

PR #448, opened 2018-09-25
WIP: Sync

anarcher:
### Github Issue

Related: #302

### Background

The first attempt (It's prototyping) for Sync is that just fetch block and all related data of it and then save (validation) it. 
This attempt does start from the genesis block. 

#### Key interfaces

```
+--------------+                +--------------+
|              |     Produce    |              |
|   Producer   +--------------->+    Consumer  |
|              |                |              |
+-----^--------+                +-------+------+
      |                                 |
      |                                 |
      |             Response            |
      +---------------------------------+
```

```
type Stopper interface {
    Stop() error
}

type Producer interface {
    Produce() <-chan *Message
    SetResponse(<-chan *Response) error
}

type Consumer interface {
    Consume(<-chan *Message) error
    Response() <-chan *Response
}

type Processor interface {
    Producer
    Consumer
}
```
- `Processor` is both a producer and a consumer. 
- `Processors` and `Consumers` are a pool of `Processor`s or `Consumer`s  and the `Processors` and `Consumers` dispatches the `*Message`  to one of them. 

----

- Basic pipelining looks like below: 
```
err := Pipeline(producer,consumer)
```

----

```
type Message struct {
    BlockHeight uint64
    Block       *block.Block
    Txs         []*block.BlockTransaction
    Ops         []*block.BlockOperation
}
```
- `*Message` has  a data  for processing. The `*Message` can be more data and metadata such as `FailedNodes`  and `RetryFailedCnt`.

---- 

```
type Response struct {
    err error
    msg *Message
}

func (r Response) Err() error {
    return r.err
}
func (r Response) Message() *Message {
    return r.msg
}
```
- The `*Response` has poor information for now.  Each component can handle the response by `*Response.Err()` type. 
- A more conditional error I imagine is inspired by `upspin/error` system. (Each error is just a struct.)
- For example, `FetchError` has retryNum and Upstream component can handle the response by the response.Err()
- But for now the sync pkg has no  custom errors [TODO]  

### Possible Drawbacks

- This design can be overdesign. `ㅠㅠ`

Issue #302, opened 2018-08-31
Store confirmed block without validation in `SYNC` state


Charleslee522:
Unless the node is fully synced yet, it should store confirmed block without validation temporarily.
1. Receive only `ACCEPT` ballot and do not validate.
1. If the number of ballots exceeds threshold, store the block.
1. The block is handled exactly same with received block from validator.
   * It means, these blocks are validated sequentially from genesis Block.
1. When it is done, the node state transits to `CONSENSUS`

Geod24:
> Unless the node is fully synced yet, it should store confirmed block without validation temporarily.

Why does it needs to bypass validation ?

Geod24:
I see, it makes perfect thanks, thanks for the clarification!

PR #467, opened 2018-09-30
Sync 

anarcher:
### Github Issue

Related: #302

### Background

IMHO, This is a much more`* N` simple version of sync than #448  `~_~`

### Solution

- Just use `WorkerPool` for syncing. (default pool size is 300)

```
func (s *Syncer) work(height uint64) {
	ctx := s.ctx
	work := func() {
	L:
		for {
			select {
			case <-ctx.Done():
				break L
			default:
				currHeight := s.currBlockHeight()
				if currHeight > 0 && height <= currHeight {
					s.logger.Info("This height has already synced", "height", height)
					break L
				}
				blockInfo, _ := s.fetcher.Fetch(ctx, height)
				if err := s.validator.Validate(ctx, blockInfo); err != nil {
					s.logger.Error("validate failure", "err", err)
					continue
				}
				break L
			}
		}
		s.logger.Info("Done sync", "height", height)
	}
	s.workPool.Add(ctx, work)
}
```

- If Fetcher and Validator need to do work concurrently, They can have their own `WorkerPool` too. `'ㅅ'`

### DIFF
--- a/cmd/sebak/cmd/run.go
+++ b/cmd/sebak/cmd/run.go
@@ -402,7 +402,7 @@ func runNode() error {
 		})
 	}
 	{
-		c := sync.NewConfig(localNode, st, nt, connectionManager)
+		c := sync.NewConfig([]byte(flagNetworkID), localNode, st, nt, connectionManager)
 		//Place setting config
 		syncer := c.NewSyncer()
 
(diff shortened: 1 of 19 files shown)


spikeekips:
❤️

Charleslee522:
@anarcher Is this in progress yet? :)

anarcher:
Not yet ㅠㅠ. I think that it’s time to talk how do interacting with consensus module with syncer :)

anarcher:
```
type SyncProgress struct {
    HighestBlock uint64 // Highest alleged block number in the chain
    CurrentBlock uint64 // Current block number where sync is at
    LastestBlock uint64 // Lastest block number which is saved in the node.
}

type SyncProgressHandler interface {
    SyncProgress(context.Context) (*SyncProgress, error)
    SetSyncTargetBlock(ctx context.Context, height uint64) error
}
```

Issue #461, opened 2018-09-28
Block inflation


soonkuk:
An inflaction is generated for every block like as the coinbase.

soonkuk:
This issue references #308 (Implement regular coin issuance)

PR #469, opened 2018-10-01
Transaction Fee Will Go To Common (Budget) Account

### LINKED ISSUE #63: Transaction Fee should be collected to 'Common Budget Account'
In SEBAK, there are 2 special accounts,
- genesis account(most of blockchains does)
- common budget account

'Common budget account' of our BOScoin is unique feature, it's balance can be used for entire BOScoin network and our community, it's like [common goods](https://en.wikipedia.org/wiki/Common_good_(economics)).

When transaction is finished, it's fee will be collected to 'Common budget' account.

spikeekips:
# Github Issues

resolves #63

# Background

Please see the #63 .

# Solution

There are severl new feature for this issue.

* new special account, 'Common (Budget) account'
* new kind of `Opertion`, to represent the 'collecting transaction fee'

### Common Account

The sum of transaction fee in ballot must be be sent to the common account, so the common account must be created at the genesis block with the genesis account. The `$ sebak genesis ...` command is changed.

```
$ go run cmd/sebak/main.go genesis -h
initialize new network

Usage: sebak genesis <public key of genesis account> <public key of common account> [flags]

Flags:
      --balance string      initial balance of genesis block (default "1,000,000,000,000.0000000")
  -h, --help                help for genesis
      --network-id string   network id
      --storage string      storage uri
```

* The transaction for genesis block has 2 'OperationCreateAccount' operations, for genesis account and common account.
* The common account will be created by same way of genesis account. To check the common account, `runner.GetCommonAccount()` will return the common account.
* `NodeRunner` initially checks the existence of common account using `runner.GetCommonAccount()`, if failed, `NewNodeRunner()` returns error.
* By default, the balance of common account is `0`.
* The details of genesis block is described in `block.MakeGenesisBlock()`.
* The transaction for making the genesis block, is signed by `keypair.Master(string(networkID))`.

### Collected Transaction Fee To Common Account

For sending the collected, the basic process is simple;

1. Proposer makes ballot with the special transaction(name is `ballot.ProposerTransaction`).
1. Each node validates `ProposerTransaction` with other transactions.
1. After all the consensus process is finished, each node will process `runner.finishBallot()`.
1. In the last part of `runner.finishBallot()`, `runner.finishProposerTransaction()` is processed. It handles `ProposerTransaction`.

For sending collected fee, `ballot.ProposerTransaction` has new operation, `transaction.OperationBodyCollectTxFee`.

* Without `ProposerTransaction`, the ballot will be failed to be validated.
* As you expected, `ProposerTransaction` can contain the multiple `Operation`s, so this will be helpful to support upcoming block inflation feature.      

### Calculating Transaction Fee

There are some misuse to calculate fee of one transaction. Originally I intened that `Transaction.B.Fee` contains the sum of fee of `Transaction.B.Operations`. In this PR, I fixed it.

### Etc

* `transaction.NewOperation` was added.
* `Transaction.TotalBaseFee()` was added.
* `SelfSelector` renamed to `FixedSelector`.

### `ProposerTransaction` in `/node/blocks`

* `ProposerTransaction` is also transaction, so it must be stored in `block.Block.Transaction`.
* `/node/blocks` api will return the `ProposerTransaction`.


lib/block/block.go
```diff
@@ -23,6 +24,7 @@ const (
 type Block struct {
 	Header
 	Transactions []string `json:"transactions"` /* []Transaction.GetHash() */
+	Transaction  string   `json:"transaction"`  /* ProposerTransaction */
```
Charleslee522: I figure, `Transaction` is too general word.
And `ProposerTransaction` is not well fit for this purpose I think.
How about `SpecialTransaction` or `ExceptionalTransaction`?

soonkuk:
Is there checking process which each node verify total fee amount that proposer suggested?
I can't find.

kfangw:
The `Transaction` already has `T` field. According to this type variable, 
is it impossible to make `ProposerTransaction` as just `Transaction` with type `ProposerTransaction`?

@spikeekips

spikeekips:
@soonkuk Please check,

* https://github.com/bosnet/sebak/pull/469/files#diff-a02cc61ff31258020f9b40e5aff2448eR65
* https://github.com/bosnet/sebak/pull/469/files#diff-a02cc61ff31258020f9b40e5aff2448eR82
* https://github.com/bosnet/sebak/pull/469/files#diff-3ac5aa21aabcbf6c057cb872f5980636R148

kfangw:
> @kfangw Yes, it could be possible to set the different `T`, but I don't want to make this `ProposerTransaction` really different from `Transaction`. The current shape of `ProposerTransaction(.Transaction)` is easy way to pass thru `TransactionChecker`. I want we have only one `Transaction` which has the `T=transaction`.

Maybe my mind is stuck to Bitcoin's coinbase. That is why I asked like above.
So, I have two more question. If you want to make T has only one value, then, it is not neccessary to keep the field. And is it possible to store `block` with only `transactions`? not add one more `transaction`?

spikeekips:
> @kfangw : If you want to make T has only one value, then, it is not neccessary to keep the field.

Frankly to say, `Transaction.T`, we call it legacy when all the incoming message including `Ballot` goes thru one channel. I heard you, yes, we can remove it because we have the separated channel for `Transaction`. 

> And is it possible to store `block` with only `transactions`? not add one more `transaction`?

I think `BlockTransaction` should store `Transaction`, I can not imagine the other kind of transaction for `BlockTransaction` at this time.

lib/block/block.go
spikeekips: @Charleslee522 'Special' and 'Exceptional' looks also too general :) I had 2 candidates,

* the current one, 'ProposerTransaction'
* and dismissed one, 'BlockTransaction'

'BlockTransaction' is already used for `block.BlockTransaction`, so ...  This transaction is proposed by proposer and it will contain the other kind of Operation, like for #461 .

Please suggest me the other wonderful name for it.

kfangw: > > And is it possible to store block with only transactions? not add one more transaction?
I think BlockTransaction should store Transaction, I can not imagine the other kind of transaction for BlockTransaction at this time.

I mean this code.
Why not store the `Transaction` onto the first position of `Transactions`?
@spikeekips

spikeekips: @kfangw You pointed very sharply. The first design of this feature was that. Instead of adding new field for `ProposerTransaction`, set it at the first item in `Ballot.B.Transactions`. But I dropped this idea because of broadcasting problem. I want `ProposerTransaction` should be validated without broadcasting. Without broadcasting, it should be some where in `Ballot`.

kfangw: Ok. understood. thanks for your kind explanation.
@spikeekips

Charleslee522: @spikeekips 
> I figure, Transaction is too general word.

It means I think 
```
type Block struct {
	Header
	Transactions []string `json:"transactions"` /* []Transaction.GetHash() */
	ProposerTransaction  string   `json:"transaction"`
}
```
or
```
type Block struct {
	Header
	Txs []string `json:"transactions"` /* []Transaction.GetHash() */
	ProposerTx  string   `json:"transaction"`
}
```
is better than
```
type Block struct {
	Header
	Transactions []string `json:"transactions"` /* []Transaction.GetHash() */
	Transaction  string   `json:"transaction"`  /* ProposerTransaction */
}
```

Charleslee522: @spikeekips And normal transactions are also proposed by client.
So I thought that is general.
What the `ProposerTransaction` means is "Transaction by Proposer", right?

spikeekips: Right.

Charleslee522: @spikeekips What do you think about this? https://github.com/bosnet/sebak/pull/469/files#r221781065

spikeekips:
@Charleslee522 Please continue to review.

Charleslee522:
@spikeekips please check this :)
https://github.com/bosnet/sebak/pull/469#discussion_r221801259

PR #431, opened 2018-09-18
Fix argon2 salt length over 16 bytes

outsideris:
### Github Issue
N/A

### Background
While implementing hashing in JavaScript, I found the salt length for Argon2 is only 5 bytes. 
However, [JavaScript Argon2 library](https://github.com/ranisalt/node-argon2) thrown an error `Error: Salt is too short` when I use `sebak` as the salt. 
[As IETF argon2 documents](https://tools.ietf.org/id/draft-josefsson-argon2-00.html#rfc.section.3.1) said:

> Nonce S, typically a random salt. May have any length from 8 to 2^32 - 1 bytes. 16 bytes is recommended for password hashing. 

### Solution
Increase the salt length over 16 bytes.

### Possible Drawbacks
I don't know. Performance maybe?

PR #470, opened 2018-10-01
Change hash algorithm from argon2 to doubl sha256

outsideris:
### Github Issue
Replace #431 

### Background
Argon2 use too many CPU resources. As @zzim2x 's testing, it took about 5 secs for hashing in browser side. 

### Solution
Use double sha256 hash like bitcoin.

argon 2 benchmark:
```
2000000000           0.03 ns/op
PASS
```

double sha256 banchmakr:
```
2000000000           0.00 ns/op
PASS
```


outsideris:
I gave up to write integration test after I tried several times.
Please save me. @Geod24 @spikeekips :)

Charleslee522:
Please check #399

outsideris:
@Charleslee522 I didn't understand what I should check in #399 
This PR changed hash, so the signature should be updated.

Charleslee522:
@outsideris Oh, does integration_tests fails because of signature?
I think it because of performance.

spikeekips:
Argon2 makes also problem in client side :(

spikeekips:
Anyway, I can make proper tx message with new hashing algo.

spikeekips:
@outsideris I generate the overdraft/*.json with your code. Check https://gist.github.com/spikeekips/0002b08c6f83ea80c300eb4e5125018a .

spikeekips:
@outsideris I used this tool to sign again. https://github.com/spikeekips/sebak-tools/tree/master/sebak-tx-sign

outsideris:
Finally, I passed CI!!

docker/entrypoint.sh
```diff
@@ -12,5 +12,5 @@ if [ $# -gt 0 ]; then
     exec ./sebak $@
 else
     # Node mode
-    exec ./sebak node --genesis=${SEBAK_GENESIS_BLOCK}
+    exec ./sebak node --genesis=${SEBAK_GENESIS_BLOCK},${SEBAK_COMMON_ACCOUNT}
```
Geod24: What does the common account has to do with this ?

tests/overdraft/request_1_2821_create_account_1.json
```diff
@@ -3,12 +3,12 @@
...
     "version": "",
     "created": "2018-09-27T16:26:22.830373928+09:00",
-    "signature": "5TVx3JQHgZhHGcNLAyLfJQfssu4Hh7QRWBfjymEd7i1yC2s3yjSjKs92nwdkupKjFsQdbUNZHdkSQhicrT8ZSVWz"
+    "signature": "44unZEoQy41vYQr8LCFwJfXv8FpM6waxddwDd4sAs7Koo1XiJy18Lvp2FH6JG18DbvMwJTs5VQnoVoNEyJdZqicM"
   },
   "B": {
     "source": "GDIRF4UWPACXPPI4GW7CMTACTCNDIKJEHZK44RITZB4TD3YUM6CCVNGJ",
     "fee": "10000",
-    "sequenceID": 0,
+    "sequenceid": 0,
```
Geod24: Why is this changed ?

outsideris: I think it caused when I print the body and signature with [sebak-tx-sign](https://github.com/spikeekips/sebak-tools/tree/master/sebak-tx-sign). Indeed, there was a reason.
BTW, it does not affect the signature.

Geod24: And sebak deserialize it okay, I suppose ? Anyway, since it has nothing to do with changing that algorithm, it should not have been changed here.

outsideris: > And sebak deserialize it okay, I suppose ?

I think so. Yes, it was my mistake. I was too happier when integration tests are passed. If it causes any problems, it can be reverted.

PR #476, opened 2018-10-02
 Waiting until prev block save,  Remove localnode when pick fetch node

anarcher:
### Background

- Waiting validator in sync until prev block is saved.
- Remove localNode when picking fetching node. 
- Validate block with prevblock's hash 

### Solution

- Using `common/observer` and `context.WithTimeout`
```
        ctx, cancelFunc := context.WithTimeout(pctx, v.prevBlockWaitTimeout)
	defer cancelFunc()

	prevHeight := height - 1

	exists, err := block.ExistsBlockByHeight(v.storage, prevHeight)
	if err != nil {
		return nil, err
	}

	if exists == false {
		waitC := make(chan struct{})
		observer.SyncBlockWaitObserver.On(string(prevHeight), func() {
			waitC <- struct{}{}
		})

		select {
		case <-waitC:
		case <-ctx.Done():
		}
	}
```

- Using `localNode.Address()` in `pickRandomNode`
- Make and hashing a block with local prev block and then compare this block's hash with fetched block's hash


lib/sync/fetcher.go
```diff
@@ -86,18 +96,12 @@ func (f *BlockFetcher) fetch(ctx context.Context, si *SyncInfo) error {
 	f.logger.Debug("Fetch start", "height", height)
 
 	n := f.pickRandomNode()
-	f.logger.Info("Fetching items from node", "node", n, "height", height)
+	f.logger.Info(fmt.Sprintf("Fetching items from node: %s", n), "fetching_node", n, "height", height)
```
Charleslee522: Isn't `n` duplicated?

lib/sync/fetcher.go
```diff
@@ -175,8 +179,16 @@ func (f *BlockFetcher) pickRandomNode() node.Node {
 	if len(ac) <= 0 {
 		return nil
 	}
-	idx := rand.Intn(len(ac))
-	node := f.connectionManager.GetNode(ac[idx])
+
+	var addressList []string
+	for _, a := range ac {
+		if f.localNode.Address() != a {
```
Charleslee522: :)

lib/sync/validator.go
```diff
@@ -154,6 +160,15 @@ func (v *BlockValidator) validateBlock(ctx context.Context, si *SyncInfo) error
...
 		txs = append(txs, tx.H.Hash)
 	}
+
+	prevBlk, err := v.getPrevBlock(ctx, si.BlockHeight)
+	if err != nil {
+		return err
+	}
+
+	round := si.Block.Round
+	round.BlockHash = prevBlk.Hash
```
Charleslee522: How about this?
`si.Block.Round.BlockHash = prevBlk.Hash`
First I saw `round` variable, it looked like unused :)

lib/sync/validator.go
```diff
@@ -191,3 +206,34 @@ func (v *BlockValidator) existsBlock(ctx context.Context, st *storage.LevelDBBac
 		return exists, nil
 	}
 }
+
+func (v *BlockValidator) getPrevBlock(pctx context.Context, height uint64) (*block.Block, error) {
```
Charleslee522: 👍

lib/sync/fetcher.go
anarcher: Yep, but msg has node is easy to see log :-)

anarcher:
@Charleslee522  Can I merge it? ㅠㅠ

anarcher:
Thanks! @Charleslee522 :-D

PR #478, opened 2018-10-02
Fix issue 395: Drop support for Go v1.10

Geod24:
Just leaving this here, for when the team think it's ready.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,14 +43,6 @@ jobs:
           name: check formatting
           command: if [[ -n "$(gofmt -l cmd lib)" ]]; then gofmt -l cmd lib; exit 1; fi
 
-  test_go1_10:
-    <<: *defaults
-    steps:
-      - <<: *workspace
-      - run:
-          name: run tests
-          command: SEBAK_LOG_HANDLER=null go test -v -timeout 5m ./...
-
   test_go1_11:
     docker:
       - image: circleci/golang:1.11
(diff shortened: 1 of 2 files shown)


anarcher:
If you want to use go module, you can use `vgo` with go 1.10 (maybe?). 
- https://github.com/golang/vgo

Geod24:
@outsideris : Could you disable "Required" on `ci/circleci: test_go1_10` then ?

anarcher:
IMHO, I would like to wait to `go.1.11.x` . `go1.11` probably has some issues.  ㅋㅋ

outsideris:
I disabled go1.10 required

Geod24:
> IMHO, I would like to wait to `go.1.11.x` . `go1.11` probably has some issues. ㅋㅋ

[Here you go](https://groups.google.com/forum/#!topic/golang-announce/pFXKAfoVJqw)

Geod24:
Actually, it looks like we haven't been running the unittests with v1.10 for a while, since it was using `defaults`, and `defaults` was updated to Go v1.11 🙈

Geod24:
@spikeekips : I don't see anything in the README that needs an update.

spikeekips:
@Geod24 I was confused, I want to say, installation doc in wiki, https://github.com/bosnet/sebak/wiki/SEBAK-Installation-Guide . There are `dep` related things.

Geod24:
@spikeekips : Shouldn't this happen after this PR is merged then ?

spikeekips:
@Geod24 👍

Geod24:
I'm going to merge this now, without the switch to Go modules, because it will already help our CI, which is overloaded. I'll submit another PR for the switch to Go modules.

PR #480, opened 2018-10-02
Add Syncer.SetSyncTargetHeight

anarcher:
### Background
- `Syncer.SetSyncTargetHeight` for setting sync target height. 
-  Consensus handles which height It need to sync talks to syncer by this function. 
- If syncer's pool is not enough to add more works , syncer just keeps this height to `SyncProgress` and then after next check time (0.5sec by default) the syncer try to add more work from current working height  to this height.

### Solution

- Add `pool.TryAdd` for unblocking add 
- Add `SyncProgress` for keeping sync progress

```
type SyncProgress struct {
    StartingBlock uint64 // Block number where sync began
    CurrentBlock  uint64 // Current block number where sync is at
    HighestBlock  uint64 // Highest alleged block number in the chain
}
```

### Possible Drawbacks

- Current, Syncer try to increase  highest height per 0.5 sec, But if Consensus handles it, This increase can be removed. :-) (Because Consensus can handle better than timer based increasing.)

### DIFF
--- a/lib/sync/syncer.go
+++ b/lib/sync/syncer.go
@@ -184,12 +184,13 @@ func (s *Syncer) sync(p *SyncProgress) {
 		}
 	)
 
-	if latestBlockHeight > p.CurrentBlock {
-		currentHeight = latestBlockHeight
+	if latestBlockHeight > currentHeight {
+		startHeight = latestBlockHeight + 1
 	}
-	if currentHeight >= highestHeight {
+	if startHeight > highestHeight {
+		p.StartingBlock = latestBlockHeight + 1
 		p.CurrentBlock = currentHeight
-		log("sync progress skip: current latest height is over than highest (requested) height")
+		log("sync progress skip: start height is over or equal than highest (requested) height")
 		return
 	}
 
--- a/lib/sync/syncer.go
+++ b/lib/sync/syncer.go
@@ -33,6 +33,8 @@ type Syncer struct {
 	ctx        context.Context
 	cancelFunc context.CancelFunc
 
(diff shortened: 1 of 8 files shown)


anarcher:
I need to time to add test for it. `~_~`

lib/sync/syncer.go
```diff
@@ -161,15 +228,15 @@ func (s *Syncer) work(height uint64) {
...
-			s.logger.Info("Done sync work", "height", height)
+			s.logger.Info("done sync work", "height", height)
 		}
 	}
-	s.workPool.Add(ctx, work)
+	return s.workPool.TryAdd(ctx, work)
 }
 
-func (s *Syncer) currBlockHeight() uint64 {
+func (s *Syncer) lastestBlockHeight() uint64 {
```
Charleslee522: How about `lastest` -> `latest` as in consensus? :)

lib/sync/syncer.go
```diff
@@ -91,52 +97,113 @@ func (s *Syncer) Stop() error {
...
 }
 
-func (s *Syncer) work(height uint64) {
+func (s *Syncer) work(height uint64) bool {
 	ctx := s.ctx
 	work := func() {
-		currHeight := s.currBlockHeight()
-		if currHeight > 0 && height <= currHeight {
-			s.logger.Info("This height has already synced", "height", height)
+		lastHeight := s.lastestBlockHeight()
```
Charleslee522: How about `latestHeight := s.latestBlockHeight()`?
`latestBlock` is used in consensus :)

PR #485, opened 2018-10-04
Switch from using dep to using vgo / go modules

Geod24:
Originally part of https://github.com/bosnet/sebak/pull/478 but this PR had other advantages (speeding up the CI, for example) which is why I split it.
I'll ask for reviews when the CI is green.


Geod24:
Weird, it was not working in the previous PR but after I split the change and rebased everything, all looks fine. I guess I must have gotten the detail wrong on the first pass.

@outsideris : You asked if using go modules is recommended. Why yes, in fact go modules don't work inside of `GOPATH` by default!

anarcher:
https://gist.github.com/davecheney/311169ef04e27fe469ca91859af9cdff

Issue #471, opened 2018-10-01
API client for testing and SDK


kfangw:
To make API test easy and support SDK in advance,
API client module needed.

There is an `HTTP2NetworkClient` already in `network` package,
but this client is only for communication for nodes

PR #486, opened 2018-10-04
Integration Testing Env for SDK.

kfangw:
### Github Issue
#471 

### Background
- To prove features working, it is very important to make a test scenario in a client position.
- Existing integration test which is written in shell script with curl, jq, ... etc is not enough to achieve the goal

### Solution
- Build a client module. By using the client module, make a test scenario in client-side. (The scenario is not written yet)

- The client module should be separated in another project. But for now, it needed to be built in same repository. Hence, the client-side container share same build output from sebak. That is why the `Dockerfile(s)` are revised.

- To make the test easy on local machine, provide the `run.sh` script. IMHO, there is a room for extending the usability of `run.sh` in future.


Dockerfile.build
```diff
@@ -0,0 +1,27 @@
...
+COPY ./ /sebak-build
+WORKDIR /sebak-build
+
+## Note that we do not get the dependencies anew
+## We carry over whatever is in `vendor`, so the user MUST run `dep ensure` in their local copy
+## This make building the container orders of magnitude faster (`dep ensure` is extremely slow),
+## greatly reduce the container's size, and gives more control to the user as to what is tested
+## (one can replace a dependency, if needed).
+## Otherwise, install dep and dependencies.(e.g Docker Hub automated build)
+RUN if [ ! -d vendor ]; then go get github.com/golang/dep/cmd/dep; dep ensure -v; fi
```
Geod24: This is outdated

kfangw: I was worked on master.
Is it updated?

docker/entrypoint.sh
```diff
@@ -1,9 +1,5 @@
 #!/bin/sh
 
-if [ -f ".env" ]; then
-    source ./.env
-fi
```
Geod24: I think @zzim2x plans to use it

Geod24:
For the tests, I have a preference for `node`, because we're really testing the core functionalities.

Dockerfile.build
Geod24: Look at what's in master

kfangw:
> For the tests, I have a preference for `node`, because we're really testing the core functionalities.

What do you mean by `node` ?
You mean not the container?

kfangw:
@spikeekips @Geod24 Please review this PR.
while you are reviewing this, I will extend the `tests/sdk/main.go` to meet test scenario.

Dockerfile.build
kfangw: I will handle this Fail. https://circleci.com/gh/bosnet/sebak/8902?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

It looks like due to this.

kfangw:
I removed the `coverage` from the integration test.

IMO, Checking coverage is a role for unit test, not for integration test.

Furthermore, The build option was `test` not `install`. It means that the unit test is executed twice at CI. It makes the CI very slow.
Hence, I changed `BUILD_MODE` of Dockerfile from `test` to `install`.
And remove relevant coverage build options.

Is it makes sense? @Geod24

kfangw:
The CI failed https://circleci.com/gh/bosnet/sebak/8933?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
is due to this PR? or came from master? @spikeekips

Geod24:
> Is it makes sense? @Geod24

Not to me, no. There are some things the unittests should / can not tests, such as anything IO related, which is why having integration tests spit out coverage is very important.

> Furthermore, The build option was `test` not `install`. It means that the unit test is executed twice at CI. It makes the CI very slow.

[No they're not](https://github.com/bosnet/sebak/blob/master/cmd/sebak/main_test.go#L13-L17)

kfangw:
> Furthermore, The build option was `test` not `install`. It means that the unit test is executed twice at CI. It makes the CI very slow.
> Hence, I changed `BUILD_MODE` of Dockerfile from `test` to `install`.
> And remove relevant coverage build options.

@Geod24
Original build command was  
https://github.com/bosnet/sebak/blob/71e6f21d7561daf2b276227f24aebe75f3cd2a59/tests/run.sh#L29-L32

And here is one of option for `go test`
```
	-c
	    Compile the test binary to pkg.test but do not run it
	    (where pkg is the last element of the package's import path).
	    The file name can be changed with the -o flag.
```

by using the `-c` flag, the original build also does not execute the test.
It was my misunderstanding.

And the coverage of the integration test will be reverted.
Please wait for my fixing.

@Geod24 @spikeekips

kfangw:
@Geod24 @spikeekips I reverted what I mentioned above.

And I don't know any reason why the CI failed https://circleci.com/gh/bosnet/sebak/8978?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
Is it related what you mentioned in the morning? @spikeekips

spikeekips:
@kfangw That failed is not part of mine, this was caused by,
```
+ docker cp :/sebak/coverage.txt ./account-creation/coverage/node1/coverage.txt
must specify at least one container source
```
You missed container name. :)

kfangw:
> @kfangw That failed is not part of mine, this was caused by,
> 
> ```
> + docker cp :/sebak/coverage.txt ./account-creation/coverage/node1/coverage.txt
> must specify at least one container source
> ```
> You missed container name. :)

Ah! I should have read the log carefully. Thanks! @spikeekips

spikeekips:
> @kfangw : What is right?

No :) not right.

Dockerfile
```diff
@@ -1,28 +1,10 @@
...
-RUN go $BUILD_MODE $BUILD_ARGS -ldflags="$(govvv -pkg boscoin.io/sebak/lib/version -flags)" -v $BUILD_PKG
-
 ## This one is much more lightweight
 FROM alpine:latest AS runner
 
 RUN apk --no-cache add ca-certificates      # For SSL requests
 COPY docker/entrypoint.sh /sebak/entrypoint.sh
 COPY docker/sebak.* /sebak/
-COPY --from=builder /go/bin/sebak /sebak/
+COPY --from=sebak:builder /go/bin/sebak /sebak/
```
Geod24: What I don't like about this setup is that it forces the user to use a specific tag.
If someone wants to use, e.g. `sebak-stable`, this will break, but the user might not see it if there's already a `sebak` tagged.
It also relies more on user's virtue to build `sebak` before building this, right ?

lib/client/client.go
```diff
@@ -0,0 +1,305 @@
...
+type Client struct {
+	URL string
+
+	HTTP *common.HTTP2Client
+}
+
+func NewClient(url string) *Client {
+	httpClient, err := common.NewHTTP2Client(0, 0, true)
+	if err != nil {
+		//TODO: handle Error
```
Geod24: If you don't want to handle the error now, it's fine, but this should `panic`

kfangw:
@Geod24 please see this commit https://github.com/bosnet/sebak/pull/486/commits/823ebada95eddf4659f98b024ba4f0e1014e5dd7

Geod24:
> What do you mean by node ?

Sorry forgot to answer that. You moved `test/*` to `test/api/*`. My suggestion was `test/node/*`

docker/self.env
```diff
@@ -4,8 +4,10 @@ SEBAK_SECRET_SEED=SBECGI3FSCYHNQIMANNCWQSVA6S5C6L4BXFKAPMBAMI5V47NWXNE37MN
 # Not strictly needed, kept for documentation purpose
 SEBAK_PUBLIC_KEY=GDIRF4UWPACXPPI4GW7CMTACTCNDIKJEHZK44RITZB4TD3YUM6CCVNGJ
 # Endpoint sebak binds to
-SEBAK_ENDPOINT=https://127.0.0.1:2830
+SEBAK_BIND=https://127.0.0.1:2830
```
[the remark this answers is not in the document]
kfangw: What test did you mean?

lib/network/api/resource/account.go
```diff
@@ -21,10 +21,10 @@ func NewAccount(ba *block.BlockAccount) *Account {
 
 func (a Account) GetMap() hal.Entry {
 	return hal.Entry{
-		"address":    a.ba.Address,
-		"sequenceid": a.ba.SequenceID,
```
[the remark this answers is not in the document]
kfangw: I prefer to follow conventions of json

tests/run.sh
```diff
@@ -5,71 +5,9 @@ set -xe
...
-    mkdir -p ${dir}/coverage/node{1,2,3}/
-    docker cp ${NODE1}:/sebak/coverage.txt ${dir}/coverage/node1/coverage.txt
-    docker cp ${NODE2}:/sebak/coverage.txt ${dir}/coverage/node2/coverage.txt
-    docker cp ${NODE3}:/sebak/coverage.txt ${dir}/coverage/node3/coverage.txt
-
-    # Cleanup
-    docker rm -f ${NODE1} ${NODE2} ${NODE3} || true
-done
+./api/run.sh
+./sdk/run.sh
```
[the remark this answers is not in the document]
kfangw: the config file will be edited in new commit

tests/sdk/build.sh
```diff
@@ -0,0 +1,29 @@
...
+
+BUILD_MODE="install"
+BUILD_PKG="./..."
+BUILD_ARGS=""
+
+## Build the docker builder image
+IMAGE=$(docker build --tag sebak:builder_sdk -q \
+    --build-arg BUILD_MODE=${BUILD_MODE} \
+    --build-arg BUILD_PKG=${BUILD_PKG} \
+    --build-arg BUILD_ARGS="${BUILD_ARGS}"  \
```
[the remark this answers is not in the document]
kfangw: You are right. wait for another commit

tests/sdk/run.sh
```diff
@@ -0,0 +1,34 @@
...
+    if [ ! -z "${NODE}" ]; then
+        docker logs ${NODE} || true
+        docker rm -f ${NODE} || true
+    fi
+}
+
+trap dumpLogsAndCleanup EXIT
+
+NODE=$(docker run -d --network host --env-file=${ROOT_DIR}/docker/self.env \
+      sebak:runner node \
```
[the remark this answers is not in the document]
kfangw: I prefer build image and run in the separated files. But I don't mind that put them one file.
Please wait another commit

Dockerfile
```diff
@@ -14,6 +10,11 @@ RUN go get github.com/ahmetb/govvv
 COPY ./ /sebak-build
 WORKDIR /sebak-build
 
+# You probably don't need to change this
+ARG BUILD_MODE="install"
+ARG BUILD_ARGS=''
+ARG BUILD_PKG="./..."
+
```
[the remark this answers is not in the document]
kfangw: yeap. It will be reverted.

tests/sdk/entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+set -xe
+echo "IN THE ENTRYPOINT"
```
Geod24: Aren't you missing something here ?

kfangw: The entrypoint and others will be filled with Another PR soon.
Please ignore that for now.

Geod24: How can I make sure that the code added in this PR works then ?

kfangw: What code you did you mean?

tests/sdk/main.go
```diff
@@ -0,0 +1,21 @@
...
+	c := client.NewClient("https://127.0.0.1:2830")
+	headers := http.Header{}
+	headers.Set("Content-Type", "application/json")
+	a, e := c.LoadAccount("GDIRF4UWPACXPPI4GW7CMTACTCNDIKJEHZK44RITZB4TD3YUM6CCVNGJ")
+	if e != nil {
+		fmt.Println(e)
+	}
+	fmt.Println(a)
+
+	os.Exit(0)
```
Geod24: That's unnecessary

tests/sdk/entrypoint.sh
kfangw: There is executable file `sdk` and it is came from the package `sdk` the package include `main.go`.
when `./sdk` statement executed then, the relevant test cases are executed.

tests/sdk/run.sh
```diff
@@ -0,0 +1,64 @@
...
+    --build-arg BUILD_MODE="install" \
+    --build-arg BUILD_PKG="./tests/sdk" \
+    ${ROOT_DIR}/ -f ${ROOT_DIR}/Dockerfile_client.build | cut -d: -f2)
+
+if [ -z ${IMAGE} ]; then
+    echo "Failed to build builder docker image" >&2
+    exit 1
+fi
+
+IMAGE=$(docker build --tag sebak:sdk_tester -q \
```
Geod24: I don't think the image should be tagged, as discussed

tests/sdk/main.go
kfangw: As I mentioned before, It is a dummy code.
The main.go and relevant code will be added to test client and relevant node features

tests/sdk/run.sh
kfangw: OK then, it will be reverted.

tests/sdk/main.go
Geod24: I am just referring to the `os.Exit` statement

tests/sdk/entrypoint.sh
Geod24: Oh, nevermind, I'm blind...
Just remove the echo and we're good 👍

Geod24: Note that you can just set the binary as entrypoint if you wish, also. But maybe you want to extend this script ?

tests/sdk/main.go
kfangw: I know what the os.Exit(0) means and it is redundant.
It is a `dummy script` to make sure that always exit with 0 returned value.

tests/sdk/entrypoint.sh
kfangw: yes. I have a plan to extend the script to support various kind of testing needs

tests/node/run.sh
```diff
@@ -0,0 +1,86 @@
...
+    ${ROOT_DIR}/ | cut -d: -f2)
+
+if [ -z ${IMAGE} ]; then
+    echo "Failed to build node runner docker image" >&2
+    exit 1
+fi
+
+
+# Build the integration test docker image
+IMAGE=$(docker build --tag sebak:api_tester -q \
```
Geod24: This file has a lot of modifications that shouldn't be there (it should be simply moved). Can you revert those ?

lib/network/api/resource/account.go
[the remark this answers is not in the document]
Geod24: Which is fine, but it should come in its own pull request. It's not necessary for the new integration tests to work.

tests/node/run.sh
kfangw: sure.

kfangw:
@Geod24 please confirm the latest commit.
I hope that it is the last commit for this PR ^^a
Thanks for your careful review.

tests/node/run.sh
```diff
@@ -72,4 +72,4 @@ for dir in ${TEST_DIRS}; do
 
     # Cleanup
     docker rm -f ${NODE1} ${NODE2} ${NODE3} || true
-done
+done
```
Geod24: Still some changes

kfangw: damn it. kkkk

tests/node/run.sh
```diff
@@ -6,7 +6,7 @@ set -xe
 # So make sure we're in the right WD
 cd -- `dirname ${BASH_SOURCE[0]}`
 TEST_DIRS=$(find . -mindepth 1 -maxdepth 1 -type d -print)
-ROOT_DIR=".."
+ROOT_DIR="../.."
```
Geod24: And here...

kfangw:
I hope that everything is in right place. @Geod24

tests/sdk/run.sh
```diff
@@ -0,0 +1,64 @@
...
+    --build-arg BUILD_ARGS="-coverpkg=./... -tags integration -c -o /go/bin/sebak"  \
+    ${ROOT_DIR}/ | cut -d: -f2)
+
+if [ -z ${NODE_IMAGE} ]; then
+    echo "Failed to build node runner docker image" >&2
+    exit 1
+fi
+
+## Build the docker builder image
+IMAGE=$(docker build --tag sebak:builder_sdk -q \
```
Geod24: There's still a tag being done here. Remove it, and instead add an `ARG` to your runner which takes the image to copy from.

kfangw: I tried that but It is impossible to do that.
In a docker file I can not set --from=imagehash

Geod24: Change your Dockerfile to the following:
```Dockerfile
## Builder part is quite heavy as it depends on the Go toolchain
## Work around Moby bug 34482
ARG BUILDER
FROM ${BUILDER} AS builder

FROM alpine:latest
LABEL maintainer="BOSCoin Developers <devteam@boscoin.io>"

COPY --from=builder /go/bin/sdk /sebak/
COPY ./entrypoint.sh /sebak/

WORKDIR /sebak/
ENTRYPOINT [ "./entrypoint.sh" ]
CMD []
```

And apply this diff to `run.sh`:
```diff
diff --git a/tests/sdk/run.sh b/tests/sdk/run.sh
index 2519f1c..2be6077 100755
--- a/tests/sdk/run.sh
+++ b/tests/sdk/run.sh
@@ -31,7 +31,7 @@ if [ -z ${NODE_IMAGE} ]; then
 fi

 ## Build the docker builder image
-IMAGE=$(docker build --tag sebak:builder_sdk -q \
+IMAGE=$(docker build -q \
     --build-arg BUILD_MODE="install" \
     --build-arg BUILD_PKG="./tests/sdk" \
     ${ROOT_DIR}/ -f ${ROOT_DIR}/Dockerfile_client.build | cut -d: -f2)
@@ -41,7 +41,7 @@ if [ -z ${IMAGE} ]; then
     exit 1
 fi

-TEST_IMAGE=$(docker build -q \
+TEST_IMAGE=$(docker build -q --build-arg BUILDER_IMAGE=${IMAGE} \
     . | cut -d: -f2)

 if [ -z ${TEST_IMAGE} ]; then
```

kfangw: wow. I thought only `COPY` statement.
Thx!

Geod24: Yep it's a bug in Moby (docker), I was about to report it but found someone already did, and there is a workaround.

kfangw: DONE!

tests/sdk/Dockerfile
```diff
@@ -1,8 +1,11 @@
 ## Builder part is quite heavy as it depends on the Go toolchain
+ARG BUILDER
```
Geod24: You forgot the mention of the bug :(

kfangw: sorry ;

PR #488, opened 2018-10-04
Syncer no incr height

anarcher:
### Github Issue

It fixed #487 test fail.

### Background

Syncer doesnot increase height itself.


anarcher:
@Charleslee522 Can you check it? :-)

PR #489, opened 2018-10-04
Add Sync Watcher

anarcher:
### Background

- `sync.Watcher` watches connected nodes by using `network.connectionManager` and then get highest height in the node network. 
- It can be useful watcher node mode(a node out of consensus network), because  with consensus mode, consensus can get highest height instead `sync.Watcher`
- TODO: How do interact with consensus with watcher and syncer?

### DIFF
--- a/lib/node/validator.go
+++ b/lib/node/validator.go
@@ -19,10 +19,11 @@ import (
 )
 
 type ValidatorFromJSON struct {
-	Alias    string           `json:"alias"`
-	Address  string           `json:"address"`
-	Endpoint *common.Endpoint `json:"endpoint"`
-	State    State            `json:"state"`
+	Alias       string           `json:"alias"`
+	Address     string           `json:"address"`
+	Endpoint    *common.Endpoint `json:"endpoint"`
+	State       State            `json:"state"`
+	BlockHeight uint64           `json:"block_height"`
 }
 
 type Validator struct {
@@ -32,6 +33,8 @@ type Validator struct {
 	alias    string
 	address  string
 	endpoint *common.Endpoint
+
+	blockHeight uint64
 }
 
 func (v *Validator) String() string {
@@ -80,10 +83,11 @@ func (v *Validator) Endpoint() *common.Endpoint {
 
 func (v *Validator) MarshalJSON() ([]byte, error) {
 	return json.Marshal(map[string]interface{}{
-		"address":  v.Address(),
-		"alias":    v.Alias(),
-		"endpoint": v.Endpoint().String(),
-		"state":    v.State().String(),
+		"address":      v.Address(),
+		"alias":        v.Alias(),
+		"endpoint":     v.Endpoint().String(),
+		"state":        v.State().String(),
+		"block_height": v.blockHeight,
 	})
 }
 
@@ -97,12 +101,17 @@ func (v *Validator) UnmarshalJSON(b []byte) error {
 	v.address = va.Address
 	v.endpoint = va.Endpoint
 	v.state = va.State
+	v.blockHeight = va.BlockHeight
 
 	return nil
 }
 
 func (v *Validator) Serialize() ([]byte, error) {
 	return json.Marshal(v)
+}
+
+func (v *Validator) BlockHeight() uint64 {
+	return v.blockHeight
 }
 
 func NewValidator(address string, endpoint *common.Endpoint, alias string) (v *Validator, err error) {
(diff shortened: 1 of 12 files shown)


anarcher:
Another approach here. 
If Iaasc.StartConsenus(height) exists, the sync.Watcher can call it. :-)

anarcher:
What do you think about that?

```
watcher.SetConsensusWatchMode() // It means watcher itself doesn't watches  but  it delegates consensus's highest height  to `syncer`.
watcher.SetNodeWatchMode() // It is default mode, watcher itself watches highest height. 

watcher.SetSyncTargetHeight(height) // too. :-) 
```

lib/node/validator.go
```diff
@@ -19,10 +19,11 @@ import (
...
 type ValidatorFromJSON struct {
-	Alias    string           `json:"alias"`
-	Address  string           `json:"address"`
-	Endpoint *common.Endpoint `json:"endpoint"`
-	State    State            `json:"state"`
+	Alias       string           `json:"alias"`
+	Address     string           `json:"address"`
+	Endpoint    *common.Endpoint `json:"endpoint"`
+	State       State            `json:"state"`
+	BlockHeight uint64           `json:"block_height"`
```
spikeekips: Usually we use `-` in json field name. For instance, please see `round.Round`

```
BlockHeight uint64 `json:"block-height"`
```

spikeekips:
@anarcher I want you to check the another approach to check or get the block height of other nodes. Instead of use API, how about using Ballot? In `Ballot`, block height is in `Round.BlockHeight` and we can set the node's BlockHeight in `Ballot` header, we can check the node's latest block height *directly*.

anarcher:
@spikeekips I changed `block_height` to `block-height` :-)

anarcher:
> @anarcher I want you to check the another approach to check or get the block height of other nodes. Instead of use API, how about using Ballot? In Ballot, block height is in Round.BlockHeight and we can set the node's BlockHeight in Ballot header, we can check the node's latest block height directly

If a node get ballot, It means the node is in consensus network, With this case, this watch is useless. the consensus module can get highest height instead of sync.watcher. 
This watcher can be useful with a node out of consensus network. 

I have already provided this method (syncer.SetSyncTargetHeight) to @Charleslee522

Charleslee522:
> > @anarcher I want you to check the another approach to check or get the block height of other nodes. Instead of use API, how about using Ballot? In Ballot, block height is in Round.BlockHeight and we can set the node's BlockHeight in Ballot header, we can check the node's latest block height directly
> 
> If a node get ballot, It means the node is in consensus network, With this case, this watch is useless. the consensus module can get highest height instead of sync.watcher.
> This watcher can be useful with a node out of consensus network.
> 
> I have a question that we dones't need to watch node (nodes out of consensus network)?
> 
> I have already provided this method (`syncer.SetSyncTargetHeight`) to @Charleslee522

@spikeekips Please check https://github.com/bosnet/sebak/pull/487 :)

spikeekips:
@anarcher : I have a question that we dones't need to watch node (nodes out of consensus network)?

Currently we PING-PING for checking they are alive or not. To checking block height, I think we don't need PING-PING :)

@Charleslee522 :  Please check #487 :)

Thanks, I did read your `syncer.SetSyncTargetBlock`, but it is not what I mentioned. I said, with the incoming ballot from other node, we can check the latest block height they have without asking block height info.

Charleslee522:
> @Charleslee522 : Please check #487 :)
> 
> Thanks, I did read your `syncer.SetSyncTargetBlock`, but it is not what I mentioned. I said, with the incoming ballot from other node, we can check the latest block height they have without asking block height info.

@spikeekips [This part](https://github.com/bosnet/sebak/pull/487/files#diff-6f58f297378e814449bd2f87bbd85f66R124) is called when ballot comes in.

anarcher:
> @anarcher : I have a question that we dones't need to watch node (nodes out of consensus network)? Currently we PING-PING for checking they are alive or not. To checking block height, I think we don't need PING-PING :)

Ok, I have no idea to check block height in a network `ㅠ_ㅠ` `network.Network` can know it?

spikeekips:
@anarcher As I introduced, we can simply find the latest block information of the other node with `Ballot`.

* Add block information in `Ballot` header
* The other node check the block information from the received `Ballot`

anarcher:
I got it. :-) This watcher is useless and not good.. `ㅠ_ㅠ`. I will close it. Sorry for interrupting `ㅠ_ㅠ`

PR #491, opened 2018-10-04
Get Missing Transactions From Proposer

### LINKED ISSUE #479: Fill the missing transaction from proposed Ballot
Under new design of current ISAAC, there will be the unknown transaction from the proposed Ballot. During consensus, these missing transactions must be fetched from proposer. This feature is already reside in code, `runner.GetMissingTransaction()`, we need to fill the code in it. :)

spikeekips:
### Github Issue
resolves #479 

### Background

Under current new designed ISAAC, the incoming transactions can not sure to be broadcasted to the entire nodes, so some node maybe receives the proposed ballot before getting some transactions of ballot.

### Solution

The `runner.GetMissingTransactions()` will get the missing transactions from proposer thru node api, `/node/transactions`(the proposer must know the transactions of ballot.)

### Changes

* `NetworkClient` will have new method, `GetTransactions`.
* `checker.BallotGetMissingTransaction()` will get the missing transaction `checker.BallotChecker` in `INIT` state.
* To get the missing `Transactions` will be processed in 2 state, `INIT` and `ACCEPT`.
    * `INIT`: node validates the incoming Ballot with the `Transactions` of `Ballot`
    * `ACCEPT`: node needs to know the all the `Transactions` of `Ballot` to store it.
* If `checker.BallotGetMissingTransaction()` fail to get the transactions from proposer in `INIT`, it will vote `Ballot.VotingNO` in `Ballot`.
* If `checker.FinishedBallotStore()` fails to get the transactions from proposer in `ACCEPT`, it will not save the new block.

### DIFF
--- a/lib/node/runner/checker.go
+++ b/lib/node/runner/checker.go
@@ -299,7 +299,7 @@ func getMissingTransaction(checker *BallotChecker) (err error) {
 			err = errors.ErrorTransactionNotFound
 			return
 		}
-		if err = tx.IsWellFormed(checker.NetworkID); err != nil {
+		if err = tx.IsWellFormed(checker.NetworkID, checker.NodeRunner.Conf); err != nil {
 			return
 		}
 
--- a/lib/node/runner/checker.go
+++ b/lib/node/runner/checker.go
@@ -465,6 +465,11 @@ func FinishedBallotStore(c common.Checker, args ...interface{}) (err error) {
 		return
 	}
 	if checker.FinishedVotingHole == ballot.VotingYES {
+		if err = getMissingTransaction(checker); err != nil {
+			checker.Log.Debug("failed to get the missing transactions of ballot", "error", err)
+			return
+		}
+
 		var theBlock block.Block
 		theBlock, err = finishBallot(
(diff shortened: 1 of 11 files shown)


lib/node/runner/checker.go
```diff
@@ -376,6 +464,11 @@ func FinishedBallotStore(c common.Checker, args ...interface{}) (err error) {
 		return
 	}
 	if checker.FinishedVotingHole == ballot.VotingYES {
+		if err = getMissingTransaction(checker); err != nil {
```
kfangw: Is there a chance that any Tx are missing when the ballot passed the consensus?
@spikeekips

spikeekips: @kfangw It will be rarely occurred, but yes.

1. One node failed to receive the ballots of previous state.
1. The node received the ballots of ACCEPT state.
1. The node must confirm the ACCEPTed ballot, but the node does not know about the transactions of ACCEPTed ballot.

kfangw: Understood. Thx

PR #494, opened 2018-10-05
Ordering txs by block.Transactions

anarcher:
### Background

- Fetched transactions would be nice to order by fetched `block.Transactions = []hash`
- I think  `blocks API` may make a response by this ordering. But the syncer should do order by `block.Transactions`.

### DIFF
--- a/lib/sync/fetcher.go
+++ b/lib/sync/fetcher.go
@@ -124,7 +124,7 @@ func (f *BlockFetcher) fetch(ctx context.Context, si *SyncInfo) error {
 	}
 	if resp.StatusCode == http.StatusNotFound {
 		//TODO:
-		err := errors.New("Fetch: block not found")
+		err := errors.New("fetch: block not found")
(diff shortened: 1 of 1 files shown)


lib/sync/fetcher.go
```diff
@@ -159,10 +161,21 @@ func (f *BlockFetcher) fetch(ctx context.Context, si *SyncInfo) error {
...
+			//TODO(anarcher): Error type for controlling timeout
+			err := fmt.Errorf("Tx: %s not found in block height %d", hash, height)
+			return err
+		}
+		si.Txs = append(si.Txs, tx)
 
-		si.Txs = append(si.Txs, &tx)
 	}
 
+	//TODO(anarcher): Remove this comments
```
Charleslee522: How about just removing this and below comments?

anarcher: Ok! I will do it :->

PR #495, opened 2018-10-05
Changes SetSyncTargetBlock(ctx,height,nodeAddrs)

anarcher:
### Background

```
type SyncController interface {
	SetSyncTargetBlock(ctx context.Context, height uint64, nodeAddressList []string) error
}
```

if syncer and fetcher get this `nodeAddressList`,  fetcher choose node of them (nodeAddressList) randomly.

### DIFF
--- a/lib/sync/fetcher.go
+++ b/lib/sync/fetcher.go
@@ -92,10 +92,13 @@ func (f *BlockFetcher) Fetch(ctx context.Context, syncInfo *SyncInfo) (*SyncInfo
 }
 
 func (f *BlockFetcher) fetch(ctx context.Context, si *SyncInfo) error {
-	var height = si.BlockHeight
+	var (
+		height    = si.BlockHeight
+		nodeAddrs = si.NodeAddrs
+	)
 	f.logger.Debug("Fetch start", "height", height)
 
-	n := f.pickRandomNode()
+	n := f.pickRandomNode(nodeAddrs)
 	f.logger.Info(fmt.Sprintf("fetching items from node: %v", n), "fetching_node", n, "height", height)
 	if n == nil {
 		return errors.New("Fetch: node not found")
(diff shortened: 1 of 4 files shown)


lib/sync/types.go
```diff
@@ -25,16 +25,19 @@ func (s SyncProgress) Clone() *SyncProgress {
...
 	Block       *block.Block
 	Txs         []*transaction.Transaction
 
-	BlockTxs []*block.BlockTransaction
-	BlockOps []*block.BlockOperation
+	NodeAddrs []string
+
+	//TODO(anarcher): Remove
+	//BlockTxs []*block.BlockTransaction
+	//BlockOps []*block.BlockOperation
```
Charleslee522: If these are not used, then I figure it would be better to remove :)

anarcher:
ping @Charleslee522 @spikeekips :->

spikeekips:
@anarcher One question, `nodeAddressList` is the validators of node?

Charleslee522:
> One question, nodeAddressList is the validators of node?

@spikeekips Yes, I think it is latest voted validators +2/3 :)

spikeekips:
> @Charleslee522 : > One question, nodeAddressList is the validators of node?
>> @spikeekips Yes, I think it is latest voted validators +2/3 :)

I see.

lib/sync/fetcher.go
```diff
@@ -174,15 +177,27 @@ func (f *BlockFetcher) fetch(ctx context.Context, si *SyncInfo) error {
 }
 
 // pickRandomNode choose one node by random. It is very protype for choosing fetching which node
-func (f *BlockFetcher) pickRandomNode() node.Node {
+func (f *BlockFetcher) pickRandomNode(nodeAddrs []string) node.Node {
```
spikeekips: If https://github.com/bosnet/sebak/pull/495#issuecomment-428420886 is right, I wish `nodeAddrs` needs another understandable name, or you can explain what is `nodeAddrs` in comment.

anarcher: type SyncInfo struct {
	BlockHeight uint64
	Block       *block.Block
	Txs         []*transaction.Transaction

	NodeAddrs []string // Fetching target node addresses
}

It have already comments :-)

spikeekips: @anarcher I wish this kind of things :)
```
NodeAddrs []string // Fetching target node addresses, `NodeAddrs` is  the validators which 
                   //participated and confirmed the consensus of latest ballot.
```

anarcher: @spikeekips Thanks a lot!! :->

PR #498, opened 2018-10-08
Validating Incoming Transactions in API

spikeekips:
### Background
If not-wellformed transaction is received, the API just eat it and return 'status ok'. If API handler returns the validated or not to client, it will be useful for client.

### Solution

`NetworkHandlerNode.MessageHandler` will do the `runner.HandleTransactionCheckerFuncs` using `Checker` and the existing `NodeRunner.handleTransaction()` was removed.

### DIFF
--- a/lib/common/router_test.go
+++ b/lib/common/router_test.go
@@ -23,7 +23,7 @@ func TestRouterHeaderMatcher(t *testing.T) {
 	u, _ := url.Parse(server.URL)
 	u.Path = "/showme"
 
-	{ // GET msut be passed
+	{ // GET must pass
 		req, _ := http.NewRequest("GET", u.String(), nil)
 		resp, err := server.Client().Do(req)
 		require.Nil(t, err)
--- a/lib/common/router_test.go
+++ b/lib/common/router_test.go
@@ -1,0 +1,99 @@
+package common
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRouterHeaderMatcher(t *testing.T) {
+	router := mux.NewRouter()
+
+	dummyHandler := func(w http.ResponseWriter, r *http.Request) {
+		defer r.Body.Close()
+		return
+	}
+
+	router.HandleFunc("/showme", dummyHandler).MatcherFunc(PostAndJSONMatcher)
+	server := httptest.NewServer(router)
+	u, _ := url.Parse(server.URL)
+	u.Path = "/showme"
+
+	{ // GET msut be passed
+		req, _ := http.NewRequest("GET", u.String(), nil)
+		resp, err := server.Client().Do(req)
+		require.Nil(t, err)
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+	}
+
+	{ // POST && empty 'Content-Type'
+		req, _ := http.NewRequest("POST", u.String(), nil)
+		resp, err := server.Client().Do(req)
+		require.Nil(t, err)
+		require.Equal(t, http.StatusNotFound, resp.StatusCode)
+	}
+
+	{ // POST && invalid 'Content-Type'
+		req, _ := http.NewRequest("POST", u.String(), nil)
+		req.Header.Set("Content-Type", "text/plain")
+		resp, err := server.Client().Do(req)
+		require.Nil(t, err)
+		require.Equal(t, http.StatusNotFound, resp.StatusCode)
+	}
+
+	{ // POST && valid 'Content-Type'
+		req, _ := http.NewRequest("POST", u.String(), nil)
+		req.Header.Set("Content-Type", "application/json")
+		resp, err := server.Client().Do(req)
+		require.Nil(t, err)
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+	}
+}
+
+func TestRouterHeaderMatcherWithMethodMatcher(t *testing.T) {
+	router := mux.NewRouter()
+
+	dummyHandler := func(w http.ResponseWriter, r *http.Request) {
+		defer r.Body.Close()
+		return
+	}
+
+	router.HandleFunc("/showme", dummyHandler).Methods("GET", "POST").MatcherFunc(PostAndJSONMatcher)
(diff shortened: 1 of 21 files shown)


lib/network/httputils/httputils.go
```diff
@@ -15,59 +16,17 @@ func IsEventStream(r *http.Request) bool {
...
-		145: 400,
-	}
+	// ErrorsToStatus defines errors.Error does not have 400 status code.
+	ErrorsToStatus = map[uint]int{}
 )
 
 func StatusCode(err error) int {
 	if e, ok := err.(*errors.Error); ok {
-		return ErrorsToStatus[e.Code]
+		if c, ok := ErrorsToStatus[e.Code]; ok {
```
kfangw: 👍

lib/transaction/transaction.go
```diff
@@ -175,6 +178,7 @@ func (tx Transaction) String() string {
 }
 
 func (tx *Transaction) Sign(kp keypair.KP, networkID []byte) {
+	tx.B.Source = kp.Address()
```
kfangw: What is it?

spikeekips: @kfangw I found the missing part of this. As we know, `Sign()` must be done by `Source` account and the existing code set the `Source` manually.

spikeekips:
@anarcher Ping~

lib/common/router_test.go
```diff
@@ -0,0 +1,99 @@
...
+		defer r.Body.Close()
+		return
+	}
+
+	router.HandleFunc("/showme", dummyHandler).MatcherFunc(PostAndJSONMatcher)
+	server := httptest.NewServer(router)
+	u, _ := url.Parse(server.URL)
+	u.Path = "/showme"
+
+	{ // GET msut be passed
```
Geod24: Typo: "must"
Also, "must pass" not "must be passed"

lib/network/http2_client.go
```diff
@@ -67,13 +67,13 @@ func (c *HTTP2NetworkClient) GetNodeInfo() (body []byte, err error) {
 		return
 	}
 	defer response.Body.Close()
+	body, err = ioutil.ReadAll(response.Body)
+
 	if response.StatusCode != http.StatusOK {
 		err = errors.ErrorHTTPProblem.Clone().SetData("status", response.StatusCode)
 		return
```
Geod24: This `return is now unnecessary

lib/network/http2_client.go
```diff
@@ -88,13 +88,13 @@ func (c *HTTP2NetworkClient) Connect(n node.Node) (body []byte, err error) {
...
```
Geod24: Ditto, no need to `return`

lib/network/memory.go
```diff
@@ -18,6 +18,8 @@ type MemoryNetwork struct {
 	receiveChannel chan common.NetworkMessage
 	// They all share the same map to find each other
 	peers map[ /* endpoint */ string]*MemoryNetwork
+
+	messageBroker MessageBroker
```
Geod24: Does it have any effect ?

lib/node/runner/node_runner_test.go
```diff
@@ -238,29 +238,6 @@ func TestCreateNodeRunner(t *testing.T) {
 	require.Equal(t, 3, len(nodeRunners))
 }
 
-// Check that when a node receives transaction, the node broadcasts it to validators
-func TestNodeRunnerTransactionBroadcast(t *testing.T) {
```
Geod24: Why is this removed ?

spikeekips: @Geod24 This test does not test to broadcast Transaction. As you see, it checks it's own `TransactionPool`. We need new test code for broadcasting.

lib/node/runner/api_message_handler_test.go
```diff
@@ -0,0 +1,200 @@
...
+	p.Prepare()
+	defer p.Done()
+
+	u := p.URL(nil)
+
+	{ // invalid signature
+		tx := p.makeTransaction()
+		tx.H.Signature = "findme"
+		errIsWellformed := tx.IsWellFormed(networkID, p.conf)
+		require.Equal(t, errors.ErrorInvalidTransaction.Code, errIsWellformed.(*errors.Error).Code)
```
Geod24: Why not just compare the errors ? Avoids the cast

Geod24:
As I've said before, it would much, much easier to review if such PRs were broken up into smaller, atomic commits.
It's extremely counter-productive to review a 500+ LoC of diff without any kind of logical separation.

lib/node/runner/api_message_handler_test.go
spikeekips:
```
lib/node/runner/api_message_handler_test.go:104:72: errIsWellformed.Code undefined (type error has no field or method Code)
```
:)

Geod24: `require.Equal(t, errors.ErrorInvalidTransaction, errIsWellformed)`

:)

spikeekips: @Geod24 :) :)

```
=== RUN   TestNodeMessageHandlerNotWellformedTransaction
--- FAIL: TestNodeMessageHandlerNotWellformedTransaction (0.01s)
    require.go:157:
        	Error Trace:	api_message_handler_test.go:104
        	Error:      	Not equal:
        	            	expected: &errors.Error{Code:0xa4, Message:"invalid transaction", Data:map[string]interface {}{}}
        	            	actual  : &errors.Error{Code:0xa4, Message:"invalid transaction", Data:map[string]interface {}{"error":"signature verification failed"}}

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,2 @@
        	            	-(*errors.Error)({"code":164,"message":"invalid transaction","data":{}})
        	            	+(*errors.Error)({"code":164,"message":"invalid transaction","data":{"error":"signature verification failed"}})

        	Test:       	TestNodeMessageHandlerNotWellformedTransaction
FAIL
```

Geod24: I see.

Then could we have an helper function somewhere which compare those errors, with documentation on why it's useful. Looking at this code, I cannot tell that the returned error has extra information. And I've seen this pattern used in a few places already.

PR #500, opened 2018-10-10
Fix wrong log message

Geod24:
Just two expression being inverted.


Geod24:
Self merging since it's trivial

Issue #375, opened 2018-09-12
Middleware for handlers


anarcher:
- For support common functionality to http handlers. 
   - e.g. Recover,Cache,RateLimit,Metric,CORS.....
- gorilla's router has a room for it (looks at `router.Use()`)

PR #510, opened 2018-10-11
Add Middleware

anarcher:
### Github Issue

Fixes: #375 

### Background

- `AddMiddleware` to `Network`

```
 AddMiddleware(string, ...mux.MiddlewareFunc)
```

- Add `RecoverMiddleware`

```
 nr.network.AddMiddleware(network.RouterNameAPI, network.RecoverMiddleware(false))
 nr.network.AddMiddleware(network.RouterNameNode, network.RecoverMiddleware(false))
```

### Possible Drawbacks

- [x] Add tests

### DIFF
--- a/lib/network/http2.go
+++ b/lib/network/http2.go
@@ -152,6 +152,16 @@ func (t *HTTP2Network) setNotReadyHandler() {
 	t.server.Handler = HTTP2Log15Handler{log: t.log, handler: t.router}
 }
 
+func (t *HTTP2Network) AddMiddleware(routerName string, mws ...mux.MiddlewareFunc) {
+	r, ok := t.routers[routerName]
+	if !ok {
+		// ignore it
+	}
+	for _, mw := range mws {
+		r.Use(mw)
+	}
+}
+
 func (t *HTTP2Network) AddHandler(pattern string, handler http.HandlerFunc) (router *mux.Route) {
 	var routerName string
 	var prefix string
--- a/lib/network/http2.go
+++ b/lib/network/http2.go
@@ -14,6 +14,7 @@ import (
 	"golang.org/x/net/http2"
 
 	"boscoin.io/sebak/lib/common"
+	"boscoin.io/sebak/lib/error"
 	"boscoin.io/sebak/lib/node"
 )
 
@@ -152,14 +153,15 @@ func (t *HTTP2Network) setNotReadyHandler() {
 	t.server.Handler = HTTP2Log15Handler{log: t.log, handler: t.router}
 }
 
-func (t *HTTP2Network) AddMiddleware(routerName string, mws ...mux.MiddlewareFunc) {
+func (t *HTTP2Network) AddMiddleware(routerName string, mws ...mux.MiddlewareFunc) error {
 	r, ok := t.routers[routerName]
 	if !ok {
-		// ignore it
+		return errors.ErrorNotMatcHTTPRouter
 	}
 	for _, mw := range mws {
 		r.Use(mw)
 	}
+	return nil
 }
 
 func (t *HTTP2Network) AddHandler(pattern string, handler http.HandlerFunc) (router *mux.Route) {
--- a/lib/network/http_middlewares.go
+++ b/lib/network/http_middlewares.go
@@ -12,6 +12,9 @@ import (
 )
 
 func RecoverMiddleware(logger log15.Logger) mux.MiddlewareFunc {
+	if logger != nil {
+		logger = log // use network.log
+	}
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			defer func() {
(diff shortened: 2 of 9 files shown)


lib/network/http2.go
```diff
@@ -152,6 +152,16 @@ func (t *HTTP2Network) setNotReadyHandler() {
 	t.server.Handler = HTTP2Log15Handler{log: t.log, handler: t.router}
 }
 
+func (t *HTTP2Network) AddMiddleware(routerName string, mws ...mux.MiddlewareFunc) {
+	r, ok := t.routers[routerName]
+	if !ok {
+		// ignore it
```
spikeekips: Instead of ignoring it, I think returning error is safe.

lib/node/runner/node_runner.go
```diff
@@ -149,6 +149,9 @@ func (nr *NodeRunner) Ready() {
 		nr.Conf,
 	)
 
+	nr.network.AddMiddleware(network.RouterNameAPI, network.RecoverMiddleware(false))
```
spikeekips: There are `-verbose` option in `cmd/sebak/cmd/run.go`, we can use it for`printStack`.

lib/network/http_middlewares.go
```diff
@@ -0,0 +1,31 @@
...
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			defer func() {
+				if r := recover(); r != nil {
+					err, ok := r.(error)
+					if !ok {
+						err = fmt.Errorf("panic: %v", r)
+					}
+					httputils.WriteJSONError(w, err)
+					log.Error("recover an panic", "err", err)
```
spikeekips: If RecoverMiddleware can have `log` argument, we can use `NodeRunner.Log()` for logging.

anarcher: You means that  `network.Middleware` use  `runner/NodeRunner.log` ?

spikeekips: At least, `RecoverMiddleware` is using inside `NodeRunner`, so it will be good.

lib/node/runner/node_runner.go
anarcher: Ok, I got it. :-)

lib/network/http2.go
anarcher: Ok, Good point :-)

kfangw:
Test cases should be provided.

lib/network/http_middlewares.go
```diff
@@ -0,0 +1,36 @@
...
+			defer func() {
+				if r := recover(); r != nil {
+					err, ok := r.(error)
+					if !ok {
+						err = fmt.Errorf("panic: %v", r)
+					}
+					httputils.WriteJSONError(w, err)
+					if logger == nil {
+						logger = log // use network.log
+					}
```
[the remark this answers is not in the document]
anarcher: Oh, If it looks great, I will move that. :->

lib/network/http_middlewares_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestRecoverMiddleware(t *testing.T) {
+	endpoint, err := common.NewEndpointFromString(
+		fmt.Sprintf("http://localhost:%s", getPort()),
+	)
+	require.Nil(t, err)
+
+	network, err := makeTestHTTP2NetworkForTLS(endpoint)
```
spikeekips: I suggest another approach for testing. Please check https://github.com/bosnet/sebak/blob/master/lib/network/api/test.go#L21 , this shows http handler test without creating HTTP2 server .

anarcher: That is better?  I thought testing with `Network.AddMiddleware` too. but if that is better, I will change it :-)

anarcher: https://github.com/bosnet/sebak/blob/master/lib/network/api/test.go#L21 is just creating `http test.Server` insteads of `network.Network`  in fact. :->

anarcher:
@spikeekips @kfangw @Geod24 Do you have any options? If not, can I merge it? :->

spikeekips:
@anarcher There can be no more 👍

Geod24:
> What kind of description do you need? :->

Usually, it's good to mention:
- Motivation (Why was the change introduced? What feature was missing? What was the bug? Etc...)
- Design specific (Why was it done this way?)
- Alternatives (What other possibilities were considered, and why they were not choosen?)

You don't always have to do all 3, but motivation is the very least.

spikeekips:
@Geod24 said 'Motivation' and I agree. The description of this PR is short, but #375 explains. 'Motivation' should be described for #375, if it is needed :(

Geod24:
Thanks @anarcher for updating #375 :)

anarcher:
> Thanks @anarcher for updating #375 :)

Welcome, @Geod24 Sorry for my fool English ㅠㅠ

PR #511, opened 2018-10-11
Node will have itself as validator

### LINKED ISSUE #503: Add node itself As it's validator automatically
Currently the node must set itself as it's validators(`--validators`) with `self`. It will be better to set node itself as it's validators without `self'.

spikeekips:
### Github Issue
resolves #503 

### Background

Please check #503 .

### Solution

`node.NewLocalNode()` will set itself as validator.

### DIFF
--- a/cmd/sebak/cmd/cmd_test.go
+++ b/cmd/sebak/cmd/cmd_test.go
@@ -30,7 +30,7 @@ func TestParseFlagSelfValidators(t *testing.T) {
 	require.Equal(t, validator.Endpoint().Port(), parsedValidator.Endpoint().Port())
 }
 
-func TestAddingSelfValidators(t *testing.T) {
+func TestAddingSelfValidatorsWithoutSelf(t *testing.T) {
 	flagNetworkID = "sebak-test-network"
 	flagValidators = "https://localhost:12346?address=GDPQ2LBYP3RL3O675H2N5IEYM6PRJNUA5QFMKXIHGTKEB5KS5T3KHFA2"
 	flagKPSecretSeed = "SCN4NSV5SVHIZWUDJFT4Z5FFVHO3TFRTOIBQLHMNPAZJ37K5A2YFSCBM"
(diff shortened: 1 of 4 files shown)


Charleslee522:
@spikeekips Can we set "self" in validators yet? I ask you because of the self mode.
We designed that only "self" in validators in self mode.

spikeekips:
@Charleslee522 I wrote, if `--validators` is empty, it will be 'self-mode'. I will update :)

Geod24:
@spikeekips : The discussion we had with @Charleslee522 is that an empty `--validators` could be a configuration mistake, while `--validators=self` is definitely intended.

spikeekips:
@Charleslee522 I updated .

cmd/sebak/cmd/cmd_test.go
```diff
@@ -17,10 +18,56 @@ func TestParseFlagSelfValidators(t *testing.T) {
...
-	validator := vs[0]
-	require.Equal(t, bindEndpoint.Host, validator.Endpoint().Host)
+	require.Equal(t, 2, len(localNode.GetValidators()))
+
+	parsedValidator, _ := node.NewValidatorFromURI(flagValidators)
+	validator := localNode.GetValidators()[parsedValidator.Address()]
+
+	require.Equal(t, validator.Address(), parsedValidator.Address())
+	require.Equal(t, validator.Endpoint().Host, parsedValidator.Endpoint().Host)
+	require.Equal(t, validator.Endpoint().Port(), parsedValidator.Endpoint().Port())
```
Charleslee522: How about adding test case only "self" in flagValidators?
Because there is no test case for specifying "self".

cmd/sebak/cmd/cmd_test.go
```diff
@@ -17,10 +18,56 @@ func TestParseFlagSelfValidators(t *testing.T) {
...
+}
+
+func TestAddingSelfValidators(t *testing.T) {
```
Charleslee522: 👍

cmd/sebak/cmd/cmd_test.go
```diff
@@ -17,10 +18,56 @@ func TestParseFlagSelfValidators(t *testing.T) {
...
+		require.True(t, found)
+	}
+}
+
+func TestAddingSelfWithoutValidators(t *testing.T) {
+	flagNetworkID = "sebak-test-network"
+	flagValidators = ""
+	flagKPSecretSeed = "SCN4NSV5SVHIZWUDJFT4Z5FFVHO3TFRTOIBQLHMNPAZJ37K5A2YFSCBM"
+	flagBindURL = "http://0.0.0.0:12345"
+	_, err := parseFlagValidators(flagValidators)
```
Charleslee522: When we see this test case, empty `--validators` is allowed.
It has same meaning with only "self". I figure it's different with our original intent.
User should understand what validator is. Therefore I think it would be better to output error if `--validators` argument is empty.
How do you think? :)

spikeekips: @Charleslee522 As you see, `flagValidators` is empty, so this test is for empty `--validators`. Am I miss something?

Charleslee522: @spikeekips I think `--validators "self"` is ok but `--validators`(empty) is not ok.
Empty validators seem like the user missed something.
How about you?

spikeekips: @Charleslee522 :) You are right, this test checks if `--validators` is empty, it will be error.

Charleslee522: Oh sorry, I missunderstood `require.NotNil(t, err)`...
Thank you for your kind explanation :)

cmd/sebak/common/util.go
```diff
@@ -23,6 +23,16 @@ func PrintFlagsError(cmd *cobra.Command, flagName string, err error) {
...
 
+func PrintError(cmd *cobra.Command, err error) {
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v\n\n", err)
+	}
+
+	cmd.Help()
+
+	os.Exit(1)
+}
```
Geod24: This is not used anywhere (and generally we should not `os.Exit` like this)

Geod24:
Wait, I don't see the motivation behind issue #503, or at least why it translates into this code.
What is the problem you saw ?

cmd/sebak/common/util.go
spikeekips: `PrintError` is used for https://github.com/bosnet/sebak/pull/511/files/efe9f927550b50897f013a48243b352ecf2bafd3#diff-f7927137f9b4e201f9f6cc347f0b0e1dR324 and this is another expression of `PrintFlagsError()`. Anyway, it should be removed because `node.NewLocalNode()` does not return `error` :)

spikeekips:
@Geod24 There is no problem without this PR. #503 is for adding node itself as it's validators regardless of setting `self`.

Geod24:
> @Geod24 There is no problem without this PR.

Then what's the goal of this PR ?

Issue #506, opened 2018-10-10
CORS support


zzim2x:
All requests from any browser should be allowed.

PR #514, opened 2018-10-12
Add CORS Middleware for client API

anarcher:
### Github Issue

Fixes: #506 

### Solution

- Using gorilla's CORS Middleware

### Possible Drawbacks

To more precise settings of CORS could be needed. 

- [X] Update go.mod
- [X] Add gorilla's CORS middleware to NodeRunner

### DIFF
--- a/lib/node/runner/node_runner.go
+++ b/lib/node/runner/node_runner.go
@@ -162,7 +162,7 @@ func (nr *NodeRunner) Ready() {
 	//CORS
 	{
 		allowedOrigins := ghandlers.AllowedOrigins([]string{"*"})
-		allowedMethods := ghandlers.AllowedMethods([]string{"GET", "HEAD", "POST", "PUT", "DELETE", "OPTIONS"})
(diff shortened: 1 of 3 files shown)


lib/node/runner/node_runner.go
```diff
@@ -158,6 +159,19 @@ func (nr *NodeRunner) Ready() {
 		return
 	}
 
+	//CORS
+	{
+		allowedOrigins := ghandlers.AllowedOrigins([]string{"*"})
+		allowedMethods := ghandlers.AllowedMethods([]string{"GET", "HEAD", "POST", "PUT", "DELETE", "OPTIONS"})
```
spikeekips: We should consider the available methods which are needed at this time. I think we can remove the methods except `GET` and `POST`.

anarcher: Oh Thanks! @spikeekips :->

anarcher:
@zzim2x l will merge it. If we have another issues about it, We will make  an issue or PR :-)

PR #515, opened 2018-10-12
Improve test code: Reduce globals, assume genesis block

Geod24:
A bunch of commit to improve the tests a little bit.
Also, I think it's reasonable to assume in some part of the code that a genesis block does exists.
I am currently working towards providing an interface that does not expose `error` when genesis block is requested, as it should be impossible to start consensus without one.


lib/node/runner/isaac_state_manager.go
```diff
@@ -45,7 +45,7 @@ func NewISAACStateManager(nr *NodeRunner, conf common.Config) *ISAACStateManager
 	genesisHeight := uint64(1)
 	genesisBlock, err := block.GetBlockByHeight(nr.storage, genesisHeight)
 	if err != nil {
-		nr.log.Error("cannot get genesis block from storage", "height", genesisHeight, "error", err)
+		panic(err)
```
spikeekips: 👍

lib/node/runner/test.go
```diff
@@ -20,16 +20,21 @@ import (
...
-	kp, _ = keypair.Random()
+	var err error
+	genesisKP, err = keypair.Random()
+	if err != nil {
+		panic(err)
+	}
+	commonKP, err = keypair.Random()
+	if err != nil {
+		panic(err)
+	}
```
spikeekips: 👍

lib/node/runner/test.go
```diff
@@ -160,14 +155,32 @@ func createNodeRunnerForTesting(n int, conf common.Config, recv chan struct{}) (
...
 	}
 	nr.isaacStateManager.blockTimeBuffer = 0
+	genesisBlock, _ := block.GetBlockByHeight(st, 1)
+	nr.Consensus().SetLatestBlock(genesisBlock)
 
 	return nr, nodes, connectionManager
 }
+
+// Initialize the blockchain with a genesis account and a common account
+func InitTestBlockchain() *storage.LevelDBBackend {
```
spikeekips: 👍

PR #518, opened 2018-10-12
Remove block_time_test

Geod24:
It's slowing down the tests a lot, as it waits for a full minute.
While the tests would usually run for a few seconds, this makes the process very slow.
This kind of test will be better done in the integration test.


Geod24:
@Charleslee522 : Yes I originally [asked for an integration test](https://github.com/bosnet/sebak/pull/426#issuecomment-422259128) ;)

Charleslee522:
@Geod24 #565 is merge :)

Issue #357, opened 2018-09-10
Throttling for API requests from clients and nodes


spikeekips:
We need to throttle the incoming requests from client and also from nodes.

* Throttling from clients, means that node can keep the transactions in `TransactionPool` to be managed constantly
* Throttling from other nodes, means to filter the requests from some problematic node.

We can find similar case from stellar horizon, https://github.com/stellar/go/blob/master/services/horizon/internal/middleware_rate_limit.go .

PR #525, opened 2018-10-14
Support Rate Limit For API

### LINKED ISSUE #25: Throttling incoming transactions from client
To prevent the network to be hanged out with massive incoming transaction from clients, node should throttle the incoming messages. For your information, the Stellar's horizon server uses https://github.com/PuerkitoBio/throttled .

spikeekips:
### Github Issue

resolves #25 #357 

### Background

Check #25 #357

### Solution

* Use github.com/ulule/limiter for rate limit

* added 2 option for rate-limit
    - `--rate-limit-api`
    - `--rate-limit-node`

* `--rate-limit-X` options examples,
    - `--rate-limit-api=10-S`: allow `10` requests per second
    - `--rate-limit-api=5-M`: allow `5` requests per minute
    - `--rate-limit-api=9-H`: allow `9` requests per hour

* By default,
    - `/api` has the rate limit, `100` requests per minute; `common.RateLimitAPI`
    - `/node` has the rate limit, `100` requests per second; `common.RateLimitNode`

* `--rate-limit-X` options support ip address, ex)
    - `--rate-limit-api="1.2.3.4=10-S"`: allow `10` requests per second from `1.2.3.4`

* In command line, multiple `--rate-limit-X` options can be set, ex)
    - `$ sebak node ... --rate-limit-api=10-S --rate-limit-api="1.2.3.4=1000-M"`: This allows `1000` requests per minute only from `1.2.3.4` and `10` requests per seconds from the others.

* If you set multiple options, the last rate limit will be applied, which does not have ip address. ex)
    - `$ sebak node ... --rate-limit-api=10-S --rate-limit-api=1000-M`: the last one, `1000-M` will be applied.

* Like other options, rate limit also can be set from shell environment variables
    - `--rate-limit-api` from `SEBAK_RATE_LIMIT_API`
    - `--rate-limit-node` from `SEBAK_RATE_LIMIT_NODE`

* Shell variables also can set the multple rate limit ex)
    - `SEBAK_RATE_LIMIT_API="10-S 1.2.3.4=1000-M"`
    - seperate values by blank space

* For unlimited limit, set the limit to `0`, ex)
    - `--rate-limit-api=0-S`

* Lowercase also possible, ex) `--rate-limit-api=10-s`, `--rate-limit-api=10-m`, `--rate-limit-api=10-h`

### DIFF
--- a/lib/network/http_middlewares.go
+++ b/lib/network/http_middlewares.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"runtime/debug"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/gorilla/mux"
@@ -91,6 +92,16 @@ func RateLimitMiddleware(logger logging.Logger, rule common.RateLimitRule) mux.M
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			ip := limiter.GetIPKey(r, true)
+
+			// from localhost, no rate limit
+			if strings.HasPrefix(ip, "127.0.") {
+				w.Header().Add("X-RateLimit-Limit", "")
+				w.Header().Add("X-RateLimit-Remaining", "")
+				w.Header().Add("X-RateLimit-Reset", "")
+
+				next.ServeHTTP(w, r)
+				return
+			}
 
 			var middleware *stdlib.Middleware
 
--- a/lib/network/http_middlewares.go
+++ b/lib/network/http_middlewares.go
@@ -51,6 +51,8 @@ func rateLimitErrorHandler(w http.ResponseWriter, r *http.Request, err error) {
 	httputils.WriteJSONError(w, errors.ErrorHTTPServerError.Clone().SetData("error", err))
 }
 
+// RateLimitMiddleware throttles the incoming requests; if `Limit` is 0, there
+// will be no limit.
 func RateLimitMiddleware(logger logging.Logger, rule common.RateLimitRule) mux.MiddlewareFunc {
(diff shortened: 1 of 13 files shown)


spikeekips:
As we expected, the expected error occurred,

```
++++ curl --insecure --request GET --header 'Accept: application/json' https://127.0.0.1:2821/api/v1/accounts/GDIRF4UWPACXPPI4GW7CMTACTCNDIKJEHZK44RITZB4TD3YUM6CCVNGJ
+++ echo '{"type":"https://boscoin.io/sebak/error/166","title":"too' many 'requests;' reached 'limit","status":429}'
```
:)

lib/network/http_middlewares.go
```diff
@@ -35,3 +43,102 @@ func RecoverMiddleware(logger log15.Logger) mux.MiddlewareFunc {
...
+				return
+			}
+
+			var middleware *stdlib.Middleware
+
+			if len(middlewaresByIP) < 1 {
+				middleware = defaultMiddleware
+			} else { // find middleware by ip
+				var found bool
+				if middleware, found = middlewaresByIP[ip]; !found {
```
anarcher: I have a question, Within this using, `middlwaresByIP` is thread-safty when multiple gorountes run for HTTP handling?

spikeekips: @anarcher `middlwaresByIP` will be created before running `http.Handler`, so reading `middlwaresByIP` will be safe in goroutine :)

anarcher: Oh, I got it. :+1:  If we don't update this map in running times, I agree that this map has thread safety :-)

PR #528, opened 2018-10-15
Add block.GetGenesis

Geod24:
Since in 51% of the cases error checking is ignored, and in the 48 other % it just panics, might as well make it its own function.

### DIFF
--- a/lib/block/block.go
+++ b/lib/block/block.go
@@ -17,8 +17,7 @@ import (
 )
 
 const (
-	maxBlockHeightStringLength int    = 20
-	EventBlockPrefix           string = "bk-saved"
+	EventBlockPrefix string = "bk-saved"
 )
 
 type Block struct {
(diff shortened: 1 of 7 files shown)


lib/block/block.go
```diff
@@ -168,8 +167,16 @@ func GetBlockKeyPrefixConfirmed(confirmed string) string {
...
 
 func GetBlockKeyPrefixHeight(height uint64) string {
-	f := fmt.Sprintf("%%s%%0%dd-", maxBlockHeightStringLength)
-	return fmt.Sprintf(f, common.BlockPrefixHeight, height)
+	return fmt.Sprintf("%s%020d", common.BlockPrefixHeight, height)
+}
+
+// Returns: Genesis block
+func GetGenesis(st *storage.LevelDBBackend) Block {
+	if blk, err := GetBlockByHeight(st, 1); err != nil {
```
spikeekips: Please use `common.GenesisBlockHeight`

Geod24: Right, fixed

PR #531, opened 2018-10-15
Move genesis/common block functionality into their own file

Geod24:
A simple refactoring / code move, we can also move it downwards in another package if the need arise, but at the moment it's better to just separate it this way. `MakeGenesisBlock` knows too much about the consensus rules, while `block` is one of (and should stay) the most basic packages we have.

### DIFF
--- a/lib/block/block.go
+++ b/lib/block/block.go
@@ -5,15 +5,12 @@ import (
 	"fmt"
 
 	"github.com/btcsuite/btcutil/base58"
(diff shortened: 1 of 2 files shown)


Geod24:
@spikeekips : I'm fine with that. Will do the change.

spikeekips:
@Geod24 It's not matter. It just suggestion. It's up to you :)

Geod24:
Did the change, will merge when the CI goes green.

PR #532, opened 2018-10-15
Remove the need to pass the testing object to helper functions

Geod24:
```
Those test helpers function are generating data to make testing easier.
In principle, they should never fail, and if they do, there's a big bad bug somewhere.
We should not be using the testing object for those kinds of checks, as they are not part of the tests.
Moreover, if that bbb (big bad bug!) happens in another goroutine, it will be ignored
with the testing object, bug cause a crash with panic.
```

### DIFF
--- a/lib/node/runner/isaac_simulation_test.go
+++ b/lib/node/runner/isaac_simulation_test.go
@@ -25,7 +25,7 @@ TestISAACSimulationProposer indicates the following:
 */
 func TestISAACSimulationProposer(t *testing.T) {
 	nr, nodes, _ := createNodeRunnerForTesting(5, common.NewConfig(), nil)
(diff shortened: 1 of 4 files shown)


spikeekips:
@Geod24 Can you introduce "bug cause a crash with panic" with error message?

Geod24:
@spikeekips : I'm not sure what you mean here ?

spikeekips:
> @Geod24 : if that bbb (big bad bug!) happens in another goroutine, it will be ignored
with the testing object, bug cause a crash with panic.

I think, the reason you submitted this PR is that you met some panic error, caused by `*testing.T` argument.

Geod24:
No I did not, I was refactoring some code in the same module and saw this, but there was no actual bug.

PR #539, opened 2018-10-16
Improve test code to be less DRY and fix some missing assignments

Geod24:
Better reviewed commit-by-commit


Geod24:
Updated to broaden the scope

PR #542, opened 2018-10-16
 Make GetLatestBlock panic on error + do more error checking

Geod24:
```
GetLatestBlock should never error out, as at least the GenesisBlock always exists
```

In addition, the last commit introduce the test-only `MustSave` function which allows to simplify test error checking. As can be seen from the diff, in many case, error checking was simply ignored. With `MustSave`, we safely provide a mean to do error checking at no cost in unittests.

### DIFF
--- a/lib/block/block.go
+++ b/lib/block/block.go
@@ -221,16 +221,15 @@ func GetBlockHeaderByHeight(st *storage.LevelDBBackend, height uint64) (bt Heade
 	return GetBlockHeader(st, hash)
 }
 
-func GetLatestBlock(st *storage.LevelDBBackend) (b Block, err error) {
+func GetLatestBlock(st *storage.LevelDBBackend) Block {
 	// get latest blocks
 	iterFunc, closeFunc := GetBlocksByConfirmed(st, storage.NewDefaultListOptions(true, nil, 1))
-	b, _, _ = iterFunc()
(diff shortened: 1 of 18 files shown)


lib/block/block.go
```diff
@@ -221,16 +221,15 @@ func GetBlockHeaderByHeight(st *storage.LevelDBBackend, height uint64) (bt Heade
...
 	// get latest blocks
 	iterFunc, closeFunc := GetBlocksByConfirmed(st, storage.NewDefaultListOptions(true, nil, 1))
-	b, _, _ = iterFunc()
+	b, _, _ := iterFunc()
 	closeFunc()
 
 	if b.Hash == "" {
-		err = errors.ErrorBlockNotFound
-		return
+		panic(errors.ErrorBlockNotFound)
```
spikeekips: 👍 If failed to get latest block, it should be panic!!!

lib/block/test.go
```diff
@@ -72,6 +72,34 @@ func InitTestBlockchain() *storage.LevelDBBackend {
 	return st
 }
 
+/// Version of `Block.Save` that panics on error, usable only in tests
+func (b *Block) MustSave(st *storage.LevelDBBackend) {
```
spikeekips: 💯

PR #546, opened 2018-10-16
support golang integration test

kfangw:
### Github Issue
#471 
What you need to do to add integration testcase is just
adding `xxx_test.go` in `./tests/client`


kfangw:
You can write an integration test case by using GoLang.
Please review.

tests/client/entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -xe
+
+if [ $# -eq 1 ]; then
```
Geod24: Does the go syntax support running multiple tests ? E.g. `./sdb_test -test.run test1 test2` ?

anarcher:
Wow! It’s amazing! :-)

tests/client/entrypoint.sh
kfangw: I did not test yet. But it supports regex.

kfangw: When I tested, It does not support multiple arguments.
But If you want to test multiple, then, using regexp will helpful
```
  -test.run regexp
    	run only tests and examples matching regexp
```

kfangw:
This PR is pretty obvious so I merge it without approvement.

PR #555, opened 2018-10-17
Fix unit test go1 11

kfangw:
### Github Issue
test cases in `tests` should not be included in  `test_go1_11` job


Geod24:
Would have been nice to wait for a review before self-merging

PR #563, opened 2018-10-18
Consensus rules cannot be broken in finishOperation

Geod24:
The consensus rules are checked in checker_ballot_transaction's `ValidateOp`.
So if this error happens, it means there's an important bug in sebak,
and that the consensus rules were bypassed.
As a result, we remove the checks which are not necessary,
and panic on operations that shouldn't return an error.

### DIFF
--- a/lib/node/runner/finish_ballot.go
+++ b/lib/node/runner/finish_ballot.go
@@ -195,42 +195,19 @@ func finishOperation(st *storage.LevelDBBackend, source string, op operation.Ope
 	}
 }
 
-func finishCreateAccount(st *storage.LevelDBBackend, source string, op operation.CreateAccount, log logging.Logger) (err error) {
-	if _, err = block.GetBlockAccount(st, source); err != nil {
-		err = errors.BlockAccountDoesNotExists
-		return
-	}
-
-	var baTarget *block.BlockAccount
-	if baTarget, err = block.GetBlockAccount(st, op.TargetAddress()); err == nil {
-		err = errors.BlockAccountAlreadyExists
-		return
-	} else {
-		err = nil
-	}
-
-	baTarget = block.NewBlockAccountLinked(
+func finishCreateAccount(st *storage.LevelDBBackend, source string, op operation.CreateAccount, log logging.Logger) error {
+	baTarget := block.NewBlockAccountLinked(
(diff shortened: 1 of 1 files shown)


spikeekips (commented):
At this time, when error occurred in `finish~` process the `panic` will be reasonable. @Charleslee522 If possible, when error or problem found in `finish~` process, how about go to Syncing?

lib/node/runner/checker.go
```diff
@@ -614,44 +614,29 @@ func finishOperation(st *storage.LevelDBBackend, source string, op operation.Ope
...
+func finishCreateAccount(st *storage.LevelDBBackend, source string, op operation.CreateAccount, log logging.Logger) error {
+	baTarget := block.NewBlockAccountLinked(
 		op.TargetAddress(),
 		op.GetAmount(),
 		op.Linked,
 	)
-	if err = baTarget.Save(st); err != nil {
-		return
+	if err := baTarget.Save(st); err != nil {
+		return err
```
spikeekips: Like `finishPayment`, this also just can make `panic`, is there any reason to return error?

Geod24: Since it writes to LevelDB, there are much more potential for failure, that's why it's treated differently.

kfangw: I agreed.
finishXXX should not return an error.
How about using `MustXXX` pattern?

spikeekips (commented):
Instead of panicking, we already use leveldb’s transaction, so

1. Discard when error
2. Sync process

How about this scenario?

spikeekips:
@Groed24 Hmm. Most error cases can be from SEBAK itself, but there are plenty cases, error occurr like,
- disk full
- db file corruption by other process
- leveldb’s own bug
- etc

Db fail -> process end, it seems unnatural. In most if failed cases, panic can be good, but disk full sebak should ensure, I think. Logging critic message will be enough.

anarcher:
I never ever use panic unless it's really *unrecoverable error* that the layer above should not be handling. It's an "abort the entire program" kind of error which you don't encounter that often.

My general rule: avoid as much as possible.

Geod24:
> but disk full sebak should ensure, I think

Yeah, that's why I only panic on read, not on write.

Issue #564, opened 2018-10-18
Post Transaction API for Client.


kfangw:
Now, the API uses node API handler.

replace the handler with a new one which can validate input and response its result.
The result may consist of errors or URLs for history to track progress of the tx by client

PR #574, opened 2018-10-22
Tx API for client

kfangw:
### Github Issue
#564 

### Background
- Transaction API for a client is needed

### Solution
- Refactoring(Move) api package https://github.com/bosnet/sebak/pull/574/commits/ace54a35eddf3a4cd086d02fb6a7c433c442fb59
- Use the message handler of node API and marshaling its request/response
- Transaction Post API response JSON+HAL or PROBLEM
- TransactionHistory API is revised and supported
- Client for this API


kfangw:
@spikeekips @anarcher Please review this PR

lib/node/runner/api_node.go
```diff
@@ -124,10 +125,10 @@ func (api NetworkHandlerNode) MessageHandler(w http.ResponseWriter, r *http.Requ
 	}
 
 	if err = common.RunChecker(checker, common.DefaultDeferFunc); err != nil {
-		if _, ok := err.(common.CheckerErrorStop); ok {
-			return
+		if len(checker.Transaction.H.Hash) > 0 {
+			block.SaveTransactionHistory(api.storage, checker.Transaction, body, block.BlockTransactionHistoryStatusRejected)
 		}
-		httputils.WriteJSONError(w, err)
+		http.Error(w, err.Error(), httputils.StatusCode(err))
```
kfangw: It response just like other node APIs
The response may be handled and wrapped with JSON if needed.

lib/node/runner/api/transaction_post.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func NewTeeReadCloser(origin io.ReadCloser, w io.Writer) io.ReadCloser {
+	return &TeeReadCloser{
+		ReadCloser: origin,
+		teeReader:  io.TeeReader(origin, w),
+	}
+}
+
+func (api NetworkHandlerAPI) PostTransactionsHandler(w http.ResponseWriter, r *http.Request, handler http.HandlerFunc) {
```
kfangw: It is the most important change in this PR.
It intercepts the input/output of the handler.
And execute the handler.
After that, wraps the output with additional information

lib/node/runner/api_node.go
anarcher: You mean that `err.Error()` just returns JSON string or plaintext string?
And response's header such as `Content-Type`? :->

anarcher:
What is a reason to move `network/api` to `node/runner/api`? :->

anarcher:
The code is great, But I'm not sure that makes tx history in persistent even it is broardcated.  How do we do any transactions  without block?  expired-and-remove? or retry-forever? :->

lib/node/runner/api/transaction_post.go
spikeekips: @kfangw One question; Did you intend to re-use the `runner.MessageHandler()` for `PostTransactionHandler()`? If so, how about implement copy-of `runner.MessageHandler()`. Code can be duplicated, but this intercept thing will be removed, code will be simpler. :)

kfangw:
@spikeekips 
1. 
At first, I wanted to copy the implementation of MassageHandler.
But, it is difficult to do that.
Hence, I detour to reuse the existing handler but take a copy of input/output to handle them for a client.

IMO, the GetXXX APIs and the PostTransaction API can reside in another server(like the horizon of stellar). When I think like this, the implementation that I uploaded is making sense.

2.
When a user want to check the transaction is confirmed or not,
there is 2 ways to do that
1. get transaction - only get the transaction is confirmed or not
2. get transaction history - get details of status for the transaction

This use case is written in the integration test `tests/client/account_test.go`
https://github.com/bosnet/sebak/pull/574/files#diff-3f43ab6a0d994d0b685c981331922189R244

kfangw:
I want to put the handler in ```NetworkHandlerAPI``` to not corrupt existing design.
To implement the checker in the handler, many of related code such as checker needed to be referenced or rewritten. 

It cause circular import, hence, I'd been tried refactoring to move the checker to separate package, but I gave up to do that.

The capturing implementation in the newly introduced API doesn't make sense?
I don't know the reason why we need to make duplicated code on purpose,  although we can solve the issue without the duplicated code

spikeekips:
@kfangw In some parts you said, I fully understand the some little annoying things can be made like circular imports and code duplication. We can choose 2 different way,

1. Make duplicated code for `PostTransaction()`,
2. or re-use the existing `MessageHandler()` with new 'Intercept' thing, which is only used for `PostTransaction()`

I choose the second one, 2. If we choose 2, the changes will be smaller than 1 :)

tests/client/congressvoting_test.go
```diff
@@ -41,8 +41,9 @@ func TestCongressVoting(t *testing.T) {
 		body, err := tx.Serialize()
 		require.Nil(t, err)
 
-		_, err = c.SubmitTransaction(body)
+		b, err := c.SubmitTransaction(body)
 		require.Nil(t, err)
+		t.Log(string(b))
```
Charleslee522: @kfangw It seems left over

lib/block/transaction_history.go
```diff
@@ -19,31 +19,44 @@ const (
...
 type BlockTransactionHistory struct {
 	Hash   string
 	Source string
 
-	Confirmed string
-	Created   string
-	Message   string
-
-	isSaved bool
+	Time    string
```
Charleslee522: 👍

lib/block/transaction_history.go
```diff
@@ -19,31 +19,44 @@ const (
 	BlockTransactionHistoryPrefixHash string = "bth-hash-" // bt-hash-<BlockTransactionHistory.Hash>
 )
 
-// TODO Is it correct to save raw `message` in BlockTransactionHistory?
-// TODO Do `BlockTransactionHistory` purge the old transactions? That is, it
-// just keep the recent transactions
+const (
+	BlockTransactionHistoryStatusSubmitted = "submitted"
```
Charleslee522: How about just the `TransactionStatusSubmitted`?

lib/node/runner/api_node.go
```diff
@@ -124,10 +125,10 @@ func (api NetworkHandlerNode) MessageHandler(w http.ResponseWriter, r *http.Requ
 	}
 
 	if err = common.RunChecker(checker, common.DefaultDeferFunc); err != nil {
-		if _, ok := err.(common.CheckerErrorStop); ok {
-			return
+		if len(checker.Transaction.H.Hash) > 0 {
+			block.SaveTransactionHistory(api.storage, checker.Transaction, body, block.BlockTransactionHistoryStatusRejected)
```
Charleslee522: 👍

lib/client/client.go
```diff
@@ -137,7 +137,7 @@ func (c *Client) LoadTransaction(id string, queries ...Q) (transaction Transacti
 	return
 }
 
-func (c *Client) LoadTransactionHistory(id string, queries ...Q) (transaction Transaction, err error) {
+func (c *Client) LoadTransactionHistory(id string, queries ...Q) (transaction TransactionHistory, err error) {
```
Charleslee522: @kfangw How about modifying the variable `transaction` to `txh` or `transactionHistory` also?

kfangw:
@spikeekips @Charleslee522 May I merge this?
Am I miss something needed to be changed?

spikeekips:
@kfangw I suggested another way in https://github.com/bosnet/sebak/pull/574#issuecomment-432527023 . Regardless my opinion, looks good :)

kfangw:
@spikeekips I miss understood ur comment. Sorry.
I thought that the second way means that I have done at this PR.

spikeekips:
> @kfangw : I thought that the second way means that I have done at this PR.

I meant, copy `MessageHandler()` to `PostTransaction()` with additional code :)

PR #576, opened 2018-10-22
BlockAccount SequenceID json key added

### LINKED ISSUE #575: CLI payment.go getSenderDetails function Error
When user makes payment transactions from same source, cli wallet can not make a transactions successful. Because cli wallet can not get the account's sequence id.

soonkuk:
### Github Issue
 `Resolves #575`

### Solution
add json key for the blockAccount sequence id.


soonkuk:
@Charleslee522 Ok I will make it.

PR #582, opened 2018-10-23
Make UnfreezingPeriod configurable, improve test and cli interface

Geod24:
Trying to write an integration test for freezing/unfreezing, I need to be able to configure the period so that we don't have to wait 2 weeks for the CI to complete 😄 
In addition, I saw that on error, the CLI was printing the string as `[]byte`.


cmd/sebak/cmd/run.go
```diff
@@ -318,6 +320,10 @@ func parseFlagsNode() {
 		threshold = int(tmpUint64)
 	}
 
+	if common.UnfreezingPeriod, err = strconv.ParseUint(flagThreshold, 10, 64); err != nil {
```
spikeekips: It should be ‘flagUnfreezingPeriod’

Geod24: Good eye!

Geod24:
Fixed

PR #583, opened 2018-10-23
HasTransaction should check block also

kfangw:
When a client submits a transaction, the node should make sure that the transaction is `not` in the transaction pool and block.


kfangw:
> If you also BlockTransactionHistory be useful, we will use it instead of BlockTransaction. This is just my personal opinion. Anyway 👍

I am working on PR #574 
The `BlockTransactionHistory` is handled there.

Geod24:
> When a client submits a transaction, the node should make sure that the transaction is not in the transaction pool and block.

Actually that should not be necessary. Because we have sequence ID, if someone submit a transaction, it cannot pass validation if it's already included in a block (because the sequence ID is bumped).

kfangw:
> > When a client submits a transaction, the node should make sure that the transaction is not in the transaction pool and block.
> 
> Actually that should not be necessary. Because we have sequence ID, if someone submit a transaction, it cannot pass validation if it's already included in a block (because the sequence ID is bumped).

This sound like `HasTransaction` is not necessary.
But, IMHO, by using the function before the validation, the node can be noticed earlier.
And the TransactionHistory needed to be stored before the validation.

Geod24:
> This sound like `HasTransaction` is not necessary.

It is useful for the pool, a priori (we could double check that)

> But, IMHO, by using the function before the validation, the node can be noticed earlier.

Except that when our blockchain grow, we'll be querying a DB which can be gigabytes for this check. On the other hand, the SequenceID check is extremely cheap (and appear quite soon afterwards).

> And the TransactionHistory needed to be stored before the validation.

Can you expand on that, I don't understand what you mean.

kfangw:
1. User submit transaction
2. Node validate syntax and check the tx is already known(in pool or in block)
3. Node store the history of the tx
4. ...

According to this flow, when user submit the same transaction which is already confirmed,
It is abnormal to create new transaction history for the tx at 3.

spikeekips:
@kfangw For details, In the second node receives tx in API and validates it and if error found, returns 400 with error message.

PR #584, opened 2018-10-23
explicit json struct tag

kfangw:
These days, many of us including ERMind confused that the key name of json data.

To overcome this, How about making a convention?
And write the json struct tag explicitly?

I propose snake_name_convention.


soonkuk:
LGTM!

PR #585, opened 2018-10-23
api update

kfangw:
When #574 is merged the API doc will be updated.


kfangw:
@anarcher @cmcm2222 please review this.

And..
@cmcm2222 Could you maintain the API documentation when I absent?

PR #587, opened 2018-10-23
Add MustWriteJSON for convenient

anarcher:
### Background

Most using of `httputils.WriteJSON` looks like below:
```
  if err := httputils.WriteJSON(w, 200, payload); err != nil {
        httputils.WriteJSONError(w, err)
    }
```

For convenient, Add `httputils.MustWriteJSON`

```
httputils.MustWriteJSON(w,200,payload)
```

Do it make sense? :->

### Possible Drawbacks

After this merges, I will change its used codes. :->

### DIFF
--- a/lib/network/api/transaction.go
+++ b/lib/network/api/transaction.go
@@ -6,13 +6,14 @@ import (
 
 	"github.com/gorilla/mux"
 
+	"strings"
+
 	"boscoin.io/sebak/lib/block"
 	"boscoin.io/sebak/lib/common/observer"
 	"boscoin.io/sebak/lib/error"
 	"boscoin.io/sebak/lib/network/api/resource"
 	"boscoin.io/sebak/lib/network/httputils"
 	"boscoin.io/sebak/lib/storage"
-	"strings"
 )
 
 func (api NetworkHandlerAPI) GetTransactionsHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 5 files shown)


Geod24:
> After this merges, I will change its used codes. :->

I'd prefer to see the change now, so it's obvious what impact it has.

anarcher:
@Geod24 it’s minor impact maybe. 
3 line to just 1 line. If you don’t like it, plz just close it without merging.  :->

Geod24:
No I'm generally in favor of those additions, however I don't like to add unused code.

anarcher:
@Geod24 I added used codes of it.

lib/network/api/transaction.go
```diff
@@ -56,10 +57,7 @@ func (api NetworkHandlerAPI) GetTransactionsHandler(w http.ResponseWriter, r *ht
 	prev := GetTransactionsHandlerPattern + "?" + options.SetReverse(true).Encode()
 	list := resource.NewResourceList(txs, self, next, prev)
 
-	if err := httputils.WriteJSON(w, 200, list); err != nil {
-		http.Error(w, "Error reading request body", http.StatusInternalServerError)
-		return
-	}
+	httputils.WriteJSON(w, 200, list)
```
Charleslee522: `MustWriteJSON`? or checking err nil?

anarcher: @Charleslee522 Let shows this pr desc.

Geod24: I think @Charleslee522 point was that you are using `WriteJSON` instead of `MustWriteJSON` here.

anarcher:
This PR is not a major issue and I am ok this PR is not approved. Thanks for review :-> I will close it.

Geod24:
@anarcher : Why do you close ? Except for the comment [here](https://github.com/bosnet/sebak/pull/587#discussion_r227639351), this LGTM.

Charleslee522:
> I think @Charleslee522 point was that you are using WriteJSON instead of MustWriteJSON here.

Yes. Exactly.
@anarcher I think this pr is good.

lib/network/api/transaction.go
anarcher: It depends on client-side, If you have more error handling, you can use `WriteJSON`, If you are not, just 
 use `MustWriteJSON` as notice as PR`desc.

Charleslee522: So then, I think using `MustWriteJSON` is right in this part. But not now :)

anarcher:
> I think @Charleslee522 point was that you are using WriteJSON instead of MustWriteJSON here.

Sorry, I misunderstood it. I will re-open this PR. thanks.

Issue #573, opened 2018-10-22
Too slow to `finishBallot` with lot of operations


spikeekips:
I am currently testing SEBAK with the stress-testing tool, name is `sebak-hot-body`(https://github.com/spikeekips/sebak-hot-body). I tested with `200` operations in each transaction and `1000` concurrent transactions. The result was terrible, most of transaction was failed to check the confirmation. The logs of sebak is here,
```
DBUG[10-22|12:44:34] payment done node=GCPQ.7DUR ballot=6iX.Vj1
DBUG[10-22|12:44:34] payment done node=GCPQ.7DUR ballot=6iX.Vj1
DBUG[10-22|12:44:34] payment done node=GCPQ.7DUR ballot=6iX.Vj1
...
DBUG[10-22|12:45:32] payment done node=GCPQ.7DUR ballot=6iX.Vj1
DBUG[10-22|12:45:32] payment done node=GCPQ.7DUR ballot=6iX.Vj1
DBUG[10-22|12:45:32] payment done node=GCPQ.7DUR ballot=6iX.Vj1
```
The `ACCEPT`ed ballot went into `finishBallot()` and done well. The only problem was it took so long time to finish, almost **1 minute**.

Considering several timeout of sebak and also timeout of hot-body side, we need to reduce the consumed time to `finishBallot()`.

Charleslee522:
Oh, I really want to see the result after modifying it :)

anarcher:
It's really slow? :->  1000 * 200 is huge operations :-)

spikeekips:
I am writing the patch for using `leveldb.Batch`. This is the latest result with `hot-body`.
```
$ cat /tmp/payment-done.log  | grep 6UBYV1U6BKNCi8x9SwFoMAGNXL8cwpjwKc9EJMt1RSmn  | wc -l
   10000
$ cat /tmp/payment-done.log  | grep 6UBYV1U6BKNCi8x9SwFoMAGNXL8cwpjwKc9EJMt1RSmn  | head -n 1
DBUG[10-23|16:08:25] payment done                             module=noderunner node=GCPQ.7DUR ballot=6UBYV1U6BKNCi8x9SwFoMAGNXL8cwpjwKc9EJMt1RSmn
$ cat /tmp/payment-done.log  | grep 6UBYV1U6BKNCi8x9SwFoMAGNXL8cwpjwKc9EJMt1RSmn  | tail -n 1
[...]

PR #599, opened 2018-10-25
Support `leveldb.Batch` in `LevelDBBackend`

spikeekips:
### Github Issue

related to #573 

### Background

Recently I found some slowness when `runner.finishBallot`. I assumed `leveldb.Transaction` is slower than `leveldb.Batch`.

### Solution

With `leveldb.Batch` there are some improvement for saving, especially `finishBalot()`. I wrote simple benchmark for `finishBallot`. This is the result,
```
$ time go test -timeout 1000h ./lib/node/runner... -v \
    -bench BenchmarkFinishBallotWith -run '^$' \
    | tee >(prettybench -no-passthrough)
goos: darwin
goarch: amd64
pkg: boscoin.io/sebak/lib/node/runner

===========================================================================+===========
 BenchmarkFinishBallotWithoutBatch_1_10-8     | 200  |      8703111 ns/op  |  
 BenchmarkFinishBallotWithBatch_1_10-8        | 200  |      8432652 ns/op  |  +3.1076%
----------------------------------------------+------+---------------------+-----------
 BenchmarkFinishBallotWithoutBatch_1_200-8    |  20  |     73501199 ns/op  |  
 BenchmarkFinishBallotWithBatch_1_200-8       |  50  |     36836526 ns/op  | +49.8830%
----------------------------------------------+------+---------------------+-----------
 BenchmarkFinishBallotWithoutBatch_1_400-8    |  20  |    111295178 ns/op  |  
 BenchmarkFinishBallotWithBatch_1_400-8       |  20  |     61944653 ns/op  | +44.3420%
----------------------------------------------+------+---------------------+-----------
 BenchmarkFinishBallotWithoutBatch_1_600-8    |  10  |    165450394 ns/op  |  
 BenchmarkFinishBallotWithBatch_1_600-8       |  20  |     94323325 ns/op  | +42.9899%
----------------------------------------------+------+---------------------+-----------
 BenchmarkFinishBallotWithoutBatch_1_800-8    |   5  |    216421041 ns/op  |  
 BenchmarkFinishBallotWithBatch_1_800-8       |  10  |    130150833 ns/op  | +39.8622%
----------------------------------------------+------+---------------------+-----------
 BenchmarkFinishBallotWithoutBatch_1_1000-8   |   5  |    267735515 ns/op  |  
 BenchmarkFinishBallotWithBatch_1_1000-8      |  10  |    144240689 ns/op  | +46.1256%
----------------------------------------------+------+---------------------+-----------
 BenchmarkFinishBallotWithoutBatch_200_10-8   |   2  |    718578099 ns/op  |  
 BenchmarkFinishBallotWithBatch_200_10-8      |   3  |    458517416 ns/op  | +36.1910%
----------------------------------------------+------+---------------------+-----------
 BenchmarkFinishBallotWithoutBatch_200_200-8  |   1  |  15948486347 ns/op  |  
 BenchmarkFinishBallotWithBatch_200_200-8     |   1  |   7036990911 ns/op  | +55.8767%
----------------------------------------------+------+---------------------+-----------
 BenchmarkFinishBallotWithoutBatch_200_400-8  |   1  |  36027407561 ns/op  |  
 BenchmarkFinishBallotWithBatch_200_400-8     |   1  |  15870237056 ns/op  | +55.9495%
----------------------------------------------+------+---------------------+-----------
 BenchmarkFinishBallotWithoutBatch_200_600-8  |   1  |  60989189227 ns/op  |  
 BenchmarkFinishBallotWithBatch_200_600-8     |   1  |  29518202778 ns/op  | +51.6009%
----------------------------------------------+------+---------------------+-----------
 BenchmarkFinishBallotWithoutBatch_200_800-8  |   1  |  75418588132 ns/op  |  
 BenchmarkFinishBallotWithBatch_200_800-8     |   1  |  39607720988 ns/op  | +47.4828%
----------------------------------------------+------+---------------------+-----------
 BenchmarkFinishBallotWithoutBatch_200_1000-8 |   1  | 101518575012 ns/op  |  
 BenchmarkFinishBallotWithBatch_200_1000-8    |   1  |  55013357793 ns/op  | +45.8095%
===========================================================================+===========


# Unknown spaces ㅜㅜㅜㅜㅜㅜ
=======================================================================================
 BenchmarkFinishBallotWithBatch_400_10-8      |   2  |    804385217 ns/op
 BenchmarkFinishBallotWithBatch_400_200-8     |   1  |  15026138629 ns/op
 BenchmarkFinishBallotWithBatch_400_400-8     |   1  |  37461945847 ns/op
 BenchmarkFinishBallotWithBatch_400_600-8     |   1  |  79445266393 ns/op
 BenchmarkFinishBallotWithBatch_400_800-8     |   1  | 127750238351 ns/op
 BenchmarkFinishBallotWithBatch_400_1000-8    |   1  | 190282134228 ns/op
 BenchmarkFinishBallotWithBatch_600_10-8      |   1  |   1641325406 ns/op
 BenchmarkFinishBallotWithBatch_600_200-8     |   1  |  34984065425 ns/op
 BenchmarkFinishBallotWithBatch_600_400-8     |   1  |  93445665848 ns/op
 BenchmarkFinishBallotWithBatch_600_600-8     |   1  | 163408571196 ns/op
 BenchmarkFinishBallotWithBatch_600_800-8     |   1  | 336742077297 ns/op
=======================================================================================
```
* `BenchmarkFinishBallotWithoutBatch_XXX` uses `leveldb.Transaction`
* `BenchmarkFinishBallotWithBatch_XXX` uses `leveldb.Batch`
* `BenchmarkFinishBallotWithBatch_600_800` means, `600` transactions and `800` operations in each transaction; this handles `240,000` operations in `finishBallot`

In most of cases the elapsed time reduced up to almost `45%` in `leveldb.Batch`.

> This benchmark test can be found at `lib/node/runner/finish_ballot_test.go`

The major changes are,

* new `LevelDBCore`, `BatchCore`; this almost acts like `leveldb.Transaction`
* `LevelDBBackend.OpenBatch()` returns new `LevelDBBackend` with `BatchCore`.
* batch `LevelDBBackend` handles data in memory(`leveldb.Batch`) and `Commit()` will actually write data in storage like `Transaction`
* `finishBallot` almost was not changed.

---
To make `finishBallot()` to be faster(#573), this PR will be helpful and the other way also be needed. At this time, the **`json.Marshal`** and **`json.Unmarshal`** seems to be so expensive. We need to find faster object serialization library to replace `json`.

### DIFF
--- a/lib/node/runner/finish_ballot_test.go
+++ b/lib/node/runner/finish_ballot_test.go
@@ -310,8 +310,8 @@ func TestBenchmarkFinishBallot(t *testing.T) {
 	var err error
 
 	err = testFinishBallotWithBatch(false, 100, 100)
-	require.Nil(t, err)
+	require.NoError(t, err)
 
 	err = testFinishBallotWithBatch(true, 100, 100)
-	require.Nil(t, err)
+	require.NoError(t, err)
 }
--- a/lib/node/runner/finish_ballot_test.go
+++ b/lib/node/runner/finish_ballot_test.go
@@ -1,0 +1,317 @@
+package runner
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"testing"
+
+	logging "github.com/inconshreveable/log15"
+	"github.com/stellar/go/keypair"
+	"github.com/stretchr/testify/require"
+
+	"boscoin.io/sebak/lib/ballot"
+	"boscoin.io/sebak/lib/block"
+	"boscoin.io/sebak/lib/common"
+	"boscoin.io/sebak/lib/consensus"
+	"boscoin.io/sebak/lib/consensus/round"
+	"boscoin.io/sebak/lib/network"
+	"boscoin.io/sebak/lib/node"
+	"boscoin.io/sebak/lib/storage"
+	"boscoin.io/sebak/lib/transaction"
+	"boscoin.io/sebak/lib/transaction/operation"
+)
+
+func createNodeRunnerForTestingWithFileStorage(n int, conf common.Config, recv chan struct{}) (*NodeRunner, []*node.LocalNode, string) {
+	var ns []*network.MemoryNetwork
+	var net *network.MemoryNetwork
+	var nodes []*node.LocalNode
+	for i := 0; i < n; i++ {
+		_, s, v := network.CreateMemoryNetwork(net)
+		net = s
+		ns = append(ns, s)
+		nodes = append(nodes, v)
+	}
+
+	for j := 0; j < n; j++ {
+		nodes[0].AddValidators(nodes[j].ConvertToValidator())
+	}
+
+	localNode := nodes[0]
+	policy, _ := consensus.NewDefaultVotingThresholdPolicy(67)
+
+	connectionManager := NewTestConnectionManager(
+		localNode,
+		ns[0],
+		policy,
+		recv,
+	)
+
+	st := &storage.LevelDBBackend{}
+	dir, err := ioutil.TempDir("", "sebak-test")
+	if err != nil {
+		panic(err)
+	}
+
+	{
+		var err error
+		config, _ := storage.NewConfigFromString(fmt.Sprintf("file://%s", dir))
+
+		if err = st.Init(config); err != nil {
+			panic(err)
+		}
+
+		block.MakeTestBlockchain(st)
+	}
+
+	is, _ := consensus.NewISAAC(networkID, localNode, policy, connectionManager, st, common.NewConfig(), nil)
+	is.SetProposerSelector(FixedSelector{localNode.Address()})
+
+	nr, err := NewNodeRunner(string(networkID), localNode, policy, ns[0], is, st, conf)
+	if err != nil {
+		panic(err)
+	}
+	nr.isaacStateManager.blockTimeBuffer = 0
+
+	return nr, nodes, dir
+}
+
+func testFinishBallotWithBatch(withBatch bool, numberOfTransactions, numberOfOperations int) error {
+	nr, localNodes, dir := createNodeRunnerForTestingWithFileStorage(1, common.NewConfig(), nil)
+	defer func() {
+		nr.Storage().Close()
+		os.RemoveAll(dir)
+	}()
+
+	proposerNode := localNodes[0]
+	nr.Consensus().SetProposerSelector(FixedSelector{proposerNode.Address()})
+
+	genesisBlock := block.GetGenesis(nr.Storage())
+	commonAccount, _ := GetCommonAccount(nr.Storage())
+	initialBalance, _ := GetGenesisBalance(nr.Storage())
+
+	var blt *ballot.Ballot
+	{
+		var txs []transaction.Transaction
+		var txHashes []string
+
+		rd := round.Round{
+			Number:      0,
+			BlockHeight: genesisBlock.Height,
+			BlockHash:   genesisBlock.Hash,
+			TotalTxs:    genesisBlock.TotalTxs,
+		}
+
+		for i := 0; i < numberOfTransactions; i++ {
+			kpA, _ := keypair.Random()
+			accountA := block.NewBlockAccount(kpA.Address(), common.Amount(common.BaseReserve))
+			accountA.MustSave(nr.Storage())
+
+			kpB, _ := keypair.Random()
+			tx := transaction.MakeTransactionCreateAccount(kpA, kpB.Address(), common.Amount(1))
+
+			var ops []operation.Operation
+			for j := 0; j < numberOfOperations-1; j++ {
+				kpC, _ := keypair.Random()
+
+				opb := operation.NewCreateAccount(kpC.Address(), common.Amount(1), "")
+				op := operation.Operation{
+					H: operation.Header{
+						Type: operation.TypeCreateAccount,
+					},
+					B: opb,
+				}
+				ops = append(ops, op)
+			}
+			tx.B.Operations = append(tx.B.Operations, ops...)
+			tx.B.SequenceID = accountA.SequenceID
+			tx.Sign(kpA, networkID)
+
+			txHashes = append(txHashes, tx.GetHash())
+			txs = append(txs, tx)
+			nr.TransactionPool.Add(tx)
+		}
+
+		blt = ballot.NewBallot(proposerNode.Address(), proposerNode.Address(), rd, txHashes)
+
+		opc, _ := ballot.NewCollectTxFeeFromBallot(*blt, commonAccount.Address, txs...)
+		opi, _ := ballot.NewInflationFromBallot(*blt, commonAccount.Address, initialBalance)
+		ptx, _ := ballot.NewProposerTransactionFromBallot(*blt, opc, opi)
+
+		blt.SetProposerTransaction(ptx)
+		blt.SetVote(ballot.StateINIT, ballot.VotingYES)
+		blt.Sign(proposerNode.Keypair(), networkID)
+	}
+
+	st := nr.Storage()
+	if withBatch {
+		st, _ = st.OpenBatch()
+	}
+
+	_, err := finishBallot(
+		st,
+		*blt,
+		nr.TransactionPool,
+		nr.Log(),
+		nr.Log(),
+	)
+
+	return err
+}
+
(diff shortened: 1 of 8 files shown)


lib/node/runner/finish_ballot_test.go
```diff
@@ -0,0 +1,317 @@
...
+
+	st := &storage.LevelDBBackend{}
+	dir, err := ioutil.TempDir("", "sebak-test")
+	if err != nil {
+		panic(err)
+	}
+
+	{
+		var err error
+		config, _ := storage.NewConfigFromString(fmt.Sprintf("file://%s", dir))
```
Geod24: We should really not be doing this in unittests. Can we move it to an integration test ?

lib/storage/batch.go
```diff
@@ -0,0 +1,130 @@
...
+// NewIterator does not work with `BatchBackend`
+func (bb *BatchCore) NewIterator(r *leveldbUtil.Range, opt *leveldbOpt.ReadOptions) leveldbIterator.Iterator {
+	return bb.core.NewIterator(r, opt)
+}
+
+func (bb *BatchCore) Put(key []byte, v []byte, opt *leveldbOpt.WriteOptions) error {
+	bb.Lock()
+	defer bb.Unlock()
+
+	bb.inserted[bb.convertKey(key)] = v
```
anarcher: I think that just put to inserted not `bb.batch.Put` and then `bb.Write`  do loop with `inserted` to `bb.batch.Put` and then batch writing. It can be made more faster and less usage memory in some case. 
(this case is  many put with a key)

anarcher:
After this PR is merged, I will change to use this batch with sync. thanks, @spikeekips

lib/storage/batch.go
spikeekips: @anarcher I also consider your approach, but `inserted` is `map` and 'leveldb.Batch' tries to keep the `Put`ting order(https://github.com/syndtr/goleveldb/blob/master/leveldb/batch.go#L88). I think it has reason. Though making `inserted` to be `[]interface{}`, `inserted` is used for checking and getting the `Put`ted content. So, this is the reason to drop the delaying `Put`.

lib/node/runner/finish_ballot_test.go
```diff
@@ -0,0 +1,317 @@
...
+
+func BenchmarkFinishBallotWithBatch_600_800(b *testing.B) {
+	benchmarkFinishBallotWithBatch(true, 600, 800, b)
+}
+
+func TestBenchmarkFinishBallot(t *testing.T) {
+	var err error
+
+	err = testFinishBallotWithBatch(false, 100, 100)
+	require.Nil(t, err)
```
Charleslee522: How about considering `require.NoError`? It's not mandatory :)

lib/node/runner/finish_ballot_test.go
```diff
@@ -0,0 +1,317 @@
...
+
+	err = testFinishBallotWithBatch(true, 100, 100)
+	require.Nil(t, err)
```
Charleslee522: Ditto :)

lib/storage/batch.go
```diff
@@ -0,0 +1,130 @@
...
+func (bb *BatchCore) convertKey(key []byte) string {
+	return string(key)
+}
+
+func (bb *BatchCore) Has(key []byte, opt *leveldbOpt.ReadOptions) (bool, error) {
+	bb.RLock()
+	defer bb.RUnlock()
+
+	var found bool
+	if _, found = bb.inserted[bb.convertKey(key)]; found {
```
Charleslee522: I figure just using `string(key)` is better to understand :)

lib/storage/storage.go
```diff
@@ -94,3 +94,8 @@ func ParseConfig(s string) (u *Config, err error) {
 
 	return
 }
+
+type CommittableBackend interface {
```
Charleslee522: What is the `Backend` mean?

lib/storage/batch.go
spikeekips: @Charleslee522 I see. I will.

lib/storage/storage.go
spikeekips: @Charleslee522 Oh, it should be fixed to 'CommittableCore', this means, `LevelDBCore` , which has `Commit` and `Discard`.

Geod24: Why not just `Commitable` ?

spikeekips: @Geod24 Ooops. My poor english :)

lib/storage/batch.go
anarcher: I don’t know the reason of batch index with application level. For replay batch, the batch index can be used. It means that if we can use replay batch. 
But l have no idea about the reason. Would you like to explain the reason? Thanks!

spikeekips: @anarcher I don't know exactly, but `leveldb.Batch` handles `Put`ed contents in slices. But one thing I think is this kind of case,

1. put A
3. delete A

The 'delete' action must be after 'put'.

lib/node/runner/finish_ballot_test.go
spikeekips: @Geod24 Hmmm, what is the problem of this? After testing, the temporary directory is removed.

lib/storage/batch.go
anarcher: > @anarcher I don't know exactly, but leveldb.Batch handles Puted contents in slices. But one thing I think is this kind of case,
> 1.  put A
> 2. delete A
> The 'delete' action must be after 'put'.

So, Last db status is `delete A` in this transaction. Is it right? 

`  Start a transaction     ->  put(A) -> put(A) -> delete(A)  ->  End a transaction (Flush/Write(batch) to LevelDB)`

Do we need to make  batch items like `put(A) (1)` and `put(A) (2)`  if we can get intermediate status of A from buffer (`inserted`) in a transaction ?

spikeekips: Yes, correct.

anarcher: Why do we need to make batch items of intermediate status of A
I'm not sure but we just make a batch.Delete(A) only,Is it not right?

spikeekips: This BatchCore also must act like LevelDBCore, Get returns the latest item. inserted is for this reason.

spikeekips: At this time, we only consider the leveldb, so we don’t need too much abstraction, but we will face to leave leveldb to another. At that time, we will apply your interfaces, I think.

lib/storage/batch.go
```diff
@@ -0,0 +1,126 @@
...
+	return nil
+}
+
+// Write will write the existing contents of `BatchBackend.batch` and then
+// argument, batch will be written.
+func (bb *BatchCore) Write(batch *leveldb.Batch, opt *leveldbOpt.WriteOptions) (err error) {
+	bb.Lock()
+	defer bb.Unlock()
+
+	if err = bb.core.Write(bb.batch, opt); err != nil {
```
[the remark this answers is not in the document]
spikeekips: > @kfangw : I also confused. When the batch argument is needed?

Yes, check https://godoc.org/github.com/syndtr/goleveldb/leveldb#DB.Write .
```
func (db *DB) Write(batch *Batch, wo *opt.WriteOptions) error
```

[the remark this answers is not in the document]
spikeekips: > @kfangw : 
> - Remove News, Sets method from `LevelDBBackend`

Yes, it could be, but I don't want to break the existing code :) We can tidy up in another issue.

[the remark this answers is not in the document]
kfangw: @spikeekips 

That is `DB`'s Write method.

If the role of the `BatchCore` is clear then it is fine.

Is the `BatchCore` DB? , Batch? or both?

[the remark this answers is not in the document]
spikeekips: The shape of `BatchCore` is almost same with `leveldb.Transaction` and `leveldb.DB`, it is one of `LevelDBCore`.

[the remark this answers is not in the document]
kfangw: Then, it is `DB` but wraps `Batch` also.
I understand your intention not to effects other code

lib/storage/storage.go
Geod24: Oops, this was not about `Commitable` vs `Committable` (the former is a typo, as @Charleslee522 realized!), but `CommittableCore` vs `Committable`.

PR #600, opened 2018-10-25
Slightly improve error code when writing to LevelDB

Geod24:
While debugging another issue, I came accross this error, which wasn't very helpful.
So here's the improved code.


anarcher:
Can you change the pkg directory from `lib/error` to `lib/errors`?   It is a good pkg name as same as dir name.

Geod24:
@anarcher : Good point! Added a commit to do that
@spikeekips : I'm not sure what you suggest here.

lib/storage/leveldb.go
```diff
@@ -183,7 +183,7 @@ func (st *LevelDBBackend) News(vs ...Item) (err error) {
 	for _, v := range vs {
 		if exists, err = st.Has(v.Key); exists || err != nil {
 			if exists {
-				err = errors.ErrorStorageRecordAlreadyExists
+				return errors.Newf("record {%v} already exists in storage", v.Key)
```
spikeekips: @Geod24 I mean, 

```
func Newf(code uint, format string, a ...interface{}) error {
...
}
```
```
return errors.Newf(ErrorStorageCoreError.Code, "record {%v} already exists in storage", v.Key)
```

anarcher:
Generally, The Sentinel errors are easy to use but have some disadvantages.  In place of the sentinel errors , I would like to use struct based errors like below:

```
type Error struct {
     Op Op   // Op string := pkg name / method name 
     Kind Kind
     Error Error
}

func Is(kind Kind,err error) bool { ... }

// Kinds
const (
     Other         Kind = iota // Unclassified error. This value is not printed in the error message.
     Invalid                   // Invalid operation for this type 
     ...
```

spikeekips:
@kfangw Yes, ‘Data’ is similar with ‘Details’ in HTTPProblem.

Geod24:
@spikeekips : Applied the change you requested

PR #601, opened 2018-10-25
Improve api/blocks test code

Geod24:
While working on frozen transaction, I realize most test code was breaking some of our invariants.
For example many places test code without a genesis block, which is a scenario we'll never have in a real-world scenario. Likewise, the blocks might not be correctly linked.
That makes any kind of change rather hard to do, as it ends up breaking seemingly unrelated places.
Here's a few improvement to test code which was actually breaking some of those invariants.

### DIFF
--- a/lib/node/runner/api/test.go
+++ b/lib/node/runner/api/test.go
@@ -36,8 +36,8 @@ func prepareAPIServer() (*httptest.Server, *storage.LevelDBBackend, error) {
 	return ts, storage, nil
 }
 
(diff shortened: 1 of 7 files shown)


lib/node/runner/api/test.go
```diff
@@ -52,14 +52,11 @@ func prepareOps(storage *storage.LevelDBBackend, blockHeight uint64, count int,
 
 	return kp, boList, nil
 }
-func prepareOpsWithoutSave(blockHeight uint64, count int, kp *keypair.Full) (*keypair.Full, []block.BlockOperation, error) {
+func prepareOpsWithoutSave(count int) (*keypair.Full, []block.BlockOperation, error) {
```
Charleslee522: 👍

PR #602, opened 2018-10-25
Improve TestBlockHeightOrdering

Geod24:
This test checks for an impossible scenario: blocks being saved out of order.
It's not something that should ever happen, and if anything, this test should not pass.

### DIFF
--- a/lib/block/block_test.go
+++ b/lib/block/block_test.go
@@ -70,34 +70,30 @@ func TestBlockConfirmedOrdering(t *testing.T) {
 }
 
 func TestBlockHeightOrdering(t *testing.T) {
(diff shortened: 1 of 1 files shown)


spikeekips:
@Geod24 This test is for checking `GetBlockByHeight()`, so 'ordering' means `GetBlockByHeight()` gets `Block` by `Block.Height`. The name of function maybe mislead the purpose.

Geod24:
@spikeekips : Following your comment, I updated the test to do just that.

spikeekips:
Test passes, I will approve.

Geod24:
Fixed the format issue

PR #604, opened 2018-10-25
Allow the additional tags in 'Content-Type' header like 'charset'

### LINKED ISSUE #590: client ( mobile wallet ) send request header's content type with `charset=utf-8` automatically.
As you can see in the attached image, client send to node `payment` request with `charset=utf-8` automatically. Therefore, payment operation can not get response through api handler but `http 404 error status code`.
contenttype

spikeekips:
### Github Issue
resolves #590 

### Background

Wallet client are suffering this problem, #590 

### Solution

`common.PostAndJSONMatcher()` patched.

### DIFF
--- a/lib/common/router.go
+++ b/lib/common/router.go
@@ -2,13 +2,23 @@ package common
 
 import (
 	"net/http"
+	"strings"
 
 	"github.com/gorilla/mux"
 )
 
 func PostAndJSONMatcher(r *http.Request, rm *mux.RouteMatch) bool {
 	if r.Method == "POST" {
-		if r.Header.Get("Content-Type") != "application/json" {
+		ct := r.Header.Get("Content-Type")
+		if len(strings.TrimSpace(ct)) < 1 {
+			return false
+		}
+		spl := strings.SplitN(ct, ";", 2)
+		if len(spl) == 2 {
+			ct = spl[0]
+		}
+
+		if strings.TrimSpace(ct) != "application/json" {
 			return false
 		}
 	}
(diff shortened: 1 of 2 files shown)


spikeekips:
This is urgent issue; please review. @anarcher @kfangw

spikeekips:
@kfangw :+1:

lib/common/router.go
```diff
@@ -2,13 +2,23 @@ package common
...
 
 func PostAndJSONMatcher(r *http.Request, rm *mux.RouteMatch) bool {
 	if r.Method == "POST" {
-		if r.Header.Get("Content-Type") != "application/json" {
+		ct := r.Header.Get("Content-Type")
+		if len(strings.TrimSpace(ct)) < 1 {
+			return false
+		}
+		spl := strings.SplitN(ct, ";", 2)
+		if len(spl) == 2 {
```
Geod24: `Content-Type` can have any number of parameters (https://www.w3.org/Protocols/rfc1341/4_Content-Type.html) so we should just take the first entry, always.

spikeekips: Yes, you correct.

kfangw:
It solves the problem but, I think the header always get only one `application/json`
It needs to be handled by the client API later.

spikeekips:
@Geod24 At first I used this func, but we just need the first part, so... :)

anarcher:
> @kfangw I understand what you pointed, but ‘PostTransactionHander’ is for client and client is widely open, sdk, web browser, curl, etc. ‘charset’ in ‘Content-Type’ is not illegal thing, so I think API should support this kind of variations.

Just I have a question: The `PostAndJSONMatcher` can be used with node api and client api too. Is it right?

spikeekips:
@anarcher Ues at this time.

Geod24:
> @Geod24 At first I used this func, but we just need the first part, so... :)

Yeah, so why splitting it into an array ? :)
I think if there's a standard function to do that, we should just use it, so this code is easier to read, even for people external to our team.

PR #608, opened 2018-10-25
Make prepareAPIServer start with a valid blockchain

Geod24:
@anarcher : That code was quite easy to read, thanks for that!


anarcher:
@Geod24  No, @kfangw did it. :->

PR #615, opened 2018-10-29
Saving `BlockOperation`s after `finishBallot()`

spikeekips:
### Background

This PR is only for performance. At this time, there are 2 points of bottleneck for elapsed time of one consensus,

* Ballot validation time
* Storing `block.Block`, `block.BlockTransaction` and `block.BlockOperation` in `finishBallot()`

In the huge amount of operations, these 2 processes become slow over timeout. If timeout occurs, the `EXP` ballot is produced and the transactions in `TransactionPool` does not proposed.

This PR is for the second problem of `finishBallot()`. This PR does not save `block.BlockOperation` in `finishBallot()`. `block.BlockOperation` will be saved after `finishBallot()`.

The detailed benchmark data can be found at https://gist.github.com/spikeekips/dfae9789dfdbc30414a5e58d5c2d42af . This result shows `40,000` operations(`200` tx and `200` operations in each tx) will be saved within 3 seconds. (Without this PR, it almost 7 seconds)

### Solution

Major parts are,

* `BlockOperation` will be saved in after `finishBallot()`
* If something wrong during saving `BlockOperation`, the `SavingBlockOperations` will check and save continuously in background.
* When node starts, `SavingBlockOperations` will check the existing `BlockOperation`s and save the missing `BlockOperation`s.


lib/block/block.go
```diff
@@ -103,6 +107,22 @@ func (b *Block) Save(st *storage.LevelDBBackend) (err error) {
...
 
+func (b Block) PreviousBlock(st *storage.LevelDBBackend) (blk Block, err error) {
+	if b.Height == common.GenesisBlockHeight {
+		err = errors.ErrorStorageRecordDoesNotExist
+		return
+	}
+
+	blk, err = GetBlockByHeight(st, b.Height-1)
+
+	return
```
Geod24: Without named return value, this function becomes much simpler

lib/block/block.go
```diff
@@ -103,6 +107,22 @@ func (b *Block) Save(st *storage.LevelDBBackend) (err error) {
...
+}
+
+func (b Block) NextBlock(st *storage.LevelDBBackend) (blk Block, err error) {
+	blk, err = GetBlockByHeight(st, b.Height+1)
+	return
```
Geod24: Same, use `return` directly

lib/block/block.go
```diff
@@ -39,6 +39,10 @@ func (bck Block) String() string {
 	return string(encoded)
 }
 
+func (bck Block) IsEmpty() bool {
+	return len(bck.Hash) < 1
```
Geod24: Wait, why do we use the length of the `Hash` ? I would expect it to be always present.

lib/block/genesis.go
```diff
@@ -123,5 +123,9 @@ func MakeGenesisBlock(st *storage.LevelDBBackend, genesisAccount BlockAccount, c
 		return
 	}
 
+	if err = bt.SaveBlockOperations(st, *blk); err != nil {
```
Geod24: `return` directly

lib/block/transaction.go
```diff
@@ -137,22 +137,7 @@ func (bt *BlockTransaction) Save(st *storage.LevelDBBackend) (err error) {
...
-			target := pop.TargetAddress()
-			if err = st.New(bt.NewBlockTransactionKeyByAccount(target), bt.Hash); err != nil {
-				return
-			}
-		}
-	}
+
 	event := "saved"
 	event += " " + fmt.Sprintf("source-%s", bt.Source)
 	event += " " + fmt.Sprintf("hash-%s", bt.Hash)
```
Geod24: Since we will now trigger the event before everything is saved, what's the strategy to deal with race condition ?

lib/block/block.go
spikeekips: I want to check the empty `block.Block` from `block.Block{}`. We can add another check points.

lib/block/transaction.go
Geod24: I think that should be handled right now. Safety trumps any performance concern.

spikeekips: @Geod24 This problem is not derived from this PR, it already exists after `Batch`. It would be better with another issue.

lib/block/block.go
Geod24: An empty `Block` is invalid, isn't it ? In this case I'd rename this method to `IsInvalid`.

lib/node/runner/block_operations.go
```diff
@@ -0,0 +1,236 @@
...
+	numWorker := int(len(blk.Transactions) / 2)
+	if numWorker > 100 {
+		numWorker = 100
+	}
+
+	for i := 1; i <= numWorker; i++ {
+		go sb.savingBlockOperationsWorker(i, st, blk, ops, results)
+	}
+	for _, txHash := range blk.Transactions {
+		ops <- txHash
```
Charleslee522: Does `ops` have `txHash`es? It is hard to understand. How about `txs`?

lib/node/runner/block_operations.go
```diff
@@ -0,0 +1,236 @@
...
+		sb.checkedBlock = blk.Height
+		checked = true
+	}
+
+	return
+}
+
+func (sb *SavingBlockOperations) savingBlockOperationsWorker(id int, st *storage.LevelDBBackend, blk block.Block, ops <-chan string, results chan<- error) {
+	for op := range ops {
+		results <- sb.CheckTransactionByBlock(st, blk, op)
```
Charleslee522: At first I saw this line, I expected that `CheckTransactionByBlock` returns some usable result,
but it returns error.
How about renaming `results` to `errors`?

lib/node/runner/block_operations.go
```diff
@@ -0,0 +1,236 @@
...
+	defer func() {
+		sb.log.Debug("end to save BlockOperation", "block", blk, "error", err)
+	}()
+
+	var st *storage.LevelDBBackend
+	if st, err = sb.st.OpenBatch(); err != nil {
+		return
+	}
+
+	if err = sb.CheckByBlock(st, blk); err != nil {
```
Charleslee522: Returned `err` from sb.CheckByBlock will be ignored in these two statements.
```
if err != nil {
   err = st.Discard()
} else {
   err = st.Commit()
}
```
Is this intended? But the returned error of `CheckByBlock` looks meaningful.

lib/node/runner/checker.go
```diff
@@ -594,6 +597,7 @@ func FinishedBallotStore(c common.Checker, args ...interface{}) (err error) {
 		}
 
 		checker.Log.Debug("ballot was stored", "block", *theBlock)
+		checker.NodeRunner.SavingBlockOperations().Save(*theBlock)
```
Charleslee522: I wonder when the `BlockOperation` is used.
Because in this pr, `BlockOperation`s are not stored immediately after the consensus is over.

Charleslee522: @spikeekips Exactly :)
Therefore, if some API(It's just my guess) wants to get `BlockOperation` after the consensus immediately, it might be not stored. Is it okay?

spikeekips: @Charleslee522 Yes, client will not get `BlockOperation` immediately after consensus, but `SavingBlockOperation` will catch up :)

Charleslee522: >  client will not get BlockOperation immediately after consensus

@spikeekips Thank you for your kind answer.
Is `client will not get` mean that client can get `BlockOperation` immediately, but don't?
or
client cannot get `BlockOperation` immediately?

spikeekips: @Charleslee522 The scenario is,

1. `finishBallot` finishes, but `BlockOperation` will be saved at this time
-> API will not return `BlockOpertion`

2. `finishBallot` calls `SavingBlockOperations().Save()`
-> API will not return `BlockOpertion`

3. after `finishBallot`, `SavingBlockOperations` will start to save `BlockOperation`
-> API will return `BlockOpertion`

Geod24:
Something is wrong with this PR: #654 
In my case, I'm not seeing the BlockOperations saved more than 20 blocks after.

spikeekips:
@Geod24 Hmm, I will check it. Thanks.

soonkuk:
@spikeekips @Geod24 I think @Geod24 's case is same with me. https://github.com/bosnet/sebak/issues/654.

PR #624, opened 2018-10-30
Implement delayed operations

Geod24:
```
With the introduction of delayed operations,
it is no more necessary to do an unfreezing after the unfreezing request have been done,
greatly simplifying the process for users.
```


lib/node/runner/checker.go
```diff
@@ -662,6 +661,11 @@ func finishBallot(st *storage.LevelDBBackend, b ballot.Ballot, transactionPool *
 		return nil, err
 	}
 
+	if err = finishDelayedOperations(blk, transactionPool.Pending, st); err != nil {
```
soonkuk: @Geod24 I think a delayed operation is not included in ballot.
It was inserted into pending pool when unfreezing request comes.
And it is executed when reached proper blockheight.
Of course, the unfreezing request was saved.
But should not the proposer add delayed operation to ballot again?

soonkuk: @Geod24 IMO, delayed operation is actual operation which affects balance of accounts.
And I think that it is also agreed through voting among the nodes.

Geod24: A delayed operation was agreed on at the time it entered the blockchain, so when it is actually applied, it does not need to be re-voted on.

Geod24:
Ping

lib/node/runner/checker.go
```diff
@@ -731,13 +736,56 @@ func FinishTransactions(blk block.Block, transactions []*transaction.Transaction
...
+//
+// Params:
+//   blk  = Pointer to the Block being finalized
+//   pool = The PendingPool to get operations from
+//   st   = Storage, used to look up the required blocks (and apply the changes)
+//
+// Returns:
+//   `nil` on success, in which case `pending` might have been mutated.
+//   If the return is non-`nil`, pending won't have been mutated.
+func finishDelayedOperations(blk *block.Block, pending *transaction.PendingPool, st *storage.LevelDBBackend) error {
```
Charleslee522: Great documentation 👍

lib/transaction/pending.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// Instantiate a new `PendingPool`
+func NewPendingPool() *PendingPool {
+	return &PendingPool{sync.RWMutex{}, list.New()}
+}
+
+// Insert a new element in the pool
+func (pp *PendingPool) Insert(height uint64, hash string) {
+	elem := &pendingPoolItem{height: height, txHash: hash}
```
Charleslee522: The second argument of `Insert()` is `opKey`, but parameter name is `hash` and it will be assigned to txHash.
`opKey` to `txHash`? I'm confused. Is there any reason?
As far as I know, `opKey` is "opHash-txHash"
Therefore, I figure just `opKey` is better.

lib/transaction/pending.go
```diff
@@ -0,0 +1,125 @@
...
+//            an empty string is returned.
+//   offset = In the event that this list contains more than one
+//            entry for a given height, this will skip `offset`
+//            entries. As such, it should be 0 on the first call.
+func (pp *PendingPool) Peek(height uint64, offset uint64) string {
+	pp.RLock()
+	defer pp.RUnlock()
+
+	nOffset := offset
+	for entry := pp.Front(); entry != nil; entry = entry.Next() {
```
Charleslee522: Is the `List` better for performance than `map`?
Doesn't it search front to target sequentially?

lib/transaction/pending.go
```diff
@@ -0,0 +1,125 @@
...
+// As a result a caller needs to do it manuall efter peeking
+// all the elements from the height.
+//
+// Params:
+//   height = The height being poped.
+func (pp *PendingPool) PopHeight(height uint64) {
+	pp.Lock()
+	defer pp.Unlock()
+
+	for entry := pp.Front(); entry != nil; entry = pp.Front() {
```
Charleslee522: Ditto

lib/node/runner/checker.go
```diff
@@ -731,13 +736,56 @@ func FinishTransactions(blk block.Block, transactions []*transaction.Transaction
...
+	var err error
+	offset := uint64(0)
+	for pendingH := pending.Peek(blk.Height, 0); pendingH != ""; pendingH = pending.Peek(blk.Height, offset) {
```
Charleslee522: I think, the `pendingHash` is better than `pendingH` because when I see this, I thought it was `pendingHeight`.

lib/transaction/pending.go
Geod24: Yes, that's the point. Elements are kept in order.

lib/node/runner/checker.go
Geod24: Good point

lib/transaction/pending.go
Charleslee522: >  Elements are kept in order.

@Geod24 Yes. But it is still O(n) because it does not binary search.

Geod24:
Rebased on master

Geod24:
Because @Charleslee522 [gave its LGTM](https://github.com/bosnet/sebak/pull/624#pullrequestreview-171856254)

zzim2x:
It was a code that did not work when I tested.

Geod24:
When did you test ? The PR was temporarily broken because master was (see #654 ).
What was your test ?

zzim2x:
not a unit test.

Geod24:
If you can find/reproduce a bug, please open an issue.

PR #620, opened 2018-10-30
Save Transaction before `finishBallot()`

spikeekips:
### Background
This is similar patch with #615 , performance concerns. To reduce the stored data size in `finishBallot()`, the `finishBallot()` will be short.

### Solution

This patch will save the incoming(or missing) transactions at when it is received. The all new transactions will be saved in `block.TransactionPool` and the existing `block.BlockTransaction` will store the transaction message(`block.BlockTransaction.Message`)

The major changes,

* New transaction receives, node will store it in `transaction.Pool` and `block.TransactionPool`
* `block.BlockTransaction` does not store the transaction message
* But `block.BlockTransaction.Message` field remains; the syncer will use it.

The benchmark shows the impressive result,

```
   1-10  :    0.0084 ->   0.0070 = +16.6587%
   1-200 :    0.0368 ->   0.0304 = +17.4210%
   1-400 :    0.0619 ->   0.0562 =  +9.1992%
   1-600 :    0.0943 ->   0.0797 = +15.4941%
   1-800 :    0.1302 ->   0.1043 = +19.8780%
   1-1000:    0.1442 ->   0.1324 =  +8.1985%
 200-10  :    0.4585 ->   0.3780 = +17.5610%
 200-200 :    7.0370 ->   6.6254 =  +5.8486%
 200-400 :   15.8702 ->  14.2655 = +10.1113%
 200-600 :   29.5182 ->  23.8177 = +19.3118%
 200-800 :   39.6077 ->  35.9260 =  +9.2954%
 200-1000:   55.0134 ->  52.8791 =  +3.8795%
 400-10  :    0.8044 ->   0.7926 =  +1.4646%
 400-200 :   15.0261 ->  14.1604 =  +5.7614%
 400-400 :   37.4619 ->  36.1464 =  +3.5116%
 400-600 :   79.4453 ->  68.6406 = +13.6001%
 400-800 :  127.7502 -> 112.0587 = +12.2830%
 400-1000:  190.2821 -> 171.1239 = +10.0683%
 600-10  :    1.6413 ->   1.3571 = +17.3160%
 600-200 :   34.9841 ->  25.5032 = +27.1006%
 600-400 :   93.4457 ->  71.3401 = +23.6560%
 600-600 :  163.4086 -> 144.0558 = +11.8432%
 600-800 :  336.7421 -> 257.9777 = +23.3901%
```
[...]

PR #635, opened 2018-11-03
Fixed to find transactions from `transaction.Pool` and also `block.TransactionPool`

spikeekips:
### Background

After #620 , the missing transactions of ballot will be saved in `block.TransactionPool`. After getting the missing transactions, the `INITBallotValidateTransactions` does get the transactions only from `transaction.Pool`, excluding `block.TransactionPool`

### Solution

* added new method, `BallotTransactionChecker.getTransaction()`
* for performance reason, `BallotTransactionChecker.transaction` will cache it.

### DIFF
--- a/lib/node/runner/checker.go
+++ b/lib/node/runner/checker.go
@@ -458,14 +458,14 @@ func INITBallotValidateTransactions(c common.Checker, args ...interface{}) (err 
 	}
 
 	transactionsChecker := &BallotTransactionChecker{
-		DefaultChecker: common.DefaultChecker{Funcs: INITBallotTransactionCheckerFuncs},
-		NodeRunner:     checker.NodeRunner,
-		LocalNode:      checker.LocalNode,
-		NetworkID:      checker.NetworkID,
-		Ballot:         checker.Ballot,
-		Transactions:   checker.Ballot.Transactions(),
-		VotingHole:     voting.NOTYET,
-		transactions:   map[string]transaction.Transaction{},
+		DefaultChecker:    common.DefaultChecker{Funcs: INITBallotTransactionCheckerFuncs},
+		NodeRunner:        checker.NodeRunner,
+		LocalNode:         checker.LocalNode,
+		NetworkID:         checker.NetworkID,
+		Ballot:            checker.Ballot,
+		Transactions:      checker.Ballot.Transactions(),
+		VotingHole:        voting.NOTYET,
+		transactionsCache: map[string]transaction.Transaction{},
 	}
 
 	err = common.RunChecker(transactionsChecker, common.DefaultDeferFunc)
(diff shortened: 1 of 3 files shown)


lib/node/runner/checker_ballot_transaction.go
```diff
@@ -25,6 +28,41 @@ type BallotTransactionChecker struct {
 	ValidTransactions     []string
 	validTransactionsMap  map[string]bool
 	CheckTransactionsOnly bool
+	transactions          map[string]transaction.Transaction
+}
+
+func (b *BallotTransactionChecker) getTransaction(hash string) (tx transaction.Transaction, found bool, err error) {
```
Charleslee522: This `getTransaction` method has two feature.
One is to insert transaction into `b.transactions` and another is getting transaction.
I figure, it would be better to split it to two methods(insertTransaction, getTransaction).

lib/node/runner/checker.go
```diff
@@ -465,6 +465,7 @@ func INITBallotValidateTransactions(c common.Checker, args ...interface{}) (err
 		Ballot:         checker.Ballot,
 		Transactions:   checker.Ballot.Transactions(),
 		VotingHole:     voting.NOTYET,
+		transactions:   map[string]transaction.Transaction{},
```
Charleslee522: There are two similar fields `Transactions` and `transactions` in `BallotTransactionChecker`.
`Transactions` is []string type and has hashes of transactions.
`transactions` is map[string]Transaction type and has transactions and hashes.
I don't know exactly but can we use only one field `transactions map[string]Transaction`?

spikeekips: @Charleslee522 `Transaction` is created by `Balot.Transactions()`, without real message. `transaction` will contain the real transaction message. How about renaming to `transactionCache`?

lib/node/runner/checker_ballot_transaction.go
spikeekips: @Charleslee522 `transaction` is just for cache :)

Charleslee522: > @Charleslee522 `transaction` is just for cache :)

@spikeekips Okay, we can split it later :)

lib/node/runner/checker.go
Charleslee522: @spikeekips Alright, the `transactionCache` is good.

lib/node/runner/checker_ballot_transaction.go
```diff
@@ -25,6 +28,41 @@ type BallotTransactionChecker struct {
...
+	b.RUnlock()
+
+	if found {
+		return
+	}
+
+	b.Lock()
+	defer b.Unlock()
+
+	tx, found = b.NodeRunner.TransactionPool.Get(hash)
```
Charleslee522: @spikeekips I have a question about this closed PR.
`b.transactionsCache` is superset of `runner.TransactionPool` and `block.TranscationPool`, right?
Then some tx can be in both. Isn't it a waste of memory?

spikeekips: @Charleslee522 Yes, it will be, but `transactionCache` will be removed after checking. It will help `TransactionPool` to be clean.

Charleslee522: @spikeekips Thanks. Clear!

PR #638, opened 2018-11-05
Fix go.mod diff when building

Geod24:


Geod24:
Self merging since it's trivial.

PR #641, opened 2018-11-05
Stop broadcasting the incoming transaction in `/node/message`

### LINKED ISSUE #639: Broadcating tx with client api only
Currently, Both of client API and node API tx handlers are broadcasting txs to all validators. 
It makes to increase tx receiving and handling. (Some nodes get many same tx * N(num of nodes) ).

I think that client API tx handler broadcasts to all nodes, but node API tx handler doesn't.

spikeekips:
### Github Issue
Resolves #639

### Background

See #639 

### Solution

* The existing `api.PostTransaction()` and `runner.MessageHandler()` will use `NetworkHandlerNode.ReceiveTransaction()`
* `NetworkHandlerNode.ReceiveTransaction()` takes `[]common.CheckerFunc` as argument
* `api.PostTransaction()` will use `HandleTransactionCheckerFuncs`(broadcasting version)
* `api.PostTransaction()` will use `HandleTransactionCheckerFuncsWithoutBroadcast`(not broadcasting version)

### DIFF
--- a/lib/node/runner/api/transaction_post.go
+++ b/lib/node/runner/api/transaction_post.go
@@ -1,15 +1,14 @@
 package api
 
 import (
-	"boscoin.io/sebak/lib/errors"
+	"io"
+	"io/ioutil"
+	"net/http"
+
+	"boscoin.io/sebak/lib/common"
 	"boscoin.io/sebak/lib/network/httputils"
 	"boscoin.io/sebak/lib/node/runner/api/resource"
 	"boscoin.io/sebak/lib/transaction"
-	"bufio"
-	"bytes"
-	"encoding/json"
-	"io"
-	"net/http"
 )
 
 type TeeReadCloser struct {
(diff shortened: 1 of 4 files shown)


lib/node/runner/api/transaction_post.go
```diff
@@ -1,15 +1,14 @@
...
 	"boscoin.io/sebak/lib/node/runner/api/resource"
 	"boscoin.io/sebak/lib/transaction"
-	"bufio"
-	"bytes"
-	"encoding/json"
-	"io"
-	"net/http"
 )
 
 type TeeReadCloser struct {
```
anarcher: `TeeReadClose` is needed?

spikeekips: No, we don’t and also InterceptXXX too. I just left these for future :)

anarcher:
IMHO,  It is better to use tx checkfuncs in the API handlers directly. It makes a room for expending functionality for client API own handling...

spikeekips:
> @anarcher : IMHO, It is better to use tx checkfuncs in the API handlers directly. It makes a room for expending functionality for client API own handling...

Please show some great hint or pseudo code :)

anarcher:
> Please show some great hint or pseudo code :)

I have already approved this PR. :-) . I wonder `PoolLimiter` for client api post tx handeler . I think that it can do with this PR's approach.

lib/node/runner/api_node.go
```diff
@@ -124,10 +125,29 @@ func (api NetworkHandlerNode) MessageHandler(w http.ResponseWriter, r *http.Requ
...
+
+	if err != nil {
 		if len(checker.Transaction.H.Hash) > 0 {
 			block.SaveTransactionHistory(api.storage, checker.Transaction, block.TransactionHistoryStatusRejected)
 		}
+		return
+	}
+
+	tx = checker.Transaction
+	return
```
Charleslee522: How about returning with variable like `return checker.Transaction, nil`?
It looks more clear.

Issue #640, opened 2018-11-05
Found block height is behind


spikeekips:
Recently I found that some node has behind `Block.Height` during hot body test. I ran hot body for 5 minutes and each node has block height like this,

```
$ for i in *.log; do echo -n $i ": "; cat $i | grep 'finish_ballot.go:54' | jq '.height' | tail -n 1; done
GAFL.DZOX.log : 29
GARD.SFM6.log : 15
GCTE.TCV6.log : 29
GCZR.RDVF.log : 29
```

I guess if some nodes lag behind block height, `syncer` will follow up to the latest, but it's not. The detailed node logs can be found at https://gist.github.com/spikeekips/701624a7c3662ca8282db7f9a0ca0363 .

anarcher:
According to `GARD.SFM6.log`,  The fetcher did try to fatch but failted downloading via block api. 
I think this is a reason that GARD.SFM6 doesn't have the same height. And fetcher tries to fetch with `172.31.22.130` and `172.31.22.219` too. 

```
{"caller":"fetcher.go:114","fetching_node":{"address":"GCTEOACB236P6R7ZSXOF2PSWWK7X6KRDSCQOP7WMOEZVFFZMTCV6U7O2","alias":"GCTE.TCV6","en
dpoint":"https://172.31.22.130:12001"},"height":17,"lvl":"dbug","module":"sync","msg":"fetching items from node: GCTE.TCV6","node":"GARD
.SFM6","t":"2018-11-04T13:31:02.966649829Z"}
{"caller":"fetcher.go:117","lvl":"dbug","module":"sync","msg":"apiClient","node":"GARD.SFM6","t":"2018-11-04T13:31:02.966704592Z","url":
"https://172.31.22.130:12001/node/blocks?height-range=17-18\u0026mode=full"}
{"caller":"fetcher.go:139","height":17,"items":2,"lvl":"dbug","module":"sync","msg":"fetch get items","node":"GARD.SFM6","t":"2018-11-04
T13:31:03.13806425Z"}
{"caller":"fetcher.go:87","err":"unexpected end of JSON input","lvl":"eror","module":"sync","msg":"unexpected end of JSON input","node":
"GARD.SFM6","t":"2018-11-04T13:31:03.138260788Z"}
```

```
$cat GARD.SFM6.log | grep sync | grep apiClient | grep 17-18
[...]

PR #642, opened 2018-11-05
Add more logging and wrap errors

anarcher:
### Github Issue

Related  #640 

### Background

For debugging, Add more log messages and warping errors using `github.com/pkg/errors`.

### DIFF
--- a/lib/errors/errors.go
+++ b/lib/errors/errors.go
@@ -3,9 +3,13 @@ package errors
 import (
 	"errors"
 	"fmt"
+
+	pkgerrors "github.com/pkg/errors"
 )
 
 var New = errors.New
+var Wrap = pkgerrors.Wrap
+var Wrapf = pkgerrors.Wrapf
 
 func Newf(err *Error, format string, args ...interface{}) error {
 	return NewError(err.Code, fmt.Sprintf(format, args...))
@@ -90,4 +94,5 @@ var (
 	TransactionSameSourceInPool               = NewError(176, "Same transaction source found in pool")
 	AlreadyCommittable                        = NewError(177, "already Committable")
 	FailedToSaveBlockOperaton                 = NewError(178, "failed to save BlockOperation")
+	NodeNotFound                              = NewError(179, "Node not found")
 )
--- a/lib/sync/fetcher.go
+++ b/lib/sync/fetcher.go
@@ -23,7 +23,6 @@ import (
 	"boscoin.io/sebak/lib/transaction"
 
 	"github.com/inconshreveable/log15"
-	pkgerrors "github.com/pkg/errors"
 )
 
 type BlockFetcher struct {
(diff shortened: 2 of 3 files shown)


lib/sync/fetcher.go
```diff
@@ -84,7 +86,7 @@ func (f *BlockFetcher) Fetch(ctx context.Context, syncInfo *SyncInfo) (*SyncInfo
 					return false, ctx.Err()
 				}
 
-				f.logger.Error(err.Error(), "err", err)
+				f.logger.Error(fmt.Sprintf("fetch err: %v", err), "err", err, "height", height)
```
spikeekips: `"error"` already exists, so instead `fmt.Sprintf("fetch err: %v", err),` `"fetch err"`will be enough.

lib/sync/fetcher.go
```diff
@@ -109,7 +111,7 @@ func (f *BlockFetcher) fetch(ctx context.Context, si *SyncInfo) error {
 
 	n := f.pickRandomNode(nodeAddrs)
 	if n == nil {
-		return errors.New("Fetch: node not found")
+		return errors.New("fetch: node not found")
```
spikeekips: How about adding new error in `boscoin.io/sebak/lib/errors`

lib/sync/fetcher.go
```diff
@@ -172,7 +184,7 @@ func (f *BlockFetcher) fetch(ctx context.Context, si *SyncInfo) error {
 		tx, ok := txmap[hash]
 		if !ok {
 			//TODO(anarcher): Error type for controlling timeout
-			err := fmt.Errorf("Tx: %s not found in block height %d", hash, height)
+			err := fmt.Errorf("tx: %s not found in block height %d", hash, height)
```
spikeekips: We already have `errors.TransactionNotFound`.

lib/sync/fetcher.go
```diff
@@ -22,6 +23,7 @@ import (
 	"boscoin.io/sebak/lib/transaction"
 
 	"github.com/inconshreveable/log15"
+	pkgerrors "github.com/pkg/errors"
```
Geod24: I think it would be better to import this in sebak's `error` package and expose `Wrap` from there.

lib/sync/fetcher.go
anarcher: Ok,

lib/sync/fetcher.go
anarcher: Ok,

anarcher:
The Sentinel errors are easy to use but it has disadvantages. I like another approach like below:

```
  type Error struct {     
      Op  Op // unually `pkg/object.Method` e.g. `file/server.Lookup`
      Kind Kind
      Err error
  }
 func Is(kind Kind, err error) bool
```

lib/errors/errors.go
```diff
@@ -88,4 +92,5 @@ var (
 	BlockTransactionHistoryDoesNotExists      = NewError(174, "transaction history does not exists in block")
 	NotCommittable                            = NewError(175, "not Committable")
 	TransactionSameSourceInPool               = NewError(176, "Same transaction source found in pool")
+	NodeNotFound                              = NewError(177, "Node not found")
```
Geod24: Shouldn't be uppercase

lib/sync/fetcher.go
```diff
@@ -133,6 +135,16 @@ func (f *BlockFetcher) fetch(ctx context.Context, si *SyncInfo) error {
 
 	items, err := f.unmarshalResp(resp.Body)
 	if err != nil {
+		err := errors.Wrap(err, "resp unmarshal")
```
Geod24: Don't use abbreviations in error message. "reponse failed to unmarshal" is much more explanatory.

lib/sync/fetcher.go
```diff
@@ -133,6 +135,16 @@ func (f *BlockFetcher) fetch(ctx context.Context, si *SyncInfo) error {
 
 	items, err := f.unmarshalResp(resp.Body)
 	if err != nil {
+		err := errors.Wrap(err, "reponse failed to unmarshal")
+
+		body := func() string {
+			body, readErr := ioutil.ReadAll(resp.Body)
```
Geod24: Won't we always get `EOF` here if the reader has been consumed in `unmarshalResp` ?

anarcher: I did use a buffer for it.  But I have a mind to remove body logging.. too.  :-|

PR #644, opened 2018-11-05
Remove `networkID` from operation.IsWellFormed, improve test code

Geod24:
Better reviewed commit-by-commit.


Geod24:
>  If config.Config can have NetworkID, how about remove networkID argument in Transaction and Ballot?

That's a good point. However the changes to do are quite big. Do you mind if I do it in another PR ?

spikeekips:
@Geod24 I will.

PR #645, opened 2018-11-05
Surpass error when genesis block is already created

spikeekips:
### Background

This is urgent request from @zzim2x :) Current `$ sebak node --genesis` will start with creating genesis block and it's accounts. If it is already created, `--genesis` option will stop to run.

### Solution
This patch will check,
* genesis account address is same
* common account address is same
* balance is same
* networkID is same

If everything is same, `--genesis` option will not make error.

### DIFF
--- a/cmd/sebak/cmd/genesis.go
+++ b/cmd/sebak/cmd/genesis.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -13,6 +12,8 @@ import (
 	"boscoin.io/sebak/lib/block"
 	"boscoin.io/sebak/lib/common"
 	"boscoin.io/sebak/lib/common/keypair"
+	"boscoin.io/sebak/lib/errors"
+	"boscoin.io/sebak/lib/node/runner"
 	"boscoin.io/sebak/lib/storage"
 )
 
@@ -122,6 +123,30 @@ func makeGenesisBlock(genesisAddress, commonAddress, networkID, balanceStr, stor
 	}
 	defer st.Close()
 
+	created, err := checkExistingAccounts(st, flagNetworkID, genesisKP.Address(), commonKP.Address(), balance)
+	if err != nil {
+		if created {
+			return "--storage", fmt.Errorf("genesis block is already created, but: %v", err)
+		} else {
+			return "--storage", err
+		}
+	} else if created {
+		if b, err := block.GetBlockByHeight(st, common.GenesisBlockHeight); err != nil {
+			return "--storage", fmt.Errorf("failed to get genesis block: %v", err)
+		} else {
+			log.Info("genesis block already created",
+				"height", b.Height,
+				"round", b.Round,
+				"timestamp", b.Timestamp,
+				"total-txs", b.TotalTxs,
+				"total-ops", b.TotalOps,
+				"proposer", b.Proposer,
+			)
+		}
+
+		return "", nil
+	}
+
 	// check account does not exists
 	if _, err = block.GetBlockAccount(st, genesisKP.Address()); err == nil {
 		return "<public key>", errors.New("account is already created")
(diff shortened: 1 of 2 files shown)


Geod24:
> This is urgent request from @zzim2x :)

What was the problem ? We have to ways to run the nodes: `sebak genesis + sebak run`, or `sebak run --genesis`. The first version is for long running nodes, the second is preferred to start a node from scratch. This patch changes the second version to behave like the first one, but that's not exactly what we want. And if we do, we should just get rid of `sebak genesis` altogether...

spikeekips:
> @Geod24 : What was the problem ? We have to ways to run the nodes: `sebak genesis + sebak run`, or `sebak run --genesis`. The first version is for long running nodes, the second is preferred to start a node from scratch. This patch changes the second version to behave like the first one, but that's not exactly what we want. And if we do, we should just get rid of `sebak genesis` altogether...

I heard, when deploying by kubernetes, there are some difficulties with `--genesis` option. If the same genesis block already exists, it will be easily deployed.

zzim2x:
Actually, it is not urgent and not difficulty. !!!! Testnet nodes already run `sebak genesis + sebak node`.

1. `./entrypoint.sh` in Dockerfile is useless in production. It fails when an application is upgraded or restarted on same disk storage. That's because genesis block already created. I can fix it by creating a new `./entrypoint.sh` including a command to ignore the `sebak genesis` command. 
2. There is no way to identify that it is the same environment such as network-id, genesis-block, etc.

It does not need to be run if it is different from the expected environment.

Geod24:
I never saw `entrypoint.sh` as a production thing, because it requires additional disk setup which is simply not there. But if that's what you want to pursue, then we should simply remove the `genesis` command.

For point 2: We could also allow a node to fetch the genesis block

spikeekips:
@Geod24 This PR can make to avoid extra code like checking existence of accounts of genesis block. The checking func of this PR can check the details of genesis block. This will be useful.

cmd/sebak/cmd/genesis.go
```diff
@@ -122,6 +123,30 @@ func makeGenesisBlock(genesisAddress, commonAddress, networkID, balanceStr, stor
...
 	defer st.Close()
 
+	created, err := checkExistingAccounts(st, flagNetworkID, genesisKP.Address(), commonKP.Address(), balance)
+	if err != nil {
+		if created {
+			return "--storage", fmt.Errorf("genesis block is already created, but: %v", err)
+		} else {
+			return "--storage", err
+		}
+	} else if created {
```
Charleslee522: I think `if created` not `else if` is easy to understand. It's not mandatory :)

lib/node/runner/util.go
```diff
@@ -10,8 +10,12 @@ import (
 	"boscoin.io/sebak/lib/version"
 )
 
-func getGenesisTransaction(st *storage.LevelDBBackend) (bt block.BlockTransaction, err error) {
-	bk := block.GetGenesis(st)
+func GetGenesisTransaction(st *storage.LevelDBBackend) (bt block.BlockTransaction, err error) {
+	var bk block.Block
+	if bk, err = block.GetBlockByHeight(st, common.GenesisBlockHeight); err != nil {
```
Charleslee522: 👍

cmd/sebak/cmd/genesis.go
```diff
@@ -153,3 +178,63 @@ func makeGenesisBlock(genesisAddress, commonAddress, networkID, balanceStr, stor
...
 }
+
+func checkExistingAccounts(st *storage.LevelDBBackend, networkID, genesisAddress, commonAddress string, balance common.Amount) (created bool, err error) {
+	// check network id
+	var bt block.BlockTransaction
+	if bt, err = runner.GetGenesisTransaction(st); err != nil {
+		if err == errors.StorageRecordDoesNotExist {
+			created = false
+			err = nil
+			return
```
Charleslee522: I think `return false, nil` is better.

cmd/sebak/cmd/genesis.go
spikeekips: `created` should be checked, because without error created can be `false`, so without error and `created=false`, creating genesis should be run.

cmd/sebak/cmd/genesis.go
spikeekips: @Charleslee522 I am still in love with named return :)

cmd/sebak/cmd/genesis.go
Charleslee522: @spikeekips I mean,
```
if err != nil {
    if created {
        return "--storage", fmt.Errorf("genesis block is already created, but: %v", err)
    } else {
        return "--storage", err
    }
}

if created {
    if b, err := block.GetBlockByHeight(st, common.GenesisBlockHeight); err != nil {
        return "--storage", fmt.Errorf("failed to get genesis block: %v", err)
    } else {
        log.Info("genesis block already created",
            "height", b.Height,
            "round", b.Round,
            "timestamp", b.Timestamp,
            "total-txs", b.TotalTxs,
            "total-ops", b.TotalOps,
            "proposer", b.Proposer,
        )
    }
    return "", nil
}
```

cmd/sebak/cmd/genesis.go
Charleslee522: This `created = false` code is redundant because created is already false.
Named return is okay :)

PR #653, opened 2018-11-06
Check 'fee' by operation type

### LINKED ISSUE #381: Implementation sending transaction fee according to operation type.
## Background
The transaction fee is sent to the specific account in general.
But for some specific transaction which has operations like as voting result, membership, issuance the transaction fee may be exempted.

## Solution
Include sending tx fee in FinishOperation function of each operation type basically.
According to operation type some FinishOperation functions can omit sending tx fee.

spikeekips:
### Github Issue
resolves #381 

### Background

At this time, the zero operation still counted in validation of Transaction fee.

### Solution

* `Operation.HasFee()` added; it returns the operation has fee or not
* `Transaction.TotalBaseFee()` counts only  `Operation.HasFee() == true`
* test cases added in `lib/transaction/transaction_test.go`

### DIFF
--- a/lib/ballot/proposer_transaction.go
+++ b/lib/ballot/proposer_transaction.go
@@ -162,9 +162,6 @@ func (p ProposerTransaction) IsWellFormedWithBallot(networkID []byte, blt Ballot
 				err = errors.InvalidOperation
 				return
 			}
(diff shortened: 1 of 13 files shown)


soonkuk:
LGTM!

PR #655, opened 2018-11-06
Failed to save `BlockOperation`s

### LINKED ISSUE #654: null operation value from `api/v1/accounts/{id}/operations`
Getting operations from `api/v1/accounts/{id}/operations` is not working.
It shows only null value.
I guess the reason may be that operation was not saved.

spikeekips:
### Github Issue

Resolves #654 

### Background

See #654 and when node is`BOOTING`, the `SavingBlockOperation` process hangs.

### Solution

There were problem in `SavingBlockOperation.CheckByBlock()`, if `blk.Transactions` is under 2, `numWorkers` was `0`.

### DIFF
--- a/lib/node/runner/block_operations.go
+++ b/lib/node/runner/block_operations.go
@@ -110,6 +110,15 @@ func (sb *SavingBlockOperations) savingBlockOperationsWorker(id int, st *storage
 }
 
 func (sb *SavingBlockOperations) CheckByBlock(st *storage.LevelDBBackend, blk block.Block) (err error) {
+	if blk.Height > common.GenesisBlockHeight { // ProposerTransaction
+		if err = sb.CheckTransactionByBlock(st, blk, blk.ProposerTransaction); err != nil {
+			return
+		}
+	}
+	if len(blk.Transactions) < 1 {
+		return
+	}
+
 	txs := make(chan string, 100)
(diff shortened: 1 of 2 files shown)


lib/node/runner/block_operations.go
```diff
@@ -110,41 +110,56 @@ func (sb *SavingBlockOperations) savingBlockOperationsWorker(id int, st *storage
 }
 
 func (sb *SavingBlockOperations) CheckByBlock(st *storage.LevelDBBackend, blk block.Block) (err error) {
+	if blk.Height > common.GenesisBlockHeight { // ProposerTransaction
+		if err = sb.CheckTransactionByBlock(st, blk, blk.ProposerTransaction); err != nil {
+			return
+		}
+	}
+	if len(blk.Transactions) < 1 {
+		return
```
Charleslee522: 👍

Geod24:
Can you add a test that would have failed before this patch ?

soonkuk:
@spikeekips I checked working of api. I can get correct operation value.

Geod24:
Will test soon with delayed ops

spikeekips:
NOTE: I will wait until test result of @Geod24 or the other developers.

Geod24:
Yes, it works 👍

PR #657, opened 2018-11-06
Sending `Transaction` must use `/api/v1/transactions`

### LINKED ISSUE #656: `sebak wallet payment` still uses `/node/message`
After #641, the client must send a transaction to `/api/v1/transactions`. Current cmd payment still uses `/node/message`.

spikeekips:
### Github Issue

Resolves #656 

### Background

See #656 and the command `unfreezeRequest` will also be applied.

### Solution

Just create new method, `NetworkClient.SendTransaction()`.

### DIFF
--- a/lib/network/base.go
+++ b/lib/network/base.go
@@ -38,6 +38,7 @@ type NetworkClient interface {
 	Connect(node node.Node) ([]byte, error)
 	GetNodeInfo() ([]byte, error)
 	SendMessage(common.Serializable) ([]byte, error)
+	SendTransaction(common.Serializable) ([]byte, error)
 	SendBallot(common.Serializable) ([]byte, error)
 	GetTransactions([]string) ([]byte, error)
 }
--- a/lib/network/http2_client.go
+++ b/lib/network/http2_client.go
@@ -77,102 +77,46 @@ func (c *HTTP2NetworkClient) GetNodeInfo() (body []byte, err error) {
 	return
 }
 
-func (c *HTTP2NetworkClient) Connect(n node.Node) (body []byte, err error) {
+func (c *HTTP2NetworkClient) Send(path string, message common.Serializable) (retBody []byte, err error) {
 	headers := c.DefaultHeaders()
 	headers.Set("Content-Type", "application/json")
 
-	serialized, _ := n.Serialize()
+	var body []byte
+	if body, err = message.Serialize(); err != nil {
+		return
+	}
+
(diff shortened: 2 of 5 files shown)


lib/network/http2_client.go
```diff
@@ -122,6 +123,32 @@ func (c *HTTP2NetworkClient) SendMessage(message common.Serializable) (retBody [
...
+func (c *HTTP2NetworkClient) SendTransaction(message common.Serializable) (retBody []byte, err error) {
+	headers := c.DefaultHeaders()
+	headers.Set("Content-Type", "application/json")
+
+	var body []byte
+	if body, err = message.Serialize(); err != nil {
+		return
+	}
+
+	u := c.resolvePath(resource.URLTransactions)
```
Charleslee522: `SendMessage` and `SendTransaction` are almost same except this part.
And `SendBallot` also.
I figure, if this duplication of code resolves, the code will get better.

spikeekips: @Charleslee522 I think so. `SendMessage` is currently using for `Transaction` and 'Connect', so I made new.

Charleslee522: @spikeekips 
I mean,
```
func (c *HTTP2NetworkClient) SendTo(message common.Serializable, uri string) (retBody []byte, err error) {
...
}

func (c *HTTP2NetworkClient) SendMessage(message common.Serializable) (retBody []byte, err error) {
    return SendTo(message, UrlPathPrefixNode + "/message")
}

func (c *HTTP2NetworkClient) SendTransaction(message common.Serializable) (retBody []byte, err error) {
    return SendTo(message, resource.URLTransactions)
}

func (c *HTTP2NetworkClient) SendBallot(message common.Serializable) (retBody []byte, err error) {
    return SendTo(message, UrlPathPrefixNode + "/ballot")
}
```

lib/network/base.go
```diff
@@ -38,6 +38,7 @@ type NetworkClient interface {
 	Connect(node node.Node) ([]byte, error)
 	GetNodeInfo() ([]byte, error)
 	SendMessage(common.Serializable) ([]byte, error)
+	SendTransaction(common.Serializable) ([]byte, error)
```
Geod24: It would be better if this could take the correct type (a `Transaction`) instead of the generic `Serializable` type.
But given the location of this file, it would probably lead to a circular import, which might hint that this interface should be in a different package. But not an urgent matter.

PR #668, opened 2018-11-07
Remove no-more-used code

Geod24:
Since delayed transactions are implemented, we don't need unfreezing requests anymore.


zzim2x:
Um. I see what you mean.

Geod24:
But this PR is useless since @spikeekips wants #624 to be reverted (see #670 ).
I'll re-submit with the delayed transaction code.

PR #672, opened 2018-11-07
PageQuery for  api resource list and storage list options

anarcher:
### Background

```
-       options, err := storage.NewDefaultListOptionsFromQuery(r.URL.Query())
-       if err != nil {
-               http.Error(w, errors.InvalidQueryString.Error(), http.StatusBadRequest)
+
+       p := httputils.NewPaginator(r)
+       if err := p.Error(); err != nil {
+               httputils.WriteJSONError(w, err)
                return
        }
+       options := storage.NewDefaultListOptions(p.Reverse(), p.Cursor(), p.Limit())
     
-       self := r.URL.String()
-       next := strings.Replace(resource.URLAccountOperations, "{id}", address, -1) + "?" + options.SetCursor(cursor).SetRevers
-       prev := strings.Replace(resource.URLAccountOperations, "{id}", address, -1) + "?" + options.SetReverse(true).Encode()
-       list := resource.NewResourceList(txs, self, next, prev)
-
+       list := resource.NewResourceList(txs, p.SelfLink(), p.NextLink(cursor), p.PrevLink(cursor))
```

### DIFF
--- a/lib/network/httputils/pagequery.go
+++ b/lib/network/httputils/pagequery.go
@@ -8,6 +8,7 @@ import (
 
 	"boscoin.io/sebak/lib/common"
 	"boscoin.io/sebak/lib/node/runner/api/resource"
+	"boscoin.io/sebak/lib/storage"
 )
 
 const DefaultMaxLimit uint64 = 100
@@ -56,6 +57,10 @@ func (p *PageQuery) NextLink(cursor []byte) string {
 	query := p.urlValues(cursor, true).Encode()
 	link := fmt.Sprintf("%s?%s", path, query)
 	return link
+}
+
+func (p *PageQuery) ListOptions() storage.ListOptions {
+	return storage.NewDefaultListOptions(p.Reverse(), p.Cursor(), p.Limit())
 }
 
 func (p *PageQuery) ResourceList(rs []resource.Resource, cursor []byte) *resource.ResourceList {
--- a/lib/network/httputils/pagequery.go
+++ b/lib/network/httputils/pagequery.go
@@ -47,14 +47,14 @@ func (p *PageQuery) SelfLink() string {
 
 func (p *PageQuery) PrevLink(cursor []byte) string {
 	path := p.request.URL.Path
-	query := p.urlValues(cursor, false).Encode()
+	query := p.urlValues(cursor, true, p.limit).Encode()
 	link := fmt.Sprintf("%s?%s", path, query)
 	return link
 }
 
 func (p *PageQuery) NextLink(cursor []byte) string {
 	path := p.request.URL.Path
-	query := p.urlValues(cursor, true).Encode()
+	query := p.urlValues(cursor, false, p.limit).Encode()
 	link := fmt.Sprintf("%s?%s", path, query)
 	return link
 }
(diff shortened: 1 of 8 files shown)


anarcher:
It can be better: 
```
       list := resource.NewResourceList(txs,options)
       list.SetSelfLink(r.URL.String())
       list.SetPrevNextLinkPrefix(resource.URLAccountOperations, []string{"{id}", address})
       list.SetPrevLinkListOptions(cursor, false)
       list.SetNextLinkListOptions(cursor, true)
```

anarcher:
How about this changes? If 2/3 is OK, l will change that parts in handlers.

spikeekips:
@anarcher Please also consider this code, https://github.com/bosnet/sebak/blob/806a8d84558294e623707e28b8d6e1f78addaafc/lib/storage/options.go#L99-L120

anarcher:
Yes I don’t like the storage.ListOption has url http things. I feel that broken layered.

spikeekips:
@anarcher Yes I also think so, we can replace listoption things with this PR.

anarcher:
I add Paginator for it

```
-       options, err := storage.NewDefaultListOptionsFromQuery(r.URL.Query())
-       if err != nil {
-               http.Error(w, errors.InvalidQueryString.Error(), http.StatusBadRequest)
+
+       p := httputils.NewPaginator(r)
+       if err := p.Error(); err != nil {
+               httputils.WriteJSONError(w, err)
                return
        }
+       options := storage.NewDefaultListOptions(p.Reverse(), p.Cursor(), p.Limit())
     
-       self := r.URL.String()
-       next := strings.Replace(resource.URLAccountOperations, "{id}", address, -1) + "?" + options.SetCursor(cursor).SetRevers
-       prev := strings.Replace(resource.URLAccountOperations, "{id}", address, -1) + "?" + options.SetReverse(true).Encode()
-       list := resource.NewResourceList(txs, self, next, prev)
-
+       list := resource.NewResourceList(txs, p.SelfLink(), p.NextLink(cursor), p.PrevLink(cursor))
```

lib/network/httputils/paginator.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (p *Paginator) Reverse() bool {
+	return p.reverse
+}
+
+func (p *Paginator) Cursor() []byte {
+	return p.cursor
+}
+
+func (p *Paginator) Error() error {
```
spikeekips: Interesting design, using `Error()` instead of `p, err := NewPaginator()`. Is there any reason of this?

anarcher: I changed `p,err := ` :-)

anarcher:
I will change the name to `PageQuery` from `Paginator` :-) thanks, @spikeekips

anarcher:
Ping  Plz @Charleslee522 @soonkuk  :-)

lib/network/httputils/pagequery.go
```diff
@@ -0,0 +1,109 @@
...
+	query := p.urlValues(cursor, false, p.limit).Encode()
+	link := fmt.Sprintf("%s?%s", path, query)
+	return link
+}
+
+func (p *PageQuery) ListOptions() storage.ListOptions {
+	return storage.NewDefaultListOptions(p.Reverse(), p.Cursor(), p.Limit())
+}
+
+func (p *PageQuery) ResourceList(rs []resource.Resource, cursor []byte) *resource.ResourceList {
```
kfangw: IMO, `httputils` package does not need to depend on the package `resource`
Why don't you call `resource.NewResourceList` at the location where call this function? `ResourceList` ?

soonkuk:
@anarcher Currently `GetFrozenAccountsByAccountHandler` and `GetFrozenAccountsHandler` in `lib/node/runner/api/account.go` also uses paging. How about updating those functions together?

lib/network/httputils/pagequery.go
anarcher: It's a good point. If `resource` pkg uses `httputils`, the `PageQuery` makes circular import.. huum..

anarcher:
I think that `PageQuery` is basically a request struct for api paging support. So I would like to move it to `lib/node/runner/api` pkg.  How do you think about it? @kfangw @soonkuk

PR #673, opened 2018-11-07
Use double linked list in place of slice for hash list in tx pool

### LINKED ISSUE #662: Performance issue in `TransactionPool` Hashes
Please see https://github.com/bosnet/sebak/blob/d02013da616a6ef19097be03929cb189f55f650e/lib/transaction/pool.go#L84

I guess, the `transaction.Pool.hashes` field type is string slice not `map[string]bool` because the order is important.
I think it's okay now, because there are less or equal than 1000 transactions in one block,
but later we should be careful.

anarcher:
### Github Issue

Fixes #662 

### Background

#662

### Solution

Use double linked list in place of slice for hash list in tx pool

### DIFF
--- a/lib/transaction/pool.go
+++ b/lib/transaction/pool.go
@@ -104,14 +104,16 @@ func (tp *Pool) AvailableTransactions(transactionLimit int) []string {
 	defer tp.RUnlock()
 
 	var ret []string
+	var cnt int
 	// first ouput by order older hash
 	for e := tp.hashList.Front(); e != nil; e = e.Next() {
-		if len(ret) == transactionLimit {
+		if cnt >= transactionLimit {
 			return ret
 		}
 		hash, ok := e.Value.(string)
 		if ok {
 			ret = append(ret, hash)
+			cnt++
 		}
 	}
 
--- a/lib/transaction/pool.go
+++ b/lib/transaction/pool.go
@@ -1,6 +1,7 @@
 package transaction
 
 import (
+	"container/list"
 	"sync"
 )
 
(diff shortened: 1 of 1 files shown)


lib/transaction/pool.go
```diff
@@ -103,12 +107,16 @@ func (tp *Pool) AvailableTransactions(transactionLimit int) []string {
 
 	var ret []string
 	// first ouput by order older hash
-	for _, key := range tp.hashes {
+	for e := tp.hashList.Front(); e != nil; e = e.Next() {
```
spikeekips: 👍

lib/transaction/pool.go
```diff
@@ -103,12 +107,16 @@ func (tp *Pool) AvailableTransactions(transactionLimit int) []string {
...
 		if len(ret) == transactionLimit {
```
spikeekips: Instead of checking `len` inside loop, how about using counter?

lib/transaction/pool.go
```diff
@@ -83,11 +89,9 @@ func (tp *Pool) Remove(hashes ...string) {
 		if tx, found := tp.Pool[hash]; found {
 			delete(tp.sources, tx.Source())
 			delete(tp.Pool, hash)
-			for i, h := range tp.hashes {
-				if h == hash {
-					tp.hashes = append(tp.hashes[:i], tp.hashes[i+1:]...)
-					break
-				}
+			if e, ok := tp.hashMap[hash]; ok {
+				tp.hashList.Remove(e)
```
Charleslee522: 👍

Charleslee522:
@Geod24 Please make it simple in next pr :)

anarcher:
> It's starting to get out of control.
@Geod24  Do you have any idea? If you have, plz make next PR. :-)
`Pool *list.List`

Geod24:
Next PR it is then! :)

anarcher:
@Geod24 Can you approved it? I need one vote for merging this PR ㅠㅠ

lib/transaction/pool.go
anarcher: @spikeekips I change to use counter instead of checking len().

PR #689, opened 2018-11-08
message versioning

soonkuk:
### Github Issue
Resolves ##31

### Background
Ballot and Transaction message Versioning.


lib/common/version.go
```diff
@@ -0,0 +1,6 @@
+package common
+
+const (
+	TransactionMessage_Version1 = "tx_version1"
```
spikeekips: Please remove `_`, `TransactionMessage_Version1` -> `TransactionMessageVersion1`. How about remove `Message` in name? Just `TransactionVersion`, `Version1` does not mean anything, I think.

lib/common/version.go
```diff
@@ -0,0 +1,6 @@
...
+	BallotMessage_Version1      = "ballot_version1"
+)
```
spikeekips: @soonkuk We talked about version, the version value should be `1`, not `tx_version1`

cmd/sebak/cmd/key/verify
```diff
@@ -0,0 +1,131 @@
+package key
```
spikeekips: @soonkuk What is this? Is it related to this PR?

cmd/sebak/cmd/wallet/payment.go
```diff
@@ -189,6 +189,7 @@ func makeTransactionCreateAccount(kpSource keypair.KP, kpDest keypair.KP, amount
 
 	tx := transaction.Transaction{
 		H: transaction.Header{
+			Version: common.TransactionVersionV1V1,
```
spikeekips: `TransactionVersionV1V1` is right? `TransactionVersionV1`?

lib/common/message.go
```diff
@@ -6,9 +6,11 @@ import (
 )
 
 const (
-	ConnectMessage     MessageType = "connect"
-	TransactionMessage MessageType = "transaction"
-	BallotMessage      MessageType = "ballot"
+	ConnectMessage       MessageType = "connect"
+	TransactionMessage   MessageType = "transaction"
+	BallotMessage        MessageType = "ballot"
```
Charleslee522: New line is better for distinguish between `MessageType` and `Version`

cmd/sebak/cmd/key/verify
soonkuk: @spikeekips Oh it's my mistake. I'll remove that file.

PR #692, opened 2018-11-08
ricardian contract with yaml format

### LINKED ISSUE #294: Ricardian Contract format.
The ricardian contract is used to describe the condition and action of PF.
It will be signed and recorded in the blockchain by the publisher.
When PF voting procedure is executed, each voting transaction must refer to the ricardian contract
and signed voting transaction will be the evidence that voter has agreed on the contents of the ricardian contract.
It will need to first define the ricardian contract format.

soonkuk:
### Github Issue
Resolves #294


zzim2x:
I think it is complicated and YAML validation logic is required. `ricardian._sample.yaml` does not fit YAML format.

I do not know why the whole document is YAML.

PR #686, opened 2018-11-08
Update License file

outsideris:
It was a raw format.

Issue #687, opened 2018-11-08
Not exporting block height field of block operation.


soonkuk:
Currently block operation export block height field.
It is used for calculating unfreezing period.
But to prevent saving height info redundantly, it is recommended that height is exported only in block not in block transaction and in block operation.
So we will change block operation and modify code about calculation of unfreezing period.

PR #693, opened 2018-11-09
Fixed 0 `block.BlockOperation.Height`

spikeekips:
### Github Issue
Resolves #686 

### Background

See #686 

### Solution

Simply patched.


outsideris:
You linked a wrong issue. #687 maybe?

lib/block/transaction.go
```diff
@@ -180,6 +180,15 @@ func (bt *BlockTransaction) SaveBlockOperations(st *storage.LevelDBBackend) (err
 		return errors.FailedToSaveBlockOperaton
 	}
 
+	if bt.blockHeight < 1 {
```
Geod24: It feels like patching after the fact. How come we call `SaveBlockTransaction` with a default-initialized height ?

spikeekips: @Geod24 It's a mistake this part is missed.

soonkuk:
@spikeekips I originally thought that height of block operation should be removed and should change the reward period calculation code not using the height of block operation but using the height of block after talked with you.
How about going to that way? 
then I'll make the PR.

spikeekips:
> @soonkuk : @spikeekips I originally thought that height of block operation should be removed and should change the reward period calculation code not using the height of block operation but using the height of block after talked with you. How about going to that way?  then I'll make the PR.

Yes, we will, but before implementing new code, the current SEBAK should be working :)

lib/block/transaction.go
Geod24: @spikeekips : Since we have `if bt.blockHeight < 1`, I assume it's not always missing. In which case is it ?

spikeekips: @Geod24 `BlockTransaction.Height` is 0, means `BlockTransaction` is get from storage. `BlockTransaction.Height` is only available when it is saved.

soonkuk:
@spikeekips I think I can fix it easily.
Maybe currently height of block operation is used only unfreezing request and frozen account api.

spikeekips:
@soonkuk The `BlockOperation` and even `BlockTransaction` need some diet in field and data to reduce the data size, so before diet, we'd better to use current impl.

soonkuk:
@spikeekips Ok, I got it.

spikeekips:
This issue blocks the other development, so I will merge if there are no more opinions.

PR #695, opened 2018-11-09
block-hash in the response of /api/v1/transactions

### LINKED ISSUE #678: Client API: block-height doesn't exist in the response of `/api/v1/transactions/{hash}`
I can't figure where a transaction is.

soonkuk:
### Github Issue
Resolves #678

### DIFF
--- a/lib/node/runner/api/resource/transaction.go
+++ b/lib/node/runner/api/resource/transaction.go
@@ -21,6 +21,7 @@ func NewTransaction(bt *block.BlockTransaction) *Transaction {
 func (t Transaction) GetMap() hal.Entry {
 	return hal.Entry{
 		"hash":            t.bt.Hash,
+		"block":           t.bt.Block,
 		"source":          t.bt.Source,
 		"fee":             t.bt.Fee.String(),
 		"sequence_id":     t.bt.SequenceID,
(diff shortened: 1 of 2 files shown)


lib/node/runner/api/transaction_test.go
```diff
@@ -49,6 +49,7 @@ func TestGetTransactionByHashHandler(t *testing.T) {
 		json.Unmarshal(readByte, &recv)
 
 		require.Equal(t, bt.Hash, recv["hash"], "hash is not same")
+		require.Equal(t, bt.Block, recv["block"], "block is not same")
```
Geod24: I know the mistake is already there, but the proper message would be "block is not the same".
However, since `Equal` already shows that message with a diff, I think that message is redundant.

lib/node/runner/api/resource/transaction.go
```diff
@@ -21,6 +21,7 @@ func NewTransaction(bt *block.BlockTransaction) *Transaction {
 func (t Transaction) GetMap() hal.Entry {
 	return hal.Entry{
 		"hash":            t.bt.Hash,
+		"block":           t.bt.Block,
```
Geod24: Doesn't that return the full block ?

lib/node/runner/api/transaction_test.go
soonkuk: @Geod24 Ok, I revised the message.

PR #696, opened 2018-11-09
fix QueryOrder

niulingyun:


Geod24:
Looks like you caught a bug! Do you want to pursue it or do you prefer us to take over ?

PR #701, opened 2018-11-11
[WIP]Apply no-fee policy to operation related to frozen account

### LINKED ISSUE #682: Apply no-fee to a transaction related to frozen account.
Apply no-fee to a transaction related to frozen account.
It affects freezing, unfreezing request and unfreezing payment.

soonkuk:
### Github Issue
Resolves #682

### Solution
Each operation type has HasFee() method to decide the policy about the payment of fee.
But that of create-account and payment is changed according to operation's target or transaction's source.
If those operations is related to frozen account, there is no fee.
But those operations is not related to frozen account, there is a fee.
So I will make freezing operation and unfreezing operation. Then we can make create-account and payment to pay a fee and freezing and unfreezing to be excluded from payment of fee.


soonkuk:
@kfangw Ok I will consider it.

cmd/sebak/cmd/wallet/payment.go
```diff
@@ -170,8 +178,8 @@ func init() {
 /// Returns:
 ///   `sebak.Transaction` = The generated `Transaction` creating the account
 ///
-func makeTransactionCreateAccount(kpSource keypair.KP, kpDest keypair.KP, amount common.Amount, seqid uint64, target string) transaction.Transaction {
-	opb := operation.NewCreateAccount(kpDest.Address(), amount, target)
+func makeTransactionCreateAccount(kpSource keypair.KP, kpDest keypair.KP, amount common.Amount, seqid uint64, linked string) transaction.Transaction {
+	opb := operation.NewCreateAccount(kpDest.Address(), amount, linked)
```
kfangw: `NewCreateAccount` method should not accept `linked` argument anymore

lib/block/operation.go
```diff
@@ -92,14 +92,14 @@ func (bo *BlockOperation) Save(st *storage.LevelDBBackend) (err error) {
...
-	if bo.Type == operation.TypeCreateAccount {
+	if bo.Type == operation.TypeFreezing {
 		if body, err = operation.UnmarshalBodyJSON(bo.Type, bo.Body); err != nil {
 			return err
 		}
-		if casted, ok = body.(operation.CreateAccount); !ok {
+		if casted, ok = body.(operation.Freezing); !ok {
 			return errors.TypeOperationBodyNotMatched
 		}
 		if casted.Linked != "" {
```
kfangw: Freezing operation should have `Linked` field always

lib/node/runner/checker_ballot_transaction.go
```diff
@@ -273,10 +273,26 @@ func ValidateOp(st *storage.LevelDBBackend, source *block.BlockAccount, op opera
 		if casted, ok = op.B.(operation.CreateAccount); !ok {
 			return errors.TypeOperationBodyNotMatched
 		}
+		if casted.Linked != "" {
```
kfangw: Just remove `Linked` field from operation `CreateAccount`

lib/node/runner/checker_ballot_transaction.go
```diff
@@ -273,10 +273,26 @@ func ValidateOp(st *storage.LevelDBBackend, source *block.BlockAccount, op opera
...
+		var ok bool
+		var casted operation.Freezing
+		if casted, ok = op.B.(operation.Freezing); !ok {
+			return errors.TypeOperationBodyNotMatched
+		}
+		if casted.Linked == "" {
+			return errors.FrozenAccountMustHaveLinkedAccount
+		}
+		var exists bool
+		if exists, err = block.ExistsBlockAccount(st, op.B.(operation.Freezing).Target); err == nil && exists {
```
kfangw: use `casted` instead of `op.B.(operation.Freezing)`

lib/transaction/operation/create_account.go
```diff
@@ -42,6 +42,11 @@ func (o CreateAccount) IsWellFormed(common.Config) (err error) {
 		return
 	}
 
+	if o.Linked != "" {
```
kfangw: remove

kfangw:
I have a question to catch up the progress of you guys
@soonkuk @spikeekips @zzim2x 

AFAIK,
Freezing is a special type of Create Account and
UnFreezing is a special type of Payment.

Why those operations granted the privilege?
It makes the system hard to maintain.

soonkuk:
@kfangw Can you explain what does means the privilege?

kfangw:
> @kfangw Can you explain what does means the privilege?

no-fee

cmd/sebak/cmd/wallet/payment.go
```diff
@@ -199,6 +207,36 @@ func makeTransactionCreateAccount(kpSource keypair.KP, kpDest keypair.KP, amount
 	return tx
 }
 
+// makeTransactionFreezing creates frozen account.
+func makeTransactionFreezing(kpSource keypair.KP, kpDest keypair.KP, amount common.Amount, seqid uint64, linked string) transaction.Transaction {
```
Geod24: Why do you make two functions ? It just just be:
```go
var fee common.Amount
 if len(linked) == 0 {
  fee = common.BaseFee
}
```

And use this when making the tx

soonkuk: @Geod24 But we should also validate the fee, when tx comes.
I think it is somehow complicated if we do not use two operations.

Geod24: Yes, and it's trivial, because if the op has a `linked` account, we just don't check for fee.

soonkuk: Yes but it also check the source of transaction.
For example, unfreezing payment is also no-fee.
To know the whether payment and create-account are unfreezing, we should know the source account' type(frozen account) also.

Geod24: Yes but we have that already. Check [ValidateTx](https://github.com/bosnet/sebak/blob/c19dec0e98ff67177dd13c2aae59bd810a118430/lib/node/runner/checker_ballot_transaction.go#L208)

soonkuk:
#715 
I make another PR for this issue. It is implemented without making new operation.
Please check this PR.

soonkuk:
I will close this PR and #715 will replace this.

PR #705, opened 2018-11-12
#659  Adaptable Discoverying Nodes

### LINKED ISSUE #659: New way to connect to the other nodes
~~Originally `/connect` was devised for connecting nodes each other, so `/connect` will send the node information and the the other node checks it. The missing thing is if the node information is updated like it's endpoint or it's validators, the other nodes should update also validators information. With updating validator information, the node will broadcast to the updated endpoint without restarting.~~

~~* To validate the data from `/connect`, the message of `/connect` should be signed by node's key~~
~~* If validated and some informations are changed, the received node should update it's validator.~~

Currently the node connects to the other validator nodes,

* Node must know the endpoint and public address of each nodes.
* Node must check the connection is made with each nodes.

This way is simple and easy to understand, but this may have some inefficiency and weaknesses for node operation,

* There is not validation process, the connected node is **really** the node.
* If the endpoint of some node is changed, the other node must restart with new endpoint.

---
To make this way to be better, I suggest new way, that is,

* We need the new 2 message, `ConnectMessage` and `ConnectedMessage`
  * `ConnectMessage` and `ConnectedMessage` have the signature, signed by sender node.
  * Node sends `ConnectMessage` and gets the `ConnectedMessage` as response.

* Node must know all the public addresses of the other nodes and **at least one of endpoints** of the other nodes.
* When node starts, node sends `ConnectMessage` to the other nodes, which the node knows their endpoints.

* The other nodes can verify the received `ConnectMessage`.
* The other nodes responses the `ConnectedMessage`, which has the nodes of the received node.
* If in responded `ConnectedMessage` there are the nodes, which is not connected, the node sends `ConnectMessage` to the unconnected nodes
* Basically `ConnectMessage` has also `node.NodeInfo`, which already contains published `Endpoint` of node.

This new way can make,

* The validation can be possible whether the connected node is what I want to connect.
* If the endpoint of node is changed, the other node easily know the new endpoint, so without shutdown, all the operation of node will be fine.

spikeekips:
## Adaptable Connecting Nodes

Resolves #659 

At this time, a node must know these things about its validators,

* all public addresses of nodes
* all endpoints of nodes

When one of the endpoints is updated, all the nodes in the network must be restarted with the updated endpoints. This will be a painful job for maintanence.

This patch will solve this problem and new endpoint of validators will be updated without downtime.

#### `--validators` only with public address

The current way to set `--validators` is,

```sh
--validators "self https://127.0.0.1:2822?address=G..W&alias=node2 https://127.0.0.1:2823/?address=G..4&alias=node3 https://127.0.0.1:2824/?address=G..O&alias=node4"
```

The new way is,

```sh
--validators "self G..W G..4 G..O"
```

Node just knows only the public address, the variable information like endpoint and alias are removed.

The environment variable also can be set,

```
SEBAK_VALIDATORS="self G..W G..4 G..O"
```

#### New option, `--discovery`

```sh
--discovery https://1.2.3.4:5678 --discovery https://8.7.6.5:4321
```

Instead of setting endpoint in `--validators` option, using `--discovery` node will try to discover the validators. When a node starts, the node will send `network.DiscoveryMessage` to `--discovery` values. Each node sends and receives `network.DiscoveryMessage`s, the node can be started until the number of **discovered** validators is over consensus threshold.

`--discovery` also can be set by an environment variable, `SEBAK_DISCOVERY`, multiple values can be separated by empty space.

Omitting `-discovery` option also be possible. Without `--discovery` node will just wait `network.DiscoveryMessage` from other nodes.

### WatcherMode

Watcher does not care about the validator, it just needs the target node to watch. Watcher will use `--discovery` instead of `--validators`, the node address will be fetched from the node info of `--discovery` node. At this process, the remote node info is compared with watcher; the network id and initial balance are matched.

### Minor changes

* `node.LocalNode.Endpoint()` will return if `publishEndpoint` is set
* `node.LocalNode.ClearValidators()` will make empty `validators`

### Etc

This patch was severely tested in integration test :)


spikeekips:
The test cases will be added.

Geod24:
There's a lot to test. And it needs a rebase. Could you update ? I'll focus on reviewing/testing this tomorrow so we can get it merged quickly.

spikeekips:
Entirely newly rewrote. Please continue to review

lib/network/discovery_message.go
```diff
@@ -0,0 +1,170 @@
...
+		return errors.InvalidMessage
+	}
+
+	// check time
+	created, err := common.ParseISO8601(dm.B.Created)
+	if err != nil {
+		return err
+	}
+	now := time.Now()
+	sub := created.Sub(now)
```
Charleslee522: How about this? I think this is a little bit better to understand :)
```
now := time.Now()
sub := now.Sub(created)

if sub > common.DiscoveryMessageCreatedAllowDuration {
    return errors.MessageHasIncorrectTime
}

if sub < 0 {
    return errors.MessageHasIncorrectTime
}
```

lib/network/discovery_message.go
```diff
@@ -0,0 +1,170 @@
...
+
+	hash := dm.B.MakeHashString()
+	signature, _ := keypair.MakeSignature(kp, networkID, hash)
+
+	dm.H.Signature = base58.Encode(signature)
+
+	return
+}
+
+func (dm DiscoveryMessage) Serialize() (encoded []byte, err error) {
```
Charleslee522: Serialize() doesn't need named return variable.

Serialize() `([]byte, error)`

lib/network/memory_client.go
```diff
@@ -31,8 +31,7 @@ func (m *MemoryTransportClient) Connect(node node.Node) (b []byte, err error) {
 }
 
 func (m *MemoryTransportClient) GetNodeInfo() (b []byte, err error) {
-	b = m.server.GetNodeInfo()
-	return
+	return []byte{}, errors.NotImplemented
```
Charleslee522: I think GetNodeInfo() doesn't need named return variable.

lib/network/validator_connection_manager.go
```diff
@@ -257,3 +266,188 @@ func (c *ValidatorConnectionManager) watchForMetrics() {
 func (c *ValidatorConnectionManager) IsReady() bool {
 	return len(c.AllConnected()) >= c.policy.Threshold()
 }
+
+// startDiscovery will try to discover the validators; it will block until
+// enough validators is discovered, and then it will keep discovering the left
+// validators.
```
Charleslee522: Good documentation 👍

lib/network/discovery_message.go
spikeekips: @Charleslee522 `sub` can be under zero, because we should check the time span 'before' and 'after'

Charleslee522: @spikeekips Okay, I got it.
This has another point also.
`now` is usually later than created, so I think `now.Sub(created)` is better to understand than `created.Sub(now)`.

lib/network/discovery_message.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	// check time
+	created, err := common.ParseISO8601(dm.B.Created)
+	if err != nil {
+		return err
+	}
+	now := time.Now()
+	sub := created.Sub(now)
+	if sub < 0 && sub < (common.DiscoveryMessageCreatedAllowDuration*-1) {
```
Charleslee522: `common.DiscoveryMessageCreatedAllowDuration*-1` is `-10 sec`, so in this condition `sub < 0 && sub < -10`, `sub < 0` is redundant.

lib/network/discovery_message.go
```diff
@@ -0,0 +1,170 @@
...
+		return errors.MessageHasIncorrectTime
+	}
+
+	if sub > 0 && sub > common.DiscoveryMessageCreatedAllowDuration {
```
Charleslee522: `common.DiscoveryMessageCreatedAllowDuration` is `10 sec`, so in this condition `sub > 0 && sub > 10`, `sub > 0` is redundant.

lib/network/validator_connection_manager.go
```diff
@@ -257,3 +266,188 @@ func (c *ValidatorConnectionManager) watchForMetrics() {
...
+		c.log.Debug(
+			"new validators found",
+			"from", dm.B.Address,
+			"received", dm.B.Validators,
+			"new", undiscovered,
+			"previous", current,
+			"after", c.discovered(),
+		)
+
+		c.setDiscovered(undiscovered...)
```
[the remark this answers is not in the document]
kfangw: Yes, I understand what `undiscovered` is.
But I wonder why to set the undiscovered with `setDiscovered` method.
`setDiscovered` method looks like set the `endpoint` of the validators. @spikeekips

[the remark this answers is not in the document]
spikeekips: @kfangw So you think, just update `Endpoint` without making a new method, `setDiscovered()`, right?

Issue #681, opened 2018-11-08
Validating Proposal Transaction with specific account key


soonkuk:
A proposal transaction should be proposed by the unique account.
We did not verify it currently.
To do this, specific account key should be set when SEBAK starting.
And nodes should verify the proposal transaction whether it was signed by specific account key.

kfangw:
resolved by #709 #719

PR #709, opened 2018-11-13
681 congress address handle

kfangw:
### Github Issue
#681 

### Background
CongressVoting(Result) Operations accepted only from CongressAddress

### Solution
- Any account can submit a transaction with CongressVoting(Result) operation
- The operation also charged transaction fee
- The account needed to be created before submitting the transaction.
- People knows what the Congress Address is.

However,
- To control the first several votings appropriately, only tx from the hard-coded account(Congress Address) is accepted temporary

### Possible Drawbacks
-


kfangw:
@spikeekips @soonkuk @outsideris please review.

tests/client/congressvoting_test.go
```diff
@@ -14,11 +14,14 @@ import (
...
 	"github.com/stellar/go/keypair"
 	"github.com/stretchr/testify/require"
+	"boscoin.io/sebak/lib/common"
 )
 
 func TestCongressVoting(t *testing.T) {
 
 	var (
+		CongressAddr  = "GAIJAH4FCEB3AWS2NVKRNDGVJ5QPD7VNOQNGZEPL2FAI6GPZQNKEXRWI"
+		CongressSecret = "SBUONPFGBX7BF76ZZ2NOWWCJPG6NQQOQS37IL27JCUBPFAXFRYS3VDI2"
```
soonkuk: This `CongreeAddr` is used also in `constant.go`. If this address is used without change.
This secret key can be known to all.
How about change the `CongreeAddr` and `CongressSecret` of test to another keypair?

kfangw: It is a client's testcode.
The Address should be the same as SEBAK's Congress Address.
When the SEBAK published, the address should be set by the environment variable.

kfangw: I mean that The string set in the constant.go is just default code.
It should be changed when launching the network. @soonkuk

cmd/sebak/cmd/run.go
```diff
@@ -488,6 +491,10 @@ func getTimeDuration(str string, defaultValue time.Duration, errMessage string)
 }
 
 func runNode() error {
+
+	// Set common variable
+	common.CongressAddr = flagCongressAddress
```
spikeekips: The other important shared variables are managed by `common.Config`.

lib/transaction/operation/congress_voting_result.go
```diff
@@ -80,5 +80,5 @@ func (o CongressVotingResult) IsWellFormed(common.Config) (err error) {
 }
 
 func (o CongressVotingResult) HasFee() bool {
-	return false
+	return true
```
spikeekips: 👍

tests/client/congressvoting_test.go
soonkuk: @kfangw Ok I got it.

kfangw:
@spikeekips @soonkuk  could you confirm this?

kfangw:
@spikeekips PING~

lib/common/config.go
```diff
@@ -50,5 +52,7 @@ func NewConfig(networkID []byte) Config {
 
 	p.HTTPCachePoolSize = HTTPCachePoolSize
 
+	p.CongressAddr = "GAIJAH4FCEB3AWS2NVKRNDGVJ5QPD7VNOQNGZEPL2FAI6GPZQNKEXRWI"
```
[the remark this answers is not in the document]
spikeekips: @kfangw Please consider upper comment.

Issue #704, opened 2018-11-12
Block Header contains Timestamp which is non deterministic


kfangw:
https://github.com/bosnet/sebak/blob/ed5458f6df19bd92697b3d5357a48d3035591bd3/lib/block/header.go#L23-L32

https://github.com/bosnet/sebak/blob/ed5458f6df19bd92697b3d5357a48d3035591bd3/lib/block/block.go#L46-L60

When the block `Hash` extracted, each node calculated Timestamp based on its own time seed.
It means that the `Hash` may diverse from each other.

If the hash is diverse then, this should be violated by consensus. At the consensus phase, the previous hash of the proposed block should be the same as a node's one.

Or the Hash calculation algorithm ignores `Timestamp` somehow?

@spikeekips @Charleslee522 @anarcher

spikeekips:
> I left for history.

I ran the simple code like this,
```go
package main

type A struct {
	T time.Time
	A string
}

func main() {
	k := time.Now()
	{
		b := A{T: k, A: "showme"}
		a, err := common.MakeObjectHash(b)
		fmt.Println(a, err)
	}

	{
		b := A{T: k, A: "showme"}
		a, err := common.MakeObjectHash(b)
		fmt.Println(a, err)
	}

	{
		b := A{A: "showme"}
		a, err := common.MakeObjectHash(b)
		fmt.Println(a, err)
	}
}
```

The result,
```
[236 12 239 200 138 104 12 173 42 213 58 69 58 16 160 140 244 125 243 171 2 65 175 93 41 37 194 234 30 199 219 69] <nil>
[236 12 239 200 138 104 12 173 42 213 58 69 58 16 160 140 244 125 243 171 2 65 175 93 41 37 194 234 30 199 219 69] <nil>
[236 12 239 200 138 104 12 173 42 213 58 69 58 16 160 140 244 125 243 171 2 65 175 93 41 37 194 234 30 199 219 69] <nil>
```

`rlp` does not encode `time.Time`, so our `Timestamp` does not affect the `block.Hash` :)

kfangw:
Hence,
A fault(misuse of RLP) prevents a failure(hash mismatch) caused by another fault(timestamp in header).

spikeekips:
@kfangw @Charleslee522 I agree. We should consider to move `Header.Timestamp` to another place.

kfangw:
Is the timestamp in the block header necessary?
[...]

Issue #703, opened 2018-11-12
TransactionsRoot of Block Header does not contains ProposerTransaction


kfangw:
https://github.com/bosnet/sebak/blob/ed5458f6df19bd92697b3d5357a48d3035591bd3/lib/block/block.go#L46-L60

Is it intended?

@spikeekips

spikeekips:
@kfangw Oops, it's important bug. Please patch it.

kfangw:
solved by #712

PR #712, opened 2018-11-13
fixes 703, 704 Tx root and Block `time` things

kfangw:
### Github Issue
#703 #704 

### Background
- Block Header should contain fields which are deterministic for each node.
- Transaction Root should contain ProposerTransaction also

### Solution
- Block header contains `ProposedTime` like timestamp of bitcoin
- ConfirmedTime is added to Block body. But not encoded for RLP and Hash of block

--> previous Timestamp is refactored to ConfirmedTime and move from header to body. And it is not encoded as RLP explicitly

### Possible Drawbacks
```
	ConfirmedTime string `json:"confirmed-time" rlp:"-"`
```
rlp:"-" is very very important for this patch.


lib/block/block.go
```diff
@@ -45,14 +45,13 @@ func (bck Block) IsEmpty() bool {
 
 // NewBlock creates new block; `ptx` represents the
 // `ProposerTransaction.GetHash()`.
-func NewBlock(proposer string, basis voting.Basis, ptx string, transactions []string, confirmed string) *Block {
+func NewBlock(proposer string, basis voting.Basis, ptx string, transactions []string, proposedTime string) *Block {
 	b := &Block{
-		Header:              *NewBlockHeader(basis, getTransactionRoot(transactions)),
+		Header:              *NewBlockHeader(basis, getTransactionRoot(append([]string{ptx}, transactions...)), proposedTime),
```
kfangw: #703 needs this single line patch only.
Every other changes are for #704

kfangw:
@Charleslee522 Can you help me for investigating the CI fail? `ci/circleci: integration_tests_long_term`

Charleslee522:
@kfangw Please rebase after merging #711 :)

lib/block/block.go
```diff
@@ -23,10 +23,10 @@ type Block struct {
...
 	//PrevConsensusResult ConsensusResult
 
-	Hash      string `json:"hash"`
-	Confirmed string `json:"confirmed"`
-	Proposer  string `json:"proposer"` /* Node.Address() */
-	Round     uint64 `json:"round"`
+	Hash          string `json:"hash"`
+	Proposer      string `json:"proposer"` /* Node.Address() */
+	Round         uint64 `json:"round"`
+	ConfirmedTime string `json:"confirmed-time" rlp:"-"`
```
spikeekips: @kfangw We have used `Confirmed` term, without `Time` suffix. I wish we follow the traditions. :)

spikeekips: @kfangw ~~Is there any reason to use `string` instead of `time.Time`? `time.Time` also can be stored in leveldb. Parsing time string will be cost.~~ I understand to use `string`, it is used for creating key.

lib/block/header.go
```diff
@@ -2,32 +2,31 @@ package block
...
-	PrevBlockHash    string    `json:"prev_block_hash"`   // TODO Uint256 type
-	TransactionsRoot string    `json:"transactions_root"` // Merkle root of Txs // TODO Uint256 type
-	Timestamp        time.Time `json:"timestamp"`
-	Height           uint64    `json:"height"`
-	TotalTxs         uint64    `json:"total-txs"`
-	TotalOps         uint64    `json:"total-ops"`
+	Version          uint32 `json:"version"`
+	PrevBlockHash    string `json:"prev_block_hash"`   // TODO Uint256 type
+	TransactionsRoot string `json:"transactions_root"` // Merkle root of Txs // TODO Uint256 type
+	ProposedTime     string `json:"proposed-time"`
```
spikeekips: 👍

Charleslee522:
> @kfangw Please rebase after merging #711 :)

now~ :)

lib/block/header.go
anarcher: I would like to use `_` in place of `-` in json.  like `transactions_root`  :-)

kfangw: It was my mistake ^^;; I prefer that too.

There are many of this case. use `-` T^T
For this case, I changed `-` to `_`
I will make another PR to refactor remains. @anarcher

lib/node/runner/isaac_state_manager.go
```diff
@@ -64,7 +64,7 @@ func (sm *ISAACStateManager) setTheFirstConsensusBlockTime() {
 	if err != nil {
 		return
 	}
-	sm.firstConsensusBlockTime = blk.Header.Timestamp
+	sm.firstConsensusBlockTime, _ = common.ParseISO8601(blk.Confirmed)
```
Geod24: Missing error checking

lib/block/block.go
```diff
@@ -74,14 +73,15 @@ func getBlockKeyPrefixHeight(height uint64) string {
 func (b Block) NewBlockKeyConfirmed() string {
 	return fmt.Sprintf(
 		"%s%s-%s%s",
-		common.BlockPrefixConfirmed, b.Confirmed,
+		common.BlockPrefixConfirmed, b.ProposedTime,
```
Geod24: Do we ever use this ?

lib/block/block.go
```diff
@@ -74,14 +73,15 @@ func getBlockKeyPrefixHeight(height uint64) string {
...
 		common.EncodeUint64ToByteSlice(b.Height),
 		common.GetUniqueIDFromUUID(),
 	)
 }
 
 func (b *Block) Save(st *storage.LevelDBBackend) (err error) {
 	key := getBlockKey(b.Hash)
+	b.Confirmed = common.NowISO8601()
```
Geod24: Are we sure the `Save` time is deterministic ?

kfangw: It is not deterministic. And not need to be deterministic. Because it is not used for calculating Hash.

spikeekips:
If possible, I wish this PR will be merged Today.

Issue #459, opened 2018-09-28
PF issuance


soonkuk:
Generate coin issuance for PF budget

soonkuk:
Generate coin issuance for PF

soonkuk:
#308

soonkuk:
PF issuance is a kind of inflation.
This inflation is a special transaction only congress voting account can do and this transaction pays no fee.
To do this, SEBAK knew the congress voting account from starting.
The congress voting account makes transactions for saving proposal, voting result and generate issuance.
The congress voting account use valid secret key and public address to sign transaction but real account is not need to be created because the congress voting account does not send coin and does not pay fee.

Issuance operation data

name | type | description
:----|:--------|:-------
voting | string | tx hash of voting result(congress voting result)
amount | uint64 | inflation amount
target | string | target address

kfangw:
I take this.

kfangw:
solved by #719

PR #719, opened 2018-11-14
inflation pf

kfangw:
### Github Issue
#459 

### Background
As a result of Congress voting, Public financing issues additional BOS Coin.
This operation should be submitted by the common budget address only
And the Coin should be sent to the address which denoted by `CongressVoting operation`

### Solution
- `InflationPF` operation is newly introduced
- Operation submitter is the common budget address only
- InflationPF Operation contains CongressVotingResult Hash
- The Operation has `Funding address`. Funding address should be the same as one set in CongressVoting


kfangw:
@spikeekips @soonkuk Please review~

lib/node/runner/checker_ballot_transaction.go
```diff
@@ -265,16 +265,15 @@ func ValidateTx(st *storage.LevelDBBackend, tx transaction.Transaction) (err err
 //   source = Account from where the transaction (and ops) come from
 //   tx = Transaction to check
 //
-func ValidateOp(st *storage.LevelDBBackend, source *block.BlockAccount, op operation.Operation) (err error) {
+func ValidateOp(st *storage.LevelDBBackend, source *block.BlockAccount, op operation.Operation) error {
```
kfangw: IMO, just `return` is fault-prune for this kind of function.
Hence, I make it explicit.

lib/transaction/operation/operation_inflation_pf.go
```diff
@@ -0,0 +1,48 @@
+package operation
+
+import (
+	"encoding/json"
+
+	"boscoin.io/sebak/lib/common"
+	"boscoin.io/sebak/lib/errors"
+)
+
+type InflationPF struct {
```
kfangw: no target needed for InflationPF

docker/node1.env
```diff
@@ -10,4 +10,6 @@ SEBAK_VALIDATORS=self https://127.0.0.1:2822?address=GAYGELM74WJMKSLDN5YP2VAMP64
 # This node's public key
 SEBAK_GENESIS_BLOCK=GDIRF4UWPACXPPI4GW7CMTACTCNDIKJEHZK44RITZB4TD3YUM6CCVNGJ
 # common account
-SEBAK_COMMON_ACCOUNT=GDIV7E7WMBHCYYZHBV6MXHCHC4VCLU4BTCEM2YUQK73EKIM5TPC66VYR
+SEBAK_COMMON_ACCOUNT=GCTVCU764UPXKRJW5DS5AWF5ETCCIYPZTWXN5U5CLUNAAJH6D5NGEIIH
```
kfangw: I couldn't find the secret key for the common account.
Hence, I created a new one.

kfangw: @Charleslee522 FYI.

lib/node/runner/checker_ballot_transaction.go
```diff
@@ -346,12 +346,30 @@ func ValidateOp(st *storage.LevelDBBackend, source *block.BlockAccount, op opera
...
+		var ok bool
+		var casted operation.InflationPF
+		if casted, ok = op.B.(operation.InflationPF); !ok {
+			return errors.TypeOperationBodyNotMatched
+		}
+		exists, err := block.ExistsBlockOperation(st, casted.VotingResult)
+		if err != nil || !exists {
+			return errors.InflationPFResultMissed
+		}
+		commonAccount, err := GetCommonAccount(st)
```
spikeekips: Getting common account is already checked when node runner starts and `runner.NodeRunner.CommonAccountAddress` has common account address. We can use it.

lib/transaction/operation/operation_inflation_pf.go
```diff
@@ -0,0 +1,48 @@
...
+	Amount       common.Amount `json:"amount"`
+	VotingResult string        `json:"voting-result"`
+}
+
+func NewInflationPF(
+	amount common.Amount,
+	votingResult string,
```
spikeekips: `votingResult` has the operation hash of `CongressVotingResult`, right? The description in comment will be helpful.

lib/node/runner/checker_ballot_transaction.go
```diff
@@ -346,12 +346,30 @@ func ValidateOp(st *storage.LevelDBBackend, source *block.BlockAccount, op opera
 		if bo.Type == operation.TypeUnfreezingRequest {
 			return errors.UnfreezingRequestAlreadyReceived
 		}
+	case operation.TypeInflationPF:
+		var ok bool
+		var casted operation.InflationPF
+		if casted, ok = op.B.(operation.InflationPF); !ok {
+			return errors.TypeOperationBodyNotMatched
+		}
+		exists, err := block.ExistsBlockOperation(st, casted.VotingResult)
```
spikeekips: 👍

soonkuk:
@kfangw PF inflation should be sent to the PF budget account.
I think target of inflation is PF budget account.

kfangw:
@soonkuk why?
And what PF budget is?
Who handles the PF budget account?
Isn't it enough to keep common budget only?

spikeekips:
> @soonkuk : @kfangw PF inflation should be sent to the PF budget account. I think target of inflation is PF budget account.

I never heard about PF budget account? What is it?

spikeekips:
@soonkuk @kfangw The scenario will be like this,

1. `CongressVoting` operation has the address, which receives the PF inflation
1. After voting passes, PF inflation operation sends the inflation to the address of that `CongressVoting` operation.

So, in PF inflation does not need to set the target address, it should be relied on `CongressVoting` operation.

kfangw:
> @soonkuk @kfangw The scenario will be like this,
> 
> 1. `CongressVoting` operation has the address, which receives the PF inflation
> 2. After voting passes, PF inflation operation sends the inflation to the address of that `CongressVoting` operation.
> 
> So, in PF inflation does not need to set the target address, it should be relied on `CongressVoting` operation.

oh. It is a new idea. looks better.

Hence, the receiver is set at the beginning of the voting right?
I will work on it.

kfangw:
@spikeekips @soonkuk Please review again
The changes, since your last review, are to support the scenario what @spikeekips mentioned above mainly.

lib/transaction/operation/operation_inflation_pf.go
kfangw: target required.

lib/transaction/operation/congress_voting_result.go
```diff
@@ -23,12 +23,14 @@ type CongressVotingResult struct {
 		No    uint64 `json:"no"`
 		ABS   uint64 `json:"abs"`
 	} `json:"result"`
+	CongressVotingHash string `json:"congress_voting_hash"`
```
kfangw: @outsideris CongressVoting and CongressVotingResult operations will be changed by this PR.

lib/transaction/operation/congress_voting.go
```diff
@@ -13,9 +13,11 @@ type CongressVoting struct {
 		Start uint64 `json:"start"`
 		End   uint64 `json:"end"`
 	} `json:"voting"`
+	FundingAddress string        `json:"funding_address"`
```
kfangw: @outsideris CongressVoting and CongressVotingResult operations will be changed by this PR.

kfangw:
@spikeekips PING~

spikeekips:
:+1:

lib/node/runner/checker_ballot_transaction_test.go
```diff
@@ -51,15 +51,15 @@ func TestValidateTxPaymentMissingBlockAccount(t *testing.T) {
 		},
 	}
 	tx.H.Hash = tx.B.MakeHashString()
-	require.Equal(t, ValidateTx(st, tx), errors.BlockAccountDoesNotExists)
+	require.Equal(t, ValidateTx(st, common.Config{}, tx), errors.BlockAccountDoesNotExists)
```
Geod24: We have `common.NewTestConfig()` for this

kfangw: I will make PR according to this. Use context.

PR #721, opened 2018-11-14
Support CIDR in rate-limit

spikeekips:
### Background

Simply to say, currently rate limit supports the rule by ip address. This patch will be able to by cidr, ip address range. The current testnet env does not indicate the exact ip address to angelbot, it can just provide ip range.

### DIFF
--- a/lib/network/http_middlewares.go
+++ b/lib/network/http_middlewares.go
@@ -2,6 +2,7 @@ package network
 
 import (
 	"fmt"
+	"net"
 	"net/http"
 	"runtime/debug"
 	"strconv"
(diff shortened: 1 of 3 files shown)


lib/network/http_middlewares.go
```diff
@@ -86,9 +90,18 @@ func RateLimitMiddleware(logger logging.Logger, rule common.RateLimitRule) mux.M
...
+		key := common.GetUniqueIDFromUUID()
+		middlewares[key] = m
+		middlewaresByIP[ip] = key
+
+		if _, ipnet, err := net.ParseCIDR(ip); err == nil {
+			byCIDRs[ip] = ipnet
+		}
 	}
 
+	middlewareCache, _ := lru.New(40000)
```
Charleslee522: I have a question. Where is this cache size(40000) from?

spikeekips: I don't have exact and proper value yet, but I guess `40000` will be enough, `40kb`.

PR #723, opened 2018-11-14
Failed to save `BlockOperation`

spikeekips:
### Background
Currently to save operations the several goroutines are used and transactions are injected by channel. The problem is the number of transaction is over channel buffer, the saving process is stuck.

### Solution

The part of inserting transaction set inside goroutine and fixed the other bugs to save `BlockOperation`.

### DIFF
--- a/lib/block/transaction.go
+++ b/lib/block/transaction.go
@@ -190,19 +190,36 @@ func (bt *BlockTransaction) SaveBlockOperations(st *storage.LevelDBBackend) (err
 	}
 
 	for _, op := range bt.Transaction().B.Operations {
-		var bo BlockOperation
-		bo, err = NewBlockOperationFromOperation(op, bt.Transaction(), bt.blockHeight)
+		if err = bt.SaveBlockOperation(st, op); err != nil {
+			return
+		}
+	}
+
+	return nil
+}
+
+func (bt *BlockTransaction) SaveBlockOperation(st *storage.LevelDBBackend, op operation.Operation) (err error) {
+	if bt.blockHeight < 1 {
+		var blk Block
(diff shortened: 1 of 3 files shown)


lib/block/transaction.go
```diff
@@ -190,19 +190,36 @@ func (bt *BlockTransaction) SaveBlockOperations(st *storage.LevelDBBackend) (err
...
 			return
 		}
-		if err = bo.Save(st); err != nil {
+	}
+
+	return nil
+}
+
+func (bt *BlockTransaction) SaveBlockOperation(st *storage.LevelDBBackend, op operation.Operation) (err error) {
+	if bt.blockHeight < 1 {
```
Charleslee522: @spikeekips This part is included `other bugs`, right?

lib/node/runner/block_operations.go
```diff
@@ -134,10 +134,12 @@ func (sb *SavingBlockOperations) CheckByBlock(st *storage.LevelDBBackend, blk bl
...
-	for _, hash := range blk.Transactions {
-		txs <- hash
-	}
-	close(txs)
+	go func() {
+		for _, hash := range blk.Transactions {
+			txs <- hash
+		}
+		close(txs)
+	}()
```
Charleslee522: This part might be main 👍

lib/block/transaction.go
spikeekips: @Charleslee522 No, I separated `SaveBlockOperation` from `SaveBlockOperations`, so this part is from `SaveBlockOperations`.

Charleslee522: @spikeekips Thanks :)

Issue #733, opened 2018-11-15
Amount.EncodeRLP is wrong.


zzim2x:
#730 

I am on my way home. Zzz.

Currently, Amount(10000) is encoded as 0x3130303030.
It should be encoded as 0x853130303030.

0x85 indicates a 5-byte string.

kfangw:
How did you tested it?

kfangw:
a.String() returns 0x3130303030. It is normal.
But when rlp encode byte array should be 0x853130303030

zzim2x:
rlp.Encode(w, a.String())

kfangw:
Lets figure out the reason tomorrow.

anarcher:
Close it :->

PR #736, opened 2018-11-16
fix amount rlp

kfangw:
### Github Issue
#733 

### Background
EncodeRLP expected to return encoded byte array, not to be encoded byte array


outsideris:
It works with desktop app.

lib/common/amount_test.go
```diff
@@ -74,3 +76,22 @@ func TestAmount_Uint64OutOfRange(t *testing.T) {
 		}
 	}
 }
+
+func TestRLPEncoding(t *testing.T) {
```
zzim2x: LGTM

PR #744, opened 2018-11-16
Txpool Limit 

### LINKED ISSUE #588: Transaction Limit For `TransactionPool`
Currently there are no limit for `NodeRunner.TransactionPool`, it totally depends on the os resource limit. 'No Limit' can exhaust memory resource, some oses like linux(OOM Killer) and osx will kill the sebak process.

* We need to add new limit for `TransactionPool`
* If over limit, the incoming transaction will get the proper HTTP status code and proper `HTTPProblem` message.

anarcher:
### Github Issue

Fixes  #588

### Background

I think that Both of client-side or node-side limiting for txpool is needed.  I think that if node flag (`--txpool-limit`) handles this limits, it works well.  If a node has this pool of it is full, consensus works with requesting missing txs routine. 

My idea is that both client-side and node-side has different limit numbers, but pol size is same.  
If the pool size is 1000 and client-limit is 200, the client can not add this txs from this client, but if node-limit is 2000, the node can do it. 

### Solution

- [x] Support client-side limit and node-side limit
- [x] Change to `Pool.Add(tx,limit...)`
- [x] Add node flags for txpool limit e.g. `sebak node --txpool-limit=100,100`
- [ ] Test

### DIFF
--- a/cmd/sebak/cmd/run.go
+++ b/cmd/sebak/cmd/run.go
@@ -368,6 +368,9 @@ func parseFlagsNode() {
 	// tx pool limits (client,node)
 	{
 		limits := strings.Split(flagTxPoolLimit, ",")
+		if len(limits) > 2 {
+			cmdcommon.PrintFlagsError(nodeCmd, "--txpool-limit", fmt.Errorf("wrong format: format:<client-limit>[,<node-limit]"))
+		}
 	L:
 		for i, l := range limits {
 			switch i {
--- a/cmd/sebak/cmd/run.go
+++ b/cmd/sebak/cmd/run.go
@@ -101,7 +101,8 @@ var (
 	httpCacheAdapter    string
 	httpCachePoolSize   int
 	httpCacheRedisAddrs map[string]string
-	txPoolLimit         uint64
+	txPoolClientLimit   uint64
+	txPoolNodeLimit     uint64
 
 	logLevel logging.Lvl
 	log      logging.Logger = logging.New("module", "main")
@@ -182,7 +183,7 @@ func init() {
 	nodeCmd.Flags().StringVar(&flagTransactionsLimit, "transactions-limit", flagTransactionsLimit, "transactions limit in a ballot")
 	nodeCmd.Flags().StringVar(&flagUnfreezingPeriod, "unfreezing-period", flagUnfreezingPeriod, "how long freezing must last")
 	nodeCmd.Flags().StringVar(&flagOperationsLimit, "operations-limit", flagOperationsLimit, "operations limit in a transaction")
-	nodeCmd.Flags().StringVar(&flagTxPoolLimit, "txpool-limit", flagTxPoolLimit, "toperation pool limit")
+	nodeCmd.Flags().StringVar(&flagTxPoolLimit, "txpool-limit", flagTxPoolLimit, "transaction pool limit: <client-side>[,<node-side>]")
 	nodeCmd.Flags().Var(
 		&flagRateLimitAPI,
 		"rate-limit-api",
@@ -364,8 +365,24 @@ func parseFlagsNode() {
 		threshold = int(tmpUint64)
 	}
 
-	if txPoolLimit, err = strconv.ParseUint(flagTxPoolLimit, 10, 64); err != nil {
-		cmdcommon.PrintFlagsError(nodeCmd, "--txpool-limit", err)
+	// tx pool limits (client,node)
+	{
+		limits := strings.Split(flagTxPoolLimit, ",")
+	L:
+		for i, l := range limits {
+			switch i {
+			case 0:
+				if txPoolClientLimit, err = strconv.ParseUint(l, 10, 64); err != nil {
+					cmdcommon.PrintFlagsError(nodeCmd, "--txpool-limit", err)
+				}
+			case 1:
+				if txPoolNodeLimit, err = strconv.ParseUint(l, 10, 64); err != nil {
+					cmdcommon.PrintFlagsError(nodeCmd, "--txpool-limit", err)
+				}
+			default:
+				break L
+			}
+		}
 	}
 
 	if common.UnfreezingPeriod, err = strconv.ParseUint(flagUnfreezingPeriod, 10, 64); err != nil {
@@ -596,7 +613,8 @@ func runNode() error {
 		HTTPCachePoolSize:      httpCachePoolSize,
 		HTTPCacheRedisAddrs:    httpCacheRedisAddrs,
 		CongressAccountAddress: flagCongressAddress,
-		TxPoolLimit:            int(txPoolLimit),
+		TxPoolClientLimit:      int(txPoolClientLimit),
+		TxPoolNodeLimit:        int(txPoolNodeLimit),
 	}
 	st, err := storage.NewStorage(storageConfig)
 	if err != nil {
--- a/cmd/sebak/cmd/run.go
+++ b/cmd/sebak/cmd/run.go
@@ -72,7 +72,7 @@ var (
 	flagHTTPCachePoolSize   string = common.GetENVValue("SEBAK_HTTP_CACHE_POOL_SIZE", "10000")
 	flagHTTPCacheRedisAddrs string = common.GetENVValue("SEBAK_HTTP_CACHE_REDIS_ADDRS", "")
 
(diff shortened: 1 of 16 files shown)


anarcher:
About node flags interface, 

``` sebak node --txpool-limit=100 ``` means  client-side=100 , node-side=inf?
``` sebak node --txpool-limit=100,200``` means client-side=100,node-side=200

OR

```sebak node --txpool-client-limit=100 --txpol-node-limit=200```

File: lib/common/config.go [hunk 1]
```diff
@@ -43,8 +44,11 @@ func NewConfig(networkID []byte) Config {
 
 	p.TxsLimit = 1000
 	p.OpsLimit = 1000
+
 	p.NetworkID = networkID
 
+	p.TxPoolLimit = 100000
```
spikeekips: The default value, `DefaultTxPoolLimit` can be used here.

lib/transaction/pool.go
```diff
@@ -13,14 +16,20 @@ type Pool struct {
 
 	hashList *list.List // Transaction.GetHash()
 	hashMap  map[ /* Transaction.GetHash() */ string]*list.Element
+
+	limit int
 }
 
-func NewPool() *Pool {
+func NewPool(limit int) *Pool {
+	if limit <= 0 {
```
spikeekips: How about, if `limit` is under `0`, make `Pool` to unlimited?

lib/transaction/pool.go
```diff
@@ -57,22 +66,26 @@ func (tp *Pool) GetFromSource(source string) (Transaction, bool) {
...
 	txHash := tx.GetHash()
 	if tp.Has(txHash) {
-		return false
+		return errors.TransactionAlreadyExistsInPool
 	}
 
 	tp.Lock()
 	defer tp.Unlock()
 
+	if len(tp.Pool) >= tp.limit {
```
spikeekips: If we use lock strictly, the upper locking part should be moved under this part.

lib/common/config.go
```diff
@@ -17,8 +17,9 @@ type Config struct {
 	BlockTime         time.Duration
 	BlockTimeDelta    time.Duration
 
-	TxsLimit int
-	OpsLimit int
+	TxsLimit    int
+	OpsLimit    int
+	TxPoolLimit int
```
spikeekips: In `run.go`, `txPoolLimit` is `unit64`. How about set `TxPoolLimit` to be same, `uint64` instead of `int`?

anarcher:
> For client side, it would be better to mention in the description about when the tx pool limit is reached. The client will get the some HTTPProblem message, something like this.

I think that  the handlers return `errors.TransactionPoolFull` with `HTTPProblem` format is OK?

lib/transaction/pool.go
anarcher: The check `len(tp.Pool)` in the lock is more strictly. Is it not right?

spikeekips:
> @anarcher : I think that  the handlers return `errors.TransactionPoolFull` with `HTTPProblem` format is OK? 

I agree and one more thing, how about set the specific status code like `423` (Locked) https://developer.mozilla.org/en-US/docs/Web/HTTP/Status .

lib/common/config.go
anarcher: I want to use `int`,  because of `pool.Len() int`. And lrucache use  the size  as `int`. https://godoc.org/github.com/hashicorp/golang-lru#New

anarcher:
Plz review it @spikeekips @Charleslee522 @kfangw @Geod24

cmd/sebak/cmd/run.go
```diff
@@ -70,11 +70,12 @@ var (
 	flagHTTPCacheAdapter    string = common.GetENVValue("SEBAK_HTTP_CACHE_ADAPTER", "")
 	flagHTTPCachePoolSize   string = common.GetENVValue("SEBAK_HTTP_CACHE_POOL_SIZE", "10000")
 	flagHTTPCacheRedisAddrs string = common.GetENVValue("SEBAK_HTTP_CACHE_REDIS_ADDRS", "")
+
+	flagTxPoolLimit string = common.GetENVValue("SEBAK_TX_POOL_LIMIT", "100000")
```
spikeekips: `DefaultTxPoolLimit ` will be used for the default value.

Geod24:
> I agree and one more thing, how about set the specific status code like 423 (Locked) 

Locked is a WebDAV code. Also it's not quite a client error. For me `503 Service Unavailable` would be more suitable.

spikeekips:
@Geod24 423 is not best one, and 503 is also too, it looks like entire service not available. :) For this case we can define new one, “450 Transaction Pool Limit Reached”

cmd/sebak/cmd/run.go
```diff
@@ -360,6 +364,26 @@ func parseFlagsNode() {
...
+			case 0:
+				if txPoolClientLimit, err = strconv.ParseUint(l, 10, 64); err != nil {
+					cmdcommon.PrintFlagsError(nodeCmd, "--txpool-limit", err)
+				}
+			case 1:
+				if txPoolNodeLimit, err = strconv.ParseUint(l, 10, 64); err != nil {
+					cmdcommon.PrintFlagsError(nodeCmd, "--txpool-limit", err)
+				}
+			default:
+				break L
```
Geod24: I think we should issue error if the user provides more than 2 args

cmd/sebak/cmd/run.go
```diff
@@ -179,6 +182,7 @@ func init() {
 	nodeCmd.Flags().StringVar(&flagTransactionsLimit, "transactions-limit", flagTransactionsLimit, "transactions limit in a ballot")
 	nodeCmd.Flags().StringVar(&flagUnfreezingPeriod, "unfreezing-period", flagUnfreezingPeriod, "how long freezing must last")
 	nodeCmd.Flags().StringVar(&flagOperationsLimit, "operations-limit", flagOperationsLimit, "operations limit in a transaction")
+	nodeCmd.Flags().StringVar(&flagTxPoolLimit, "txpool-limit", flagTxPoolLimit, "transaction pool limit: <client-side>[,<node-side>]")
```
Geod24: It should mention that `(0=no limit)`

cmd/sebak/cmd/run.go
anarcher: @Geod24 What kinds of issue?

cmd/sebak/cmd/run.go
anarcher: I added the mention.

anarcher:
> IMHO the user of the Pool.Add method should not be the one providing the limit.

@Geod24 Ok, Do you agree that `Pool.AddFromClient` and `Pool.AddFromNode`?

lib/node/runner/checker_message.go
```diff
@@ -128,19 +128,60 @@ func MessageValidate(c common.Checker, args ...interface{}) (err error) {
 
 // PushIntoTransactionPool add the incoming
 // transactions into `Pool`.
-func PushIntoTransactionPool(c common.Checker, args ...interface{}) (err error) {
+func PushIntoTransactionPool(c common.Checker, args ...interface{}) error {
 	checker := c.(*MessageChecker)
 
 	tx := checker.Transaction
-	checker.TransactionPool.Add(tx)
+	err := checker.TransactionPool.Add(tx, 0) // 0 is unlimited
```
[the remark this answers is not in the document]
anarcher: IMHO, If Pool doesn't know that client-side, It means that don't make dependency.  But if we agree that `AddFromXXX`, I have a mind to change to it. :->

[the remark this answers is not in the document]
anarcher: I changed it @Charleslee522

cmd/sebak/cmd/run.go
Geod24: "issue" as a verb means "to give" or "to make".
So: "I think we should really have an error instead of the `default`"

anarcher: Oh, I got it.

anarcher: `error: invalid '--txpool-limit'; wrong format: format:<client-limit>[,<node-limit]` 
How about it? @Geod24

Geod24: Sounds good

Issue #595, opened 2018-10-25
Prevent direct recursive broadcasting


Charleslee522:
from https://github.com/bosnet/sebak/issues/592#issuecomment-432553837

Charleslee522:
If a broadcasting target is ballot, a node have to send ballot to sender when exceeding threshold because sender also should know my ballot state.
If it is the transaction, no more recursive braodcasting because of #641.
@spikeekips Therefore this issue is resolved, am I right? :)

Charleslee522:
resolved in #745 and #805

PR #745, opened 2018-11-16
Skip broadcasting self-signed Ballot

spikeekips:
### Github Issue

Slightly Related to #595 

### Background
Currently the node will broadcast ballot to the validators, including itself. This patch will skip to broadcast to node itself.

### Solution
This patch will reduce the broadcasting time and consensus time.

Old:
```
Broadcasting Ballot -> HTTP2 `/node/message` -> `NodeRunner.HandleBallot()`
```

New:

```
Broadcasting Ballot -> `NodeRunner.HandleBallot()`
```

### DIFF
--- a/lib/node/runner/node_runner.go
+++ b/lib/node/runner/node_runner.go
@@ -372,7 +372,6 @@ func (nr *NodeRunner) Start() (err error) {
 func (nr *NodeRunner) Stop() {
 	nr.network.Stop()
 	nr.isaacStateManager.Stop()
-	close(nr.ballotBroadcastChannel)
 }
 
 func (nr *NodeRunner) Node() *node.LocalNode {
@@ -671,7 +670,6 @@ func (nr *NodeRunner) proposeNewBallot(round uint64) (ballot.Ballot, error) {
 
 	nr.log.Debug("new ballot created", "ballot", theBallot)
 
-	//nr.ConnectionManager().Broadcast(*theBallot)
 	nr.broadcastBallot(*theBallot)
 
 	return *theBallot, nil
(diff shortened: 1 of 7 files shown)


spikeekips:
I will also test in real world.

lib/node/runner/node_runner.go
```diff
@@ -651,11 +663,35 @@ func (nr *NodeRunner) proposeNewBallot(round uint64) (ballot.Ballot, error) {
...
+
+func (nr *NodeRunner) BroadcastBallot(b ballot.Ballot) {
+	go func() {
+		nr.ballotBroadcastChannel <- b
+	}()
+}
+
+func (nr *NodeRunner) broadcastBallot(b ballot.Ballot) {
+	go func() {
+		nr.handleBallotMessage(b)
```
anarcher: I have a question:  Why node do `broadcastBallot` and `handleBallotMessage` at the same time? :->  It means that `b ballot.Ballot` is not from node itself?

spikeekips:
@Charleslee522 Please leave your review :)

lib/node/runner/node_runner.go
```diff
@@ -470,22 +476,30 @@ func (nr *NodeRunner) handleMessage(message common.NetworkMessage) {
...
 		Log:                nr.Log(),
 		VotingHole:         voting.NOTYET,
 		LatestBlockSources: []string{},
 	}
+
+	if message, ok := msg.(common.NetworkMessage); ok {
+		nr.log.Debug("got ballot")
+		baseChecker.Message = message
+	} else if ball, ok := msg.(ballot.Ballot); ok {
+		baseChecker.Ballot = ball
```
Charleslee522: This part is ballot from itself, right? How about logging here, as well?

PR #753, opened 2018-11-19
Add MaxLimitExceed check

anarcher:
### Background

To protect resources, Add paging MaxLimitExceed checking.  (DefaultMaxLimit=1000)

### Solution

The `PageQuery` can check this limit parameter, all handlers to use this `PageQuery` do check limit automatically.

### DIFF
--- a/lib/errors/errors.go
+++ b/lib/errors/errors.go
@@ -102,5 +102,5 @@ var (
 	InflationPFAmountMissMatched              = NewError(184, "inflation pf amount missmatched")
 	InflationPFFundingAddressMissMatched      = NewError(185, "inflation pf funding address missmatched")
 	CongressAddressMisMatched                 = NewError(186, "congress address mismatched")
-	MaxLimitExceed                            = NewError(187, "max limit exceed")
+	PageQueryLimitMaxExceed                   = NewError(187, "maximum value of limit query parameter exceed")
 )
--- a/lib/errors/errors.go
+++ b/lib/errors/errors.go
@@ -102,4 +102,5 @@ var (
 	InflationPFAmountMissMatched              = NewError(184, "inflation pf amount missmatched")
 	InflationPFFundingAddressMissMatched      = NewError(185, "inflation pf funding address missmatched")
 	CongressAddressMisMatched                 = NewError(186, "congress address mismatched")
+	MaxLimitExceed                            = NewError(187, "max limit exceed")
 )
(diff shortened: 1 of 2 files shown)


lib/node/runner/api/pagequery.go
```diff
@@ -7,11 +7,15 @@ import (
...
 	"boscoin.io/sebak/lib/common"
+	"boscoin.io/sebak/lib/errors"
 	"boscoin.io/sebak/lib/node/runner/api/resource"
 	"boscoin.io/sebak/lib/storage"
 )
 
-const DefaultMaxLimit uint64 = 100
+const (
+	DefaultLimit    uint64 = 100
+	DefaultMaxLimit uint64 = 1000
```
spikeekips: `1000` is too big number, I think. `DefaultLimit` is `20` and `DefaultMaxLimit` to `100`, how do you think?

anarcher: I changed it! :->

anarcher:
Plz review it @kfangw @cmcm2222

lib/node/runner/api/pagequery.go
```diff
@@ -7,11 +7,15 @@ import (
...
 	"boscoin.io/sebak/lib/common"
+	"boscoin.io/sebak/lib/errors"
 	"boscoin.io/sebak/lib/node/runner/api/resource"
 	"boscoin.io/sebak/lib/storage"
 )
 
-const DefaultMaxLimit uint64 = 100
+const (
+	DefaultLimit    uint64 = 20
+	DefaultMaxLimit uint64 = 100
```
Geod24: The naming here is inconsistent.
One is the default value for `limit` (`DefaultLimit`), the other is the maximum value for `limit` (`DefaultMaxLimit`).
Since there is no `Default`, it should probably be something like `MaxLimit`.
Also, since `limit` is ambiguous, you could say something like `LimitValue` (so `MaxLimitValue`, `DefaultLimitValue`) to make it more clear it's referring to the parameter name.

lib/errors/errors.go
```diff
@@ -102,4 +102,5 @@ var (
 	InflationPFAmountMissMatched              = NewError(184, "inflation pf amount missmatched")
 	InflationPFFundingAddressMissMatched      = NewError(185, "inflation pf funding address missmatched")
 	CongressAddressMisMatched                 = NewError(186, "congress address mismatched")
+	MaxLimitExceed                            = NewError(187, "max limit exceed")
```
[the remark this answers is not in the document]
anarcher: Thanks! @Geod24

lib/node/runner/api/pagequery.go
anarcher: Ok, I got it.

anarcher:
IMO, limit is 100 is good enough.  To short limit make a lot of request.

PR #756, opened 2018-11-19
Change CongressVoting's Contract field type from []byte to string

kfangw:
### Background
To improve the usability of client-side,
[]byte is changed to string

### DIFF
--- a/lib/client/response.go
+++ b/lib/client/response.go
@@ -133,7 +133,7 @@ type OperationsPage struct {
 }
 
 type CongressVoting struct {
-	Contract []byte `json:"contract"`
+	Contract string `json:"contract"`
 	Voting   struct {
 		Start uint64 `json:"start"`
 		End   uint64 `json:"end"`
--- a/lib/transaction/operation/congress_voting.go
+++ b/lib/transaction/operation/congress_voting.go
@@ -8,7 +8,7 @@ import (
 )
 
 type CongressVoting struct {
-	Contract []byte `json:"contract"`
+	Contract string `json:"contract"`
 	Voting   struct {
 		Start uint64 `json:"start"`
 		End   uint64 `json:"end"`
@@ -17,7 +17,7 @@ type CongressVoting struct {
 	Amount         common.Amount `json:"amount"`
 }
 
-func NewCongressVoting(contract []byte, start, end uint64, amount common.Amount, fundingAddress string) CongressVoting {
+func NewCongressVoting(contract string, start, end uint64, amount common.Amount, fundingAddress string) CongressVoting {
 
 	return CongressVoting{
 		Contract: contract,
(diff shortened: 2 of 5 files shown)


zzim2x:
LGTM!!

Geod24:
1) Would be better to give a more useful name to the commit / PR. E.g. "Change CongressVoting.Contract type to string". We have [guidelines](https://github.com/bosnet/sebak/blob/13dbe0a17626858b51b61b4dde078554b5597458/CONTRIBUTING.md#pull-request) which reference [7 great rules to write a commit message](https://chris.beams.io/posts/git-commit/)
2) What is the content of `Contract` supposed to be ? I thought that it would be bytecode ?

outsideris:
It passed signature validation.

The response is `submitted` when I send a tx for congress voting

```
{ _links:
   { history:
      { href:
         '/api/v1/transactions/ADKdkaDm1Qcy3Jy7w4YzuyiZ6CwabN9BKdaqBdDNTaXL/history' },
     self: { href: '/api/v1/transactions' } },
  hash: 'ADKdkaDm1Qcy3Jy7w4YzuyiZ6CwabN9BKdaqBdDNTaXL',
  message:
   { source: 'GCT67AX6PSARM47MJYAEMAQTS4LQRJDJFSMBAEEKHVCDIP5VRC3P2RPC',
     fee: '10000',
     sequence_id: 0,
     operations: [ [Object] ] },
  status: 'submitted' }
```

However, the transaction status is `rejected`.

```
curl --insecure https://localhost:12345/api/v1/transactions/ADKdkaDm1Qcy3Jy7w4YzuyiZ6CwabN9BKdaqBdDNTaXL/history | jq .
{
  "_links": {
    "account": {
      "href": "/api/v1/accounts/GCT67AX6PSARM47MJYAEMAQTS4LQRJDJFSMBAEEKHVCDIP5VRC3P2RPC"
    },
    "self": {
      "href": "/api/v1/transactions"
    },
    "transaction": {
      "href": "/api/v1/transactions/ADKdkaDm1Qcy3Jy7w4YzuyiZ6CwabN9BKdaqBdDNTaXL"
    }
  },
  "confirmed": "2018-11-19T16:14:32.748815000+09:00",
  "created": "2018-11-19T16:14:32.735000000+09:00",
  "hash": "ADKdkaDm1Qcy3Jy7w4YzuyiZ6CwabN9BKdaqBdDNTaXL",
  "source": "GCT67AX6PSARM47MJYAEMAQTS4LQRJDJFSMBAEEKHVCDIP5VRC3P2RPC",
  "status": "rejected"
}
```

Geod24:
@outsideris : The question is: "Is that field supposed to be string, or byte code?".
If it's supposed to be string, it LGTM. But I thought it was supposed to be bytecode and we just use strings for testing.

outsideris:
> For transfer that's fine, but if we store it this way, it means a 33% overhead.

In this aspect, the interface for RLP should be string at least. We can't separate both things?

Geod24:
I would like to have them separated, but at the moment we don't (e.g. we save in JSON)

outsideris:
I think we can merge this because `rejected` case is open in #757 
And separating RLP data type and database data type can be handled in another issue. 
How do you think? @Geod24

outsideris:
Tests should be passed, of course. :)

Issue #680, opened 2018-11-08
Txs or operations are not indexed when someone received a tx.


zzim2x:
Payable operations such as payment, create-account, inflation, collect-tx-fee, etc should be indexed.

Any payable operations should exist on a list of operations of the target account.
Any transactions having payable operations should exist on a list of transactions of target account.

Charleslee522:
Does anyone follow this issue?

Charleslee522:
#4 can be resolved together.

kfangw:
depends on #765 #766

spikeekips:
This issue was resolved, right?

kfangw:
solved by #763

PR #763, opened 2018-11-19
operation index for target also

kfangw:
### Github Issue
#680 

### Background
Current `/accounts/{id}/operations` API returns operations which are sent by source only
The client such as wallet and Explorer requires that the API returns operations which are related not only source but also the target.

### Solution
* Add target index for block operation
* Add peers index for block operation

* query operations for peers index


kfangw:
when I added testcases, description will be filled, and I will request the review.

kfangw:
This PR depends on #765 and #766.
Hence, this will be resumed when those are merged.

kfangw:
@anarcher review please ~~

anarcher:
Can you rebase from #765 ?  This PR is merged from #765. I think that It merged after  #765 is merged. 
What do you think this and that (#765) ? @kfangw @spikeekips @Charleslee522 @Geod24

kfangw:
I rebased it on master not #765 
Could you review again? @anarcher

spikeekips:
Too late, 👍

PR #766, opened 2018-11-20
The item of `/api/v1/transactions/<hash>/operations` will have `Confirmed` time.

### LINKED ISSUE #751: Include the creation date in the operation response data
Right now the only way to obtain the creation date of an operation, is to make a separate call to the correlated transaction. It would be handy if the creation date would be included in the operation response data. This allows third parties to easily obtain all related operation information in one request.

Example request
https://testnet-sebak.blockchainos.org/api/v1/transactions/ANF9WSAyUnK2Y7cYdW5AnNL2brZk1uGfdFNjseqSyv7Z/operations

spikeekips:
### Github Issue

Resolves #751 

### Background

See #751 , toperation api will

### Solution

Simply `Block.Confirmed` time and `BlockOperation.block_height` in `resource.Operation`.

* `confirmed` time can be considered the creation or stored time of the block, transactions and operations, this value can be different by nodes.
* `proposed_time` is the proposed time by proposer, this must be same between nodes.
* `block_height` is the height of block.

The output will be,
```json
{
  "_embedded": {
    "records": [
      {
        "_links": {
          "self": {
            "href": "/api/v1/operations/HiM4q3P82HuByuqvjGuyj8CxUdK5jk52EZ4YgATafSLr-8N43ibLQLABnYc2PpqbDzBWUBfF69Z5WTFSkf7eChqqw"
          },
          "transaction": {
            "href": "/api/v1/transactions/8N43ibLQLABnYc2PpqbDzBWUBfF69Z5WTFSkf7eChqqw"
          }
        },
        "block_height": 2,
        "body": {
          "target": "GCDPYNLZRGRZEJNDHJMKP45QKT4ZNZWP4ZAWFNNGDT4TG7TJQ73DZINW",
          "amount": "3081"
        },
        "confirmed": "2018-11-20T14:47:42.552511000+09:00",
        "proposed_time": "2018-11-20T14:47:40.152513000+09:00",
        "hash": "HiM4q3P82HuByuqvjGuyj8CxUdK5jk52EZ4YgATafSLr-8N43ibLQLABnYc2PpqbDzBWUBfF69Z5WTFSkf7eChqqw",
        "source": "GCHKDI3GZBSME6SFHNBGI2RS3NM6AD5I5VM522CXKE6YBBDDZB6UPZ2D",
        "tx_hash": "8N43ibLQLABnYc2PpqbDzBWUBfF69Z5WTFSkf7eChqqw",
        "type": "payment"
      }
    ]
  },
  "_links": {
    "next": {
      "href": "/transactions/8N43ibLQLABnYc2PpqbDzBWUBfF69Z5WTFSkf7eChqqw/operations?limit=20&reverse=false"
    },
    "prev": {
      "href": "/transactions/8N43ibLQLABnYc2PpqbDzBWUBfF69Z5WTFSkf7eChqqw/operations?limit=20&reverse=true"
    },
    "self": {
      "href": "/transactions/8N43ibLQLABnYc2PpqbDzBWUBfF69Z5WTFSkf7eChqqw/operations"
    }
  }
}
```

### DIFF
--- a/lib/node/runner/api/tx_operations.go
+++ b/lib/node/runner/api/tx_operations.go
@@ -1,8 +1,11 @@
 package api
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
+
+	"github.com/gorilla/mux"
 
 	"boscoin.io/sebak/lib/block"
 	"boscoin.io/sebak/lib/common/observer"
@@ -10,7 +13,6 @@ import (
 	"boscoin.io/sebak/lib/network/httputils"
 	"boscoin.io/sebak/lib/node/runner/api/resource"
 	"boscoin.io/sebak/lib/storage"
-	"github.com/gorilla/mux"
 )
 
 func (api NetworkHandlerAPI) GetOperationsByTxHashHandler(w http.ResponseWriter, r *http.Request) {
@@ -31,17 +33,51 @@ func (api NetworkHandlerAPI) GetOperationsByTxHashHandler(w http.ResponseWriter,
 	options := p.ListOptions()
 
 	if httputils.IsEventStream(r) {
-		event := fmt.Sprintf("txhash-%s", hash)
-		es := NewEventStream(w, r, renderEventStream, DefaultContentType)
-		ops, _ := api.getOperationsByTxHash(hash, options)
-		for _, op := range ops {
-			es.Render(op)
+		var blk *block.Block
+		es := NewEventStream(
+			w,
+			r,
+			func(args ...interface{}) ([]byte, error) {
+				if len(args) <= 1 {
+					return nil, fmt.Errorf("render: value is empty")
+				}
+				i := args[1]
+
+				if i == nil {
+					return nil, nil
+				}
+
+				if blk == nil {
+					if blk, err = api.getBlockByTxHash(hash); err != nil {
+						return nil, err
+					}
+				}
+
+				r := resource.NewOperation(i.(*block.BlockOperation))
+				r.Block = blk
+				return json.Marshal(r.Resource())
+			},
+			DefaultContentType,
+		)
+
+		var err error
+		if blk, err = api.getBlockByTxHash(hash); err == nil {
+			ops, _ := api.getOperationsByTxHash(hash, blk, options)
+			for _, op := range ops {
+				es.Render(op)
+			}
 		}
-		es.Run(observer.BlockOperationObserver, event)
+		es.Run(observer.BlockOperationObserver, fmt.Sprintf("txhash-%s", hash))
 		return
 	}
 
-	ops, cursor := api.getOperationsByTxHash(hash, options)
+	var blk *block.Block
+	if blk, err = api.getBlockByTxHash(hash); err != nil {
+		httputils.WriteJSONError(w, err)
+		return
+	}
+
+	ops, cursor := api.getOperationsByTxHash(hash, blk, options)
 	if len(ops) < 1 {
 		httputils.WriteJSONError(w, errors.BlockTransactionDoesNotExists)
 		return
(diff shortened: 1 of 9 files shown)


kfangw:
@spikeekips `/accounts/{id}/operations` needs this also.

lib/node/runner/api/tx_operations.go
```diff
@@ -31,17 +33,51 @@ func (api NetworkHandlerAPI) GetOperationsByTxHashHandler(w http.ResponseWriter,
...
+			}
 		}
-		es.Run(observer.BlockOperationObserver, event)
+		es.Run(observer.BlockOperationObserver, fmt.Sprintf("txhash-%s", hash))
 		return
 	}
 
-	ops, cursor := api.getOperationsByTxHash(hash, options)
+	var blk *block.Block
+	if blk, err = api.getBlockByTxHash(hash); err != nil {
```
anarcher: How about that `blk, err = api.getBlockByTxHash` to line 35. 
EventStream and normal rendering can use it because of this tx hashes has same block.

spikeekips: @anarcher The event stream should work if the block and operations is not saved yet.

lib/node/runner/api/tx_operations.go
```diff
@@ -31,17 +33,51 @@ func (api NetworkHandlerAPI) GetOperationsByTxHashHandler(w http.ResponseWriter,
...
+					return nil, nil
+				}
+
+				if blk == nil {
+					if blk, err = api.getBlockByTxHash(hash); err != nil {
+						return nil, err
+					}
+				}
+
+				r := resource.NewOperation(i.(*block.BlockOperation))
```
anarcher: What do you think that type checking about `i.(*blocks.BlockOperation)` ?

spikeekips: @anarcher `i` is `interface{}`.

anarcher:
```suggestion
                             bo,ok := i.(*block.BlockOperation)
                             if !ok {
                                 return nil,nil
                             }
				r := resource.NewOperation(bo)
```

anarcher: But, Observer.Trigger do guarantee i as `*block.BlockOperation`. This checking is not needed.

spikeekips: @anarcher In this block, `renderFunc` should be get `*block.BlockOperation`, so as you suggested checking thru `ok` is unnecessary. Without type assertion, `i.(*block.BlockOperation)` it causes,
```
cannot use i (type interface {}) as type *block.BlockOperation in argument to resource.NewOperation: need type assertion
```

anarcher: `t := i.(T)` :  If i does not hold a T, the statement will trigger a panic.

`t, ok := i.(T)` :  If not, ok will be false and t will be the zero value of type T, and no panic occurs.

My suggestion is  `t,ok:=i.(T)`. Plz correct me :->

anarcher:
I think we support  Only `proposed_time` with `block.Operation`.   
For example, stellar exposes only `created_at`.

```
$http "https://horizon-testnet.stellar.org/ledgers/190" | jq .closed_at
"2018-10-04T09:42:33Z"
$http "https://horizon-testnet.stellar.org/transactions/49ef3e2489e9ed40c9cba8a0eda4a5ffa4a0a5a31565a6b30f073b9229da30fe" | jq .created_at
"2018-10-04T09:42:33Z"
$http "https://horizon-testnet.stellar.org/transactions/49ef3e2489e9ed40c9cba8a0eda4a5ffa4a0a5a31565a6b30f073b9229da30fe/operations" | jq ._embedded.records[].created_at
"2018-10-04T09:42:33Z" 
```

spikeekips:
> @anarcher : I think we support Only proposed_time with block.Operation.

I agree. One thing I worried about is, `proposed_time` is just Ballot created time, so it does not mean the confirmed time. I wish `confirmed` value will be helpful to the user in despite it is not correct and different between nodes.

spikeekips:
@kfangw Ping~

Issue #760, opened 2018-11-19
Valid transaction can be denied


spikeekips:
The rule is simple; valid transaction must be handed over to consensus process. We can think several cases inspite of of this rule. 

The most often case will be,

A. User sent valid tx0
A. It is added into TransactionPool
A. Node died with any kind of reasons and resurrected.
A. After long enough time, user notices tx0 is not stored
A. User send tx0 again
A. But node check the hash of tx0 and found it in block.TransactionPool, not TransactionPool.

anarcher:
When the node is started,  How about that adding txs of the block.TransactionPool to TransactionPool?

spikeekips:
> @anarcher : When the node is started,  How about that adding txs of the block.TransactionPool to TransactionPool? 

I think it makes sense, but I need some fundamental solution for this issue. My scenario is,

* For checking tx is already received or not, node should check,
  - whether tx is in `transaction.Pool`
  - whether tx is in `block.BlockTransaction`
  - not check in `block.TransactionPool`

When node is dead and restarted, the already received, but not yet confirmed transaction can be passed.

anarcher:
> When node is dead and restarted, the already received, but not yet confirmed transaction can be passed.

If TxPool has persistent such as disk saved pending txs, When a node is started, TxPool loads pending txs from disk. It make low performance when TxPool.Add or Remove....  :-\

spikeekips:
> @anarcher : If TxPool has persistent such as disk saved pending txs, When a node is started, TxPool loads pending txs from disk. It make low performance when TxPool.Add or Remove.... :-\

Yes, it makes sense if we have super-speed persistant storage. :)

spikeekips:
#768 will resolve this issue.

kfangw:
@spikeekips do you think that this issue can be closed?

spikeekips:
@kfangw Yes, #768  solves this issue 👍

Issue #757, opened 2018-11-19
Submitting transaction is refused


spikeekips:
https://github.com/bosnet/sebak/blob/13dbe0a17626858b51b61b4dde078554b5597458/lib/node/runner/api_node.go#L133-L135

Situation is,

1. Submit the transaction
1. The API response is 200 ok
1. Checking the transaction in history API, 'rejected'
1. The client can not send transaction any more.

@outsideris reported.

outsideris:
@soonkuk The request body is here.

```
{ H:
   { version: '1',
     created: '2018-11-19T18:39:11.832000000+09:00',
     signature:
      'L3yHPqUd5TbzHfxvRCm4Fs9RFKnexizPEz44tDKtNYyLujWJ95J7dYaPnqxiT4HSmV1iXU5ev9jfrcoAEEAT1aF' },
  B:
   { source: 'GCT67AX6PSARM47MJYAEMAQTS4LQRJDJFSMBAEEKHVCDIP5VRC3P2RPC',
     fee: '10000',
     sequence_id: 0,
     operations:
      [ { H: { type: 'congress-voting' },
          B:
           { contract:
              'dHlwZTogbWV0YQoKdGl0bGU6IE1lbWJlcnNoaXAgUmV3YXJkIFBGCgppZDogUEZfUl8wMC0wMDAtQQoKcHJvcG9zZXI6IEJsb2NrY2hhaW5PUyBJbmMKCnByb3Bvc2VyX2FjY291bnQ6IEdCTlVUV1NNNEZSU0VVTFZNSFpGN05GUVdJQkdFREY1WDVPSFhGT1pKQjZTSDVNSUVERUpFSjJGCgpleGVjdXRpb25fZHVyYXRpb246IDYzMDcyMDAKCmFtb3VudF9vZl9pc3N1YW5jZTogMTYwODMzNjAzCgpwZl9idWRnZXRfYWNjb3VudDogR0JXQ01XRFVaSzY3WU5VWjQ0VVBOVkZZWlJTQ0NTNE9MRTZPUldENFpMSTJNVkdZNEtKRFBITU8KCgo=',
             voting: { start: 300, end: 500 },
             funding_address: 'GBWCMWDUZK67YNUZ44UPNVFYZRSCCS4OLE6ORWD4ZLI2MVGY4KJDPHMO',
             amount: '160833603' } } ] } }
```

outsideris:
It was my mistake but it is still a bug.

[...]

Issue #754, opened 2018-11-19
Transaction History appended as immutable rows


kfangw:
Current Implementation of Transaction History is mutable.
Because every transaction was decidable when it is submitted,
there can exist state like this `submitted` -> `accepted` or `submitted` -> `rejected`

However, for the special payment operation(refund from the frozen account), the validity of the transaction depends on the state of blockchain(unfreezing operation and block height).
It means that a transaction may become valid even if it was known as invalid.

Hence,
`submitted` -> `rejected` -> `accepted` state transition is needed.

To solve this issue, there can be solutions.

1. A transaction can be submitted when the transaction is not registered in BlockTransaction or Transaction Pool. Transaction History is ignored

2. transaction history is immutable. And just append the state to the database.

3. the transaction history should not be related to business logic.

kfangw:
@spikeekips @anarcher @soonkuk 
What do you think about this?

soonkuk:
I think the operation(like as refund from frozen account is not wrong operation.
It is rejected because that operation did not satisfy the condition(unfreezing operation and block height).
So How about add some status like pending.

submitted -> accepted
submitted -> rejected
submitted -> pending -> accepted

kfangw:
Is pending can exist in BlockChain? @soonkuk

soonkuk:
@kfangw My opinion is to create a different state because the rejected transaction can not be accepted again.

kfangw:
solved by #768

PR #768, opened 2018-11-20
Remove BlockTransactionHistory and Add Transaction Status API

kfangw:
### Github Issue
#757 #760 #754 

### Background
`BlockTransactionHistory` was originated from CheckPoint.
Usage of this is changed to trace the history of a transaction for the client.

But this is no longer required. And should not affect to validation of transaction.

### Solution
* Remove BlockTransactionHistory and related API
* Add a new API `transaction/{hash}/status`
  * No additional data needed to be stored
  * The result of the API is extracted from BlockTransaction and Block.TransactionPool
  * Has 3 status. "notfound", "submitted", and "confirmed"


kfangw:
@spikeekips @soonkuk Please review.

lib/node/runner/api/transaction.go
```diff
@@ -136,3 +137,62 @@ func (api NetworkHandlerAPI) GetTransactionsByAccountHandler(w http.ResponseWrit
...
+			status = "submitted"
+		} else if found, _ = block.ExistsBlockTransaction(api.storage, key); found {
+			status = "confirmed"
+		}
+		payload = resource.NewTransactionStatus(key, status)
+		return payload, nil
+	}
+
+	if httputils.IsEventStream(r) {
+		event := fmt.Sprintf("hash-%s", key)
```
spikeekips: @kfangw This code will work fine. One thing suggestion. The event stream for this case, checking status of transaction, will be only meaningful when the transaction is 'not found'. So this block can be rewrite like,

* check the transaction is exists in pool or block transaction,
* if found, respond it.
* if not found and the request is event stream, watch observer.
* if not found and not event stream, just 404.

kfangw: good idea. I will patch it.

lib/node/runner/api/resource/transaction.go
```diff
@@ -40,3 +40,33 @@ func (t Transaction) Resource() *hal.Resource {
 func (t Transaction) LinkSelf() string {
 	return strings.Replace(URLTransactions, "{id}", t.bt.Hash, -1)
 }
+
+type TransactionStatus struct {
+	Hash   string
+	Status string
+}
+
+func NewTransactionStatus(hash, status string) *TransactionStatus {
```
Geod24: Since both fields are public, and it's only used in one place (`GetTransactionStatusByHashHandler`), I think we could just remove this method and use `TransactionStatus` (or `TxStatus` if you feel the name is too long) directly. It would also remove the memory allocation.

lib/node/runner/api/transaction_test.go
```diff
@@ -99,6 +99,108 @@ func TestGetTransactionByHashHandlerStream(t *testing.T) {
 	wg.Wait()
 }
 
+func TestGetTransactionStatusByHashHandler(t *testing.T) {
```
Geod24: Any chance this could be an integration test instead ?

soonkuk:
I tested for freezing and unfreezing. It works fine!

PR #772, opened 2018-11-21
Remove BlockAccountSequenceID

kfangw:
### Background
`BlockAccountSequenceID` is originated from CheckPoint.
But IMO, it is no longer required. Because BlockAccount keeps `SequenceID`.


kfangw:
@spikeekips @Geod24 @anarcher please review it.

spikeekips:
@kfangw You pointed, this block does not used in SEBAK, but I think this will be useful to check the change history of `BlockAccount`. How do you think?

lib/node/runner/checker_ballot_transaction.go
```diff
@@ -225,17 +225,10 @@ func ValidateTx(st *storage.LevelDBBackend, config common.Config, tx transaction
...
-	var bac block.BlockAccountSequenceID
-	bac, err = block.GetBlockAccountSequenceID(st, tx.B.Source, tx.B.SequenceID)
-	if err != nil {
-		return
-	}
-
 	totalAmount := tx.TotalAmount(true)
 
 	// check, have enough balance at sequenceID
-	if bac.Balance < totalAmount {
```
kfangw: > @kfangw You pointed, this block does not used in SEBAK, but I think this will be useful to check the change history of `BlockAccount`. How do you think?

Although that purpose, this should be changed like that I changed.

kfangw:
@spikeekips @Geod24 @anarcher @Charleslee522 @soonkuk @outsideris @zzim2x 
I want to discuss this Issue with you guys.
When we fully agreed about this change. I will merge it.

If I have a chance then, I will upload issue with detailed documentation.
You can wait until the issue uploaded or think about this changes in advance.
I am happy when you choose second. haha.

lib/node/runner/checker_ballot_transaction.go
kfangw: This change can be merged without debating, I think.
Hence, make another PR for it.
https://github.com/bosnet/sebak/pull/775

Geod24:
Needs a rebase

Geod24:
@kfangw : Do you need anything else ?

kfangw:
@Geod24 As @spikeekips said we need a method to track history of the account.
It is the only way to trace that(except repeating the transactions from the beginning). 
So I am looking for a better way. Until then, it can not be merged I think.

kfangw:
Will be reopened when the statedb introduced.

PR #773, opened 2018-11-21
Limit total operations in ballot

### LINKED ISSUE #747: Limit total number of operations
Currently we limit the number of transactions and the number of operations in each transaction. With this limit, we can not control and limit the total operations of each ballot. The number of operations affects the elapsed time to validate and save the ballot, so we should limit the total number of operations.

spikeekips:
### Github Issue
Resolves #747 

### Background

See #747 

### Solution

* new cmd option, `--operations-in-ballot-limit`(`SEBAK_OPERATIONS_IN_BALLOT_LIMIT)
* `common.Config.OpsInBallotLimit` keep this limit
* proposed Ballot will be generated up to this limit
* `runner.BallotTransactionsOperationLimit()` checks INIT Ballot

### DIFF
--- a/lib/common/config.go
+++ b/lib/common/config.go
@@ -19,6 +19,7 @@ type Config struct {
 
 	TxsLimit          int
 	OpsLimit          int
+	OpsInBallotLimit  int
 	TxPoolClientLimit int
 	TxPoolNodeLimit   int
 
@@ -46,8 +47,9 @@ func NewConfig(networkID []byte) Config {
 	p.BlockTime = 5 * time.Second
 	p.BlockTimeDelta = 1 * time.Second
 
-	p.TxsLimit = 1000
-	p.OpsLimit = 1000
+	p.TxsLimit = DefaultTransactionsInBallotLimit
+	p.OpsLimit = DefaultOperationsInTransactionLimit
+	p.OpsInBallotLimit = DefaultOperationsInBallotLimit
 
 	p.NetworkID = networkID
 
--- a/lib/node/runner/node_runner.go
+++ b/lib/node/runner/node_runner.go
@@ -633,18 +633,35 @@ func (nr *NodeRunner) proposeNewBallot(round uint64) (ballot.Ballot, error) {
 	// remove invalid transactions
 	nr.TransactionPool.Remove(transactionsChecker.InvalidTransactions()...)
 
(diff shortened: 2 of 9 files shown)


lib/common/config.go
```diff
@@ -46,8 +47,9 @@ func NewConfig(networkID []byte) Config {
 	p.BlockTime = 5 * time.Second
 	p.BlockTimeDelta = 1 * time.Second
 
-	p.TxsLimit = 1000
-	p.OpsLimit = 1000
+	p.TxsLimit = DefaultTransactionsInBallotLimit
+	p.OpsLimit = DefaultOperationsInTransactionLimit
+	p.OpsInBallotLimit = DefaultOperationsInBallotLimit
```
kfangw: OpsInBallotLimit should be always greater than OpsLimit
If not, some transaction may starve

spikeekips: @kfangw Yes, right. I want to these opinion :)

lib/node/runner/node_runner.go
```diff
@@ -626,19 +626,36 @@ func (nr *NodeRunner) proposeNewBallot(round uint64) (ballot.Ballot, error) {
...
+		if tx, found, err = transactionsChecker.transactionCache.Get(hash); err != nil {
+			return ballot.Ballot{}, err
+		} else if !found {
 			return ballot.Ballot{}, errors.TransactionNotFound
-		} else {
-			validTransactions = append(validTransactions, tx)
+		}
+
+		if ops+len(tx.B.Operations) > nr.Conf.OpsInBallotLimit {
+			continue
```
Charleslee522: 👍

lib/node/runner/node_runner.go
```diff
@@ -626,19 +626,36 @@ func (nr *NodeRunner) proposeNewBallot(round uint64) (ballot.Ballot, error) {
...
+		}
+
+		validTransactionHashes = append(validTransactionHashes, hash)
+		validTransactions = append(validTransactions, tx)
+
+		ops += len(tx.B.Operations)
+		if ops == nr.Conf.OpsInBallotLimit {
+			break
```
Charleslee522: 👍

lib/common/config.go
spikeekips: @kfangw This patch does not change the order of incoming transaction.

PR #774, opened 2018-11-21
add condition ratio for ricardian contract sample

outsideris:
* add `execution_condition_ratio` to calculate final voting result. This field is treated as required field in voting server.
* remove new line for print user screens which has vary width.

### DIFF
--- a/docs/ricardian._sample.yaml
+++ b/docs/ricardian._sample.yaml
@@ -3,9 +3,7 @@ type: meta
 title: Membership Reward PF
 
 abstract:
-    This document is a contract between you and BlockchainOS Inc., which will be referred to as Membership reward PF contract.
-    The contract will be effective and hence executed if Membership reward PF proposal is approved by Congress Voting.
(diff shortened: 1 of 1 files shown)


Geod24:
Do we have any formal documentation for those fields ?

soonkuk:
@Geod24 @spikeekips 
execution_condition_ratio : The execution_ratio is used to determine the voting threshold in execution condition. The voting threshold is calculated by the execution_ration of the number of members as of end of voting.

execution_condition:
    Membership reward PF Proposal will be regarded as being approved by Congress Members via Congress Voting if the following condition is met;
      - A - B ≥ C × 0.1
      - A = the number of "Agree" votes.
      - B = the number of "Disagree" votes.
      - C = the number of Members as of end of voting.

spikeekips:
@soonkuk I got it and please add this definition to https://github.com/bosnet/docs/issues/1 .

soonkuk:
@spikeekips Ok I will add it.

spikeekips:
@outsideris Please merge~

Issue #777, opened 2018-11-22
Sync failure


Charleslee522:
A node requests sync to height 100~, but sync work done only 2 and 3 height.

[sync-failed.txt](https://github.com/bosnet/sebak/files/2606232/sync-failed.txt)

Geod24:
> Second checking goroutine is checking per 10min.

Do you think we should reduce it ? We aim for a block every 5 seconds, so does it means we might be 120 blocks behind ?

anarcher:
> Do you think we should reduce it ? We aim for a block every 5 seconds, so does it means we might be 120 blocks behind ?

I added a flag (`--sync-check-prevblock`) for it. and the default value of it is 1min.  Almost case should be handled by `Observer` instantly...

spikeekips:
Can we close this?

Charleslee522:
@spikeekips  @anarcher I think we can.
The `sync` in long-term test, has one minute checking term, so 30 sec of default `CheckPrevBlockInterval` is okay.
If the test is failed again, I'll add this term option value as a parameter shorter than now.

anarcher:
I will close it.

PR #786, opened 2018-11-22
Add node flag for SEBAK_SYNC_CHECK_PREVBLOCK and refactoring

anarcher:
### Github Issue

Related #777

### Background

- For `long-term` testing,  to control  syncer waiting prevblock via node flag.
- Refactoring  `syncer.Config` and `syncer.NewXXX`
- Add more debugging logs. (`start ...` and `end ...` pattern)

### Solution

```
      --sync-check-prevblock string     sync check interval for previous block (default "10m")
```

### DIFF
--- a/lib/sync/config.go
+++ b/lib/sync/config.go
@@ -15,6 +15,7 @@ const (
 	FetchTimeout                    = 1 * time.Minute
 	RetryInterval                   = 10 * time.Second
 	CheckBlockHeightInterval        = 30 * time.Second
+	CheckPrevBlockInterval          = 10 * time.Minute
 )
 
 type Config struct {
@@ -29,6 +30,7 @@ type Config struct {
 	FetchTimeout             time.Duration
 	RetryInterval            time.Duration
 	CheckBlockHeightInterval time.Duration
+	CheckPrevBlockInterval   time.Duration
 }
 
 func NewConfig(localNode *node.LocalNode,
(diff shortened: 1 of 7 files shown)


anarcher:
Plz review it @spikeekips @Charleslee522 @Geod24  :->

lib/sync/config.go
```diff
@@ -15,6 +15,7 @@ const (
 	FetchTimeout                    = 1 * time.Minute
 	RetryInterval                   = 10 * time.Second
 	CheckBlockHeightInterval        = 30 * time.Second
+	CheckPrevBlockInterval          = 10 * time.Minute
```
spikeekips: @anarcher You think 10 minutes is best? I did not think deeply yet, but 10 minutes is too long. How about 30 seconds?

lib/sync/config.go
```diff
@@ -53,23 +55,51 @@ func NewConfig(localNode *node.LocalNode,
...
+		func(v *BlockValidator) {
+			v.prevBlockWaitTimeout = c.CheckPrevBlockInterval
+			v.logger = c.logger.New("module", "sync/validator")
+		})
+	return v
+}
+
 func (c *Config) LoggingConfig() {
 	c.logger.Info("syncer config",
+		"module", "sync/config",
```
spikeekips: We can use this manner, separate by `/` for logging, setting sub module.

lib/sync/init.go
```diff
@@ -6,7 +6,7 @@ import (
 	"boscoin.io/sebak/lib/common"
 )
 
-var log logging.Logger = logging.New("module", "sync")
+var log logging.Logger = logging.New()
```
spikeekips: By default, leaving the `module` will be better.

lib/sync/config.go
anarcher: https://github.com/inconshreveable/log15#the-varargs-style-is-brittle-and-error-prone-can-i-have-type-safety-please

image


I copied that. :-> but I am ok with `module` and `submodule` :->

spikeekips: Why did you remove module part, I like this manner.

anarcher: @spikeekips  `logger` in `init.go` has `module` part :-)

lib/sync/syncer.go
```diff
@@ -267,18 +244,21 @@ func (s *Syncer) work(height uint64, nodeAddrs []string) bool {
 				begin := time.Now()
 				syncInfo, err = s.fetcher.Fetch(ctx, syncInfo)
 				if err != nil {
-					if err != context.Canceled {
-						s.logger.Error("fetch failure", "err", err, "height", height)
+					if err == context.Canceled {
+						break L
 					}
+					s.logger.Error("fetch failure", "err", err, "height", height)
+					continue
```
Charleslee522: How about adding sleep here?
In our environment, to occur this error means same error can be occurred again by a block from any other nodes.
Because the last block is invalid.
So it makes tons of error log messages even only 30 seconds.

anarcher: No, it is not related to check prev block interval. Dont display err logs every 30 min

anarcher:
Plz review it @spikeekips @Charleslee522 @Geod24 @kfangw

lib/sync/init.go
anarcher: I changed it! @spikeekips  :->

anarcher:
Plz review it @spikeekips @Charleslee522 @Geod24 :->

anarcher:
Plz review it @Charleslee522 @Geod24

Issue #788, opened 2018-11-22
Observable and db batch writing timing


kfangw:
SEBAK uses levelDB batch.

Bellow code cannot guarantee the data is stored in the storage.
https://github.com/bosnet/sebak/blob/f64151c6d566b992a63f95aff2f3ca9e252a197f/lib/block/account.go#L77

When I write an integration test following scenario occurred occasionally

1. Register Account listening by streaming API
2. Submit transaction
3. Get event from streaming
4. Call API to get the Account
5. ERROR: the Account does not exist in the block 

From 3 to 5, the levelDB batch on SEBAK is not completed, 
hence, error 5 occurs

To resolve this issue,
the Observer should be triggered after the batch is committed.

kfangw:
solved by #790

kfangw:
solved by #790

PR #790, opened 2018-11-22
Subscribe API

kfangw:
### Github Issue
#788 

### Background
* `Streaming mode` for API is very complicated and not easy to use
* Event and DB writing time issue #788 

### Solution
* Remove Streaming mode from APIs
* Add `Subscribe` API to handle the requirements
* one Observer is needed
* Event structure for register and trigger event
* client code is revised

* NOW easy to use the streaming MODE~

### Possible Drawbacks
Need to update API documentation

### DIFF
--- a/lib/block/transaction_pool.go
+++ b/lib/block/transaction_pool.go
@@ -53,7 +53,7 @@ func (tp TransactionPool) Save(st *storage.LevelDBBackend) (err error) {
 		return
 	}
 
-	event := observer.NewSubscribe(observer.NewEvent(observer.ResourceTransactionPool, observer.ConditionTxHash, tp.Hash)).String()
+	event := observer.NewCondition(observer.ResourceTransactionPool, observer.KeyTxHash, tp.Hash).Event()
 	go observer.ResourceObserver.Trigger(event, &tp)
 
 	return nil
--- a/lib/block/transaction_pool.go
+++ b/lib/block/transaction_pool.go
@@ -53,8 +53,8 @@ func (tp TransactionPool) Save(st *storage.LevelDBBackend) (err error) {
 		return
 	}
 
-	event := fmt.Sprintf("pushed-%s", tp.Hash)
-	observer.BlockTransactionObserver.Trigger(event, &tp)
+	event := observer.NewSubscribe(observer.NewEvent(observer.ResourceTransactionPool, observer.ConditionTxHash, tp.Hash)).String()
+	go observer.ResourceObserver.Trigger(event, &tp)
 
 	return nil
 }
--- a/lib/common/observer/observer.go
+++ b/lib/common/observer/observer.go
@@ -13,16 +13,17 @@ var SyncBlockWaitObserver = observable.New()
 var ResourceObserver = observable.New()
 
 const (
-	ResourceTransaction = "tx"
-	ResourceOperation   = "op"
-	ResourceAccount     = "ac"
-	ConditionAll        = "*"
-	ConditionSource     = "source"
-	ConditionTarget     = "target"
-	ConditionType       = "type"
-	ConditionOpHash     = "ophash"
-	ConditionTxHash     = "txhash"
-	ConditionAddress    = "address"
+	ResourceTransaction     = "tx"
+	ResourceTransactionPool = "txpool"
+	ResourceOperation       = "op"
+	ResourceAccount         = "ac"
+	ConditionAll            = "*"
+	ConditionSource         = "source"
+	ConditionTarget         = "target"
+	ConditionType           = "type"
+	ConditionOpHash         = "ophash"
+	ConditionTxHash         = "txhash"
+	ConditionAddress        = "address"
 )
 
 type Event struct {
--- a/lib/common/observer/observer.go
+++ b/lib/common/observer/observer.go
@@ -7,3 +7,5 @@ var BlockTransactionObserver = observable.New()
 var BlockObserver = observable.New()
 var BlockOperationObserver = observable.New()
 var SyncBlockWaitObserver = observable.New()
+
+var XXXObserver = observable.New()
--- a/lib/common/observer/observer.go
+++ b/lib/common/observer/observer.go
@@ -1,6 +1,8 @@
 package observer
 
-import "github.com/GianlucaGuarini/go-observable"
+import (
+	"github.com/GianlucaGuarini/go-observable"
+)
 
 var BlockAccountObserver = observable.New()
 var BlockTransactionObserver = observable.New()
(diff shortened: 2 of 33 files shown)


kfangw:
@anarcher @spikeekips please review ~

lib/block/transaction_pool.go
```diff
@@ -53,8 +53,8 @@ func (tp TransactionPool) Save(st *storage.LevelDBBackend) (err error) {
 		return
 	}
 
-	event := fmt.Sprintf("pushed-%s", tp.Hash)
-	observer.BlockTransactionObserver.Trigger(event, &tp)
+	event := observer.NewSubscribe(observer.NewEvent(observer.ResourceTransactionPool, observer.ConditionTxHash, tp.Hash)).String()
+	go observer.ResourceObserver.Trigger(event, &tp)
```
anarcher: :+1:

lib/node/runner/api/stream.go
```diff
@@ -170,12 +202,15 @@ func (s *EventStream) Start(ob *observable.Observable, events ...string) func()
...
 				s.flusher.Flush()
 			case <-s.request.Context().Done():
-				close(stop)
+				close(s.stop)
 				return
 			}
 		}
 	}
 }
+func (s *EventStream) Stop() {
```
anarcher: :+1:

lib/node/runner/api/api.go
```diff
@@ -56,6 +60,50 @@ func (api NetworkHandlerAPI) HandlerURLPattern(pattern string) string {
...
 }
 
+func TriggerEvent(st *storage.LevelDBBackend, transactions []*transaction.Transaction) {
+	accountMap := make(map[string]struct{})
+	for _, tx := range transactions {
+		source := tx.Source()
+		accountMap[source] = struct{}{}
+		txHash := tx.H.Hash
+
+		txEvent := observer.NewSubscribe(observer.NewEvent(observer.ResourceTransaction, observer.ConditionAll, "")).String()
```
anarcher: IHMO,  `txEvent` is kind of event. but  it has events. 
What do you think that like below:
```
type Events []Event
func (es Events) String() string {
     var sb []string
     for _, e := range es {
         sb = append(sb,e)
    }
    return strings.Join(sb," ")
}

txEvent := o.Events{
    o.NewEvent(o.ResourceTransaction,o.ConditionAll,""),
    o.NewEvent(o.ResourceTransaction,o.ConditionSource,source),
}

go observer.ResourceObserver.Trigger(txEvent.String(), &opBlock)

```

lib/node/runner/node_runner.go
```diff
@@ -548,6 +552,8 @@ func (nr *NodeRunner) handleBallotMessage(msg interface{}) (err error) {
 			nr.log.Debug("failed to handle ballot", "error", err, "state", baseChecker.Ballot.State())
 			return
 		}
+
+		api.TriggerEvent(nr.storage, checker.StoredProposedTransactions)
```
spikeekips: This part triggers the confirmed transactions, right? If right, this is not proper place to run. How about run it as goroutine in the last `runner.saveBlock()`?

lib/node/runner/checker.go
```diff
@@ -42,17 +42,18 @@ func (c CheckerStopCloseConsensus) Checker() common.Checker {
...
+	Message                    common.NetworkMessage
+	IsNew                      bool
+	IsMine                     bool
+	Ballot                     ballot.Ballot
+	VotingHole                 voting.Hole
+	Result                     consensus.RoundVoteResult
+	VotingFinished             bool
+	FinishedVotingHole         voting.Hole
+	LatestBlockSources         []string
+	StoredProposedTransactions []*transaction.Transaction
```
spikeekips: ❤️

lib/common/observer/observer.go
```diff
@@ -1,9 +1,74 @@
...
+	if e.Condition == ConditionAll {
+		toStr += e.Condition
+	} else {
+		toStr += e.Condition + "="
+		toStr += e.Id
+	}
+	return toStr
+}
+
+type Subscribe struct {
```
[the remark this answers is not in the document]
kfangw: Yes it is.
And it is used to trigger the events.

lib/block/transaction_pool.go
```diff
@@ -53,8 +53,8 @@ func (tp TransactionPool) Save(st *storage.LevelDBBackend) (err error) {
 		return
 	}
 
-	event := fmt.Sprintf("pushed-%s", tp.Hash)
-	observer.BlockTransactionObserver.Trigger(event, &tp)
+	event := observer.NewSubscribe(observer.NewEvent(observer.ResourceTransactionPool, observer.ConditionTxHash, tp.Hash)).String()
```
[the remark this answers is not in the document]
kfangw: yes it can be.

lib/node/runner/api/api.go
kfangw: oh it is better.

kfangw: But,
I want to make an `and` condition
As you can see here,
```
type Subscribe struct {
	Events []Event `json:"resources"`
}
func (s Subscribe) String() string {
	toStr := ""
	for i, e := range s.Events {
		toStr += e.String()
		if i == len(s.Events)-1 && len(s.Events) != 1 {
			toStr += "&"
		}
	}
	return toStr
}
```

Events in Subscribe are combined with `AND` condition.

to make `OR` condition, []Subscribe is needed.

lib/common/observer/observer.go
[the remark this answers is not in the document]
anarcher: If it is, subscribe should be used with api or http layer. IMHO, the Subscribe can be in httputils :-)

lib/block/transaction_pool.go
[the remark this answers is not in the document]
kfangw: But I want to make it consistent with other events for trigger

lib/node/runner/node_runner.go
kfangw: Ah.. it was my mistake.
as-is code trigger event for every ballot ^^;;;
It should be in the function `func FinishedBallotStore(c common.Checker, args ...interface{}) error {`

Thx!

lib/common/observer/observer.go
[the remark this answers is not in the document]
kfangw: IMHO, it is closer to `observer` than `httputils`
because the Subscribe, Event can be used not only api but also other modules
If needed, there can be another structure for httputils which converted to the Subscribe and Event

kfangw:
> @kfangw This also not good place, I think. I think the event must be triggered for confirmed transactions, if true, the trigger part should be moved under `case voting.YES`.

OK then, I remove the `checker.StoredProposedTransactions` and call it at that place.

lib/common/observer/observer.go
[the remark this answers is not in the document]
anarcher: This object just makes event's string, It doesn't do subscribe..

lib/node/runner/api/api.go
anarcher: What means of `&'? :-> We can use it like `o.Trigger(o.Subscribe(events...)..)`?

kfangw: yes ~ we can use the trigger like that

subscribes are 'or' condition
events are 'and' condition

kfangw: event things are refactored. please review again~

kfangw:
@spikeekips could you finalize your review?

PR #799, opened 2018-11-23
Do more error checking on return value

Geod24:
A bunch of return values were ignored. This checks that there is no such an error happening.

### DIFF
--- a/lib/common/util.go
+++ b/lib/common/util.go
@@ -62,6 +62,18 @@ func InStringMap(a map[string]bool, s string) (found bool) {
 	return
 }
 
+//
+// Function to wrap calls to `json.Unmarshall` that cannot fail
+//
+// This function should only be used when doing calls that cannot fails,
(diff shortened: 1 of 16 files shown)


lib/common/util.go
```diff
@@ -62,6 +62,18 @@ func InStringMap(a map[string]bool, s string) (found bool) {
 	return
 }
 
+//
+// Function to wrap calls to `json.Unmarshall` that cannot fail
+//
+// This function should only be used when doing calls that cannot fails,
+// e.g. reading the content of the on-disk storage which was serialized by sebak.
+// It ensures no silent corruption of data can happen
+func JSONMustUnmarshal(data []byte, v interface{}) {
```
anarcher: How about `MustUnmarshalJSON` ? :->

Geod24: That does sounds better, will update 👍

Geod24:
Updated according to @anarcher 's comment

Issue #796, opened 2018-11-23
Failed to sync in 4 docker-based node


spikeekips:
I deployed 4 node, 2 docker containers in each docker host. I found one node, which lagged block height from others.

I stopped all nodes and the block heights of each nodes are,
```
n0 : 25
n1 : 25
n2 : 25
n3 : 9
```

Only `n3` has height, `9`. I attached the log, it contains only sync part(the original log is so huge)
https://gist.github.com/spikeekips/303dd93d880fd7bc6a8ba77de4227ce5 .

spikeekips:
@Charleslee522 @anarcher Please check this phenomenon.

anarcher:
According to this log file, This issue is invalid account sequenceID issue.

```
{"caller":"syncer.go:279","err":{"code":133,"message":"invalid sequenceID found","data":{}},"height":28,"lvl":"eror","module":"sync","msg":"validate failure","node":"GBF2.ZIJE","t":"2018-11-23T03:08:51.402112105Z"}
```

From 2 to 28,  all nodes have the same hash at each height?

@spikeekips Can I call via client api?
@Charleslee522 It is fixed `same height but different hash`?

Charleslee522:
>  It is fixed same height but different hash?

@anarcher not fixed yet.

anarcher:
@spikeekips  Can you give me logs of all nodes included `(height,hash)` or all logs ? :->

spikeekips:
@anarcher Yes I will, but it so huge. I can find the way :)

Charleslee522:
@anarcher 
The block with height 18 means nothing, right? It's because of "28" exists in hash.
Oh, `scn.n0-1` confirms two blocks B(28, 3), B(28, 4)

Charleslee522:
First, if the ISAAC design that does not vote for multiple rounds at the same time, this error will not happen. I'm implementing this but not yet.

Charleslee522:
@anarcher @spikeekips 
Second, it's because of the processing ballot simultaneously.
Normally, confirming block order is like this,
```
"got ballot" -> "NewBlock created" -> "ballot was stored" -> 
"got ballot" -> "NewBlock created" -> "ballot was stored" -> 
```
and these "got ballot" to "stored" scenario is all in checker.

But in this situation, order is,
```
"got ballot" -> "NewBlock created(28, 3)" -> "NewBlock created(28, 4)" -> "ballot was stored(28, 3)" -> ...
```
[...]

PR #805, opened 2018-11-24
Self-signed ballot will use `network.MessageBroker` instead of `ballotBroadcastChannel`

spikeekips:
### Github Issue

Related to #796 

### Background

See #796's comment, https://github.com/bosnet/sebak/issues/796#issuecomment-441344594 .

### Solution

We already patched to prevent node's signed ballot to be broadcasted to itself in #745 . In #745 to pass the ballot to checker, the `ballotBroadcastChannel` was created. @Charleslee522 pointed in https://github.com/bosnet/sebak/issues/796#issuecomment-441344594 , this way made the unexpected problem, conjunction of incoming ballot. This patch uses `network.MessageBroker` like other incoming ballot for self-signed ballot. This will help the consensus to be stable.

### DIFF
--- a/lib/network/memory.go
+++ b/lib/network/memory.go
@@ -1,12 +1,14 @@
 package network
 
 import (
+	"io"
 	"net"
 	"net/http"
 
-	"boscoin.io/sebak/lib/common"
 	"github.com/google/uuid"
 	"github.com/gorilla/mux"
+
+	"boscoin.io/sebak/lib/common"
 )
 
 type MemoryNetwork struct {
@@ -121,6 +123,7 @@ func (prev *MemoryNetwork) NewMemoryNetwork() *MemoryNetwork {
 	}
 
 	n.peers[n.endpoint.String()] = n
+	n.messageBroker = &MemoryMessageBroker{network: n}
 
 	return n
 }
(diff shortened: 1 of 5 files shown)


lib/node/runner/node_runner.go
```diff
@@ -682,25 +670,11 @@ func (nr *NodeRunner) NodeInfo() node.NodeInfo {
...
-
-func (nr *NodeRunner) broadcastBallot(b ballot.Ballot) {
-	go func() {
-		nr.handleBallotMessage(b)
+		encoded, _ := b.Serialize()
+		nr.Network().MessageBroker().Receive(common.NewNetworkMessage(common.BallotMessage, encoded))
 	}()
 
-	nr.ConnectionManager().Broadcast(b)
+	go nr.ConnectionManager().Broadcast(b)
```
Charleslee522: Should we call this as a go routine?
In this `Broadcast` function, all sending call is in go routine. And it has read lock.

lib/network/memory.go
```diff
@@ -132,3 +135,17 @@ func (p *MemoryNetwork) AddHandler(string, http.HandlerFunc) *mux.Route {
 func (p *MemoryNetwork) AddMiddleware(string, ...mux.MiddlewareFunc) error {
 	return nil
 }
+
+type MemoryMessageBroker struct {
```
Charleslee522: 👍

lib/node/runner/node_runner.go
spikeekips: @Charleslee522 Yes, you right, it should not. I'll fix it.

PR #806, opened 2018-11-24
modify blockOperation

soonkuk:
### Background
This PR is made to resolve frozen account issue.
The problem was that `/accounts/{addr}/frozen-accounts` return no records even if creating frozen accounts are a success.

### DIFF
--- a/lib/block/operation.go
+++ b/lib/block/operation.go
@@ -55,7 +55,7 @@ func NewBlockOperationFromOperation(op operation.Operation, tx transaction.Trans
 	}
 
 	linked := ""
-	if createAccount, ok := op.B.(*operation.CreateAccount); ok {
+	if createAccount, ok := op.B.(operation.CreateAccount); ok {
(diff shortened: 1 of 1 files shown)


outsideris:
To keep history, the problem was that `/accounts/{addr}/frozen-accounts` return no records even if creating frozen accounts are a success.

soonkuk:
@outsideris Ok I will add it to the background

outsideris:
On my local test, it works correctly.
I have questions. If i had frozen accounts already before this PR, they are not returened. Right? 
Only new frozen accounts are returned after this PR.

soonkuk:
@outsideris Yes that was not saved.

kfangw:
@soonkuk integration_tests_sdk  failed occasionally.
I will investigate this. Please rerun until succeed.
And I will make an integration test for this case soon.

soonkuk:
@kfangw Ok I got it.

PR #809, opened 2018-11-24
Sync latest nodes

anarcher:
### Background

Current sync doesn't share latest node address from the latest consensus's result with every fetching height. Sebak has static validator's address list ( It means without restarting the node, validators are not changed).  

But If some validator is Byzantine validator or  the node can update validators dynamically,  The consensus or node can know that and can update new latest nodes to syncer. With this, the syncer can try re-fetching with new latest nodes when the syncer has still failed heights or heights not fetched.

### Solution

- Syncer updates latest node address. 
- Fetcher uses the latest node address every time it try.
- And some refactoring about unused objects and `NewXXX`

### Possible Drawbacks

- `NodeList` has `RWMutex`. So It could decrease a litte bit of performance.

### DIFF
--- a/lib/sync/config.go
+++ b/lib/sync/config.go
@@ -25,6 +25,7 @@ type Config struct {
 	connectionManager network.ConnectionManager
 	tp                *transaction.Pool
 	localNode         *node.LocalNode
+	nodelist          *NodeList
 	logger            log15.Logger
 	commonCfg         common.Config
 
@@ -49,6 +50,7 @@ func NewConfig(localNode *node.LocalNode,
 		logger:            log.New(log15.Ctx{"node": localNode.Alias()}),
 		commonCfg:         cfg,
 		localNode:         localNode,
+		nodelist:          &NodeList{},
 
 		SyncPoolSize:             SyncPoolSize,
 		FetchTimeout:             FetchTimeout,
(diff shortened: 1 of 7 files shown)


spikeekips:
@anarcher Does this PR help to resolve #796 ?

anarcher:
> @anarcher Does this PR help to resolve #796 ?

Unfortunately, I don't think so.  This PR can support dynamic validators and avoid byzantine validators a litte bit..

anarcher:
> @anarcher Does this PR help to resolve #796 ?

Huum, It can help #796 If consensus can remove  ill-behaved nodes of latest nodes. But  without state db, syncer can't revert db. ㅠㅠ But, with #778,  if node make wrong block, It and syncer can't help it.

anarcher:
Plz Review it

Charleslee522:
@anarcher I have a question.
Does `Syncer.SetSyncTargetBlock(ctx context.Context, height uint64, nodeAddrs []string)` still need nodeAddrs??

Charleslee522:
> Yes absolutely as you know

@anarcher I asked you because I didn't know that :)
Oh `SetSyncTargetBlock ` is the entry point. LGTM

PR #810, opened 2018-11-24
Set log output for HTTP request & response

spikeekips:
### Background

Currently http request and response log messages are mixed with other messages, and they are produced huge messages.

But most of all, http request and response log should be able to be separated to another output, because http log has client's information and we should keep it safe and managing.

### Solution

* add new `--http-log`, it set the http request and response log output.
* basically the output format is 'json' list

### log 
```json
{
  "content-length": 0,
  "content-type": "",
  "headers": {},
  "host": "localhost:12345",
  "id": "1a712c78-fa80-4f4a-b291-09dc24a4839f",
  "lvl": "dbug",
  "method": "GET",
  "module": "http",
  "msg": "request",
  "node": "GCPQ.7DUR",
  "proto": "HTTP/1.1",
  "referer": "",
  "remote": "127.0.0.1:55083",
  "t": "2018-11-24T19:20:42.3382+09:00",
  "uri": "/",
  "user-agent": "curl/7.62.0",
  "x-forwarded-for": ""
}
{
  "elapsed": "1.583469ms",
  "id": "1a712c78-fa80-4f4a-b291-09dc24a4839f",
  "lvl": "dbug",
  "module": "http",
  "msg": "response",
  "node": "GCPQ.7DUR",
  "size": 1147,
  "status": 200,
  "t": "2018-11-24T19:20:42.339759+09:00"
}
```

There are 2 log for one request, first one is request itself and the second is it's response. In request, there are the basic informations and 'x-forwarded-for'. In response, there is 'elapsed' time. The request and response can be joined by `id`.

### DIFF
--- a/cmd/sebak/cmd/run.go
+++ b/cmd/sebak/cmd/run.go
@@ -43,6 +43,7 @@ var (
 	flagDebugPProf      bool   = common.GetENVValue("SEBAK_DEBUG_PPROF", "0") == "1"
 	flagKPSecretSeed    string = common.GetENVValue("SEBAK_SECRET_SEED", "")
 	flagLog             string = common.GetENVValue("SEBAK_LOG", "")
+	flagHTTPLog         string = common.GetENVValue("SEBAK_HTTP_LOG", "")
 	flagLogLevel        string = common.GetENVValue("SEBAK_LOG_LEVEL", defaultLogLevel.String())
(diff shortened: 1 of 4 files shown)


outsideris:
Could you show log output for examples?

spikeekips:
@outsideris I updated the description with example.

PR #812, opened 2018-11-25
Clean up txs  in txpool when sync up

anarcher:
### Background

Clean up txs in  txpool when sync up 
When sync up,  synced block's txs and proposerTx should be removed within txpool if a node state is `sync` and the node recv txs.

### Solution

In `finishBlock`,  `txpool.Remove(blk.Transactions...)` and `txpool.Remove(blk.ProposerTransaction)` too.

### Possible Drawbacks

To use `txpool` with `syncer`, txpoll is created in   `cmd/run.go`.  It makes a change of `NewNodeRunnder`.  :->

### DIFF
--- a/lib/sync/fetcher.go
+++ b/lib/sync/fetcher.go
@@ -125,6 +125,7 @@ func (f *BlockFetcher) fetch(ctx context.Context, si *SyncInfo) error {
 		f.logger.Error("request err", "err", err, "height", height)
 		return err
 	}
+	req = req.WithContext(ctx)
 
 	resp, err := f.apiClient.Do(req)
 	if err != nil {
--- a/cmd/sebak/cmd/run.go
+++ b/cmd/sebak/cmd/run.go
@@ -27,6 +27,7 @@ import (
 	"boscoin.io/sebak/lib/node/runner"
 	"boscoin.io/sebak/lib/storage"
 	"boscoin.io/sebak/lib/sync"
+	"boscoin.io/sebak/lib/transaction"
 )
 
 const (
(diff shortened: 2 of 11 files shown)


anarcher:
@spikeekips how to do? Let me teach me ㅠ

spikeekips:
@anarcher You already know the way :) Simply move `sync.NewConfig()` inside `NewNodeRunner()`.
```go
c := sync.NewConfig(
    nr.Node(),
    nr.Storage(),
    nr.Network(),
    nr.ConnectionManager(),
    nr.TransactionPool,
    nr.Conf,
)
```
In the https://github.com/bosnet/sebak/blob/e3fe777b44bbfcefe9b500572300c051d72e52ed/lib/node/runner/node_runner.go#L158 .

anarcher:
@spikeekips IHMO, It's not simple. As you know, `consensus.ISAAC` should be started with `syncer`.

 https://github.com/bosnet/sebak/blob/master/cmd/sebak/cmd/run.go#L644

Like stroage.LeveldbBackend, IMHO, it's not bad createing `txpool` in main.go

spikeekips:
@anarcher Hmm, I understood. I worried the `NodeRunner` is getting fat with a many arguments :) I don't want to interrupt this PR, but we can find another way, How about move `NewSyncer()` and `ISAAC` also  inside `NodeRunner`?

anarcher:
> How about move NewSyncer() and ISAAC also inside NodeRunner?

IHMO, It just makes NodeRunner is more complicated. Many arguments issues point out that NodeRunner has to do many works. Syncer and NodeRunner don't need to do any interacting. They do not need to know each other.

lib/sync/fetcher.go
```diff
@@ -125,6 +125,7 @@ func (f *BlockFetcher) fetch(ctx context.Context, si *SyncInfo) error {
 		f.logger.Error("request err", "err", err, "height", height)
 		return err
 	}
+	req = req.WithContext(ctx)
```
spikeekips: 👍

PR #813, opened 2018-11-25
Move test.go to base_test.go

anarcher:
Because  *_test.go is only included to test build.  Maybe `test.go` is included to build binary.

https://godoc.org/boscoin.io/sebak/lib/block#TestMakeNewBlock

image


spikeekips:
@anarcher If possible, it will be good. The reason why test.go in not xx_test.go was the functions in test.go should be used by another packages. If this PR makes it better, 👍

anarcher:
> @anarcher If possible, it will be good. The reason why test.go in not xx_test.go was the functions in test.go should be used by another packages. If this PR makes it better, 👍

ㅠㅠ, Our naming is not fit with PR.  For example, `TestMakeBlockAccount()` is OK  with test.go but  is not OK with base_test.go ㅠㅠㅠㅠ

PR #821, opened 2018-11-26
Integration test fail sdk

kfangw:
### Background
Integration test for SDK failed occasionally
It is due to JSON unmarshal object failed

And due to 2a7a05

### DIFF
--- a/lib/client/client.go
+++ b/lib/client/client.go
@@ -3,8 +3,11 @@ package client
 import (
 	"boscoin.io/sebak/lib/common/observer"
 	"bufio"
+	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
+	"io/ioutil"
 	"net/http"
 	neturl "net/url"
 	"strconv"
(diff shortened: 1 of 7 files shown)


kfangw:
#824 report an issue to think more later.

lib/client/client.go
```diff
@@ -118,7 +121,14 @@ func NewDefaultListOptionsFromQuery(v neturl.Values) (options *storage.DefaultLi
 
 func (c *Client) ToResponse(resp *http.Response, response interface{}) (err error) {
 	defer resp.Body.Close()
-	decoder := json.NewDecoder(resp.Body)
+
+	b, err := ioutil.ReadAll(resp.Body)
+
+	fmt.Println(string(b))
```
Charleslee522: @kfangw Isn't this leftover?

lib/client/client.go
```diff
@@ -236,14 +246,15 @@ func (c *Client) SubmitTransaction(tx []byte) (pTransaction TransactionPost, err
 }
 
 func (c *Client) SubmitTransactionAndWait(hash string, tx []byte) (pTransaction TransactionPost, err error) {
-
+	fmt.Println(string(hash))
```
Charleslee522: @kfangw How about this? :)

lib/client/client.go
Charleslee522: @kfangw Oh, sorry :)

Issue #434, opened 2018-09-19
Serialize blockchain data / tx data to binary


Geod24:
Currently we are communicating through `JSON`.
While this is great for readability, it uses a lot more bandwidth / memory than it should and will make it hard to scale. It also starts to show problems such as [this](https://github.com/bosnet/sebak/issues/412).

We should switch to something a lightweight format. Bitcoin and ethereum use a simple binary serialization, Stellar uses XDR, and I'm pretty certain some chains do use protobuf.
I am personally leaning towards binary because it's the simplest, more lightweight format.

kfangw:
Good Point.
Many of team members have been worrying about it. But, it has not a high priority.

BTW, There is one more point to consider. SEBAK store data as JSON
When you research binary format for communicating, please mind this point also.
@Geod24

Geod24:
@kfangw : We had a long discussion with @Charleslee522 about this yesterday.
My point was that all immutable data (blockchain data) should be serialized in the most efficient manner, e.g. as binary files to the disk (like bitcoin daemon does).
There's 2 reasons for that:
- We cannot prune those data, nor reconstruct it, so as soon as its here, it's not supposed to change, ever. As a result, the more overhead we add to it, the worst our disk space.
- It makes accessing it externally simpler to have it defined: at the moment we didn't define the format, which makes writing tools impossible.

I am of the opinion that we should separate blockchain data from application data. We can LevelDB for all application data (that is, data that we can afford to loose / are able to reconstruct from other data) and a binary serialization for the moment is only the blockchain.

Does it make sense?

kfangw:
@Geod24 I got your point except for the definition of blockchain data.
Could you enlight me that how distinguish blockchain data/application data according to SEBAK implementation?
Stored data in LevelDB including Block, BlockAccount, BlockTransaction, and BlockOperation,  is blockchain data? or application data?
IMO, we have application data only for the indexes for the BlockXXXs

Geod24:
> Block, BlockAccount, BlockTransaction, and BlockOperation, is blockchain data

Yep.
Application data would be things like indexes, transaction pool, etc...

anarcher:
- https://github.com/alecthomas/go_serialization_benchmarks

anarcher:
- https://github.com/tendermint/go-amino

spikeekips:
> from #606, 
[...]

PR #826, opened 2018-11-26
Refactor serialization code to be more compact

Geod24:
This is a preliminary to fixing #434 

While working on reducing the memory / disk consumption, I had to deal with the `Serialize` method a lot. One problem we have is that serialization should not be done a single way. E.g. we don't really want to have binary serialization for the HTTP endpoints.

This leads to the issue with `Serialize`: It is a wrong abstraction, as the caller of the method (e.g. the LevelDB code or the HTTP code) should select the right serialization method for the data structure, not the other way around (`BlockOperation` should not defined how it is serialized, but it should define that it can be serialized as Binary/JSON/etc...).

This PR removes `Serialize` methods and `Serialized`, pushed the `json` code where needed and remove needless wrapper.

### DIFF
--- a/lib/storage/leveldb.go
+++ b/lib/storage/leveldb.go
@@ -1,6 +1,7 @@
 package storage
 
 import (
+	"encoding"
 	"encoding/json"
 
 	"github.com/syndtr/goleveldb/leveldb"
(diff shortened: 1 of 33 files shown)


Geod24:
Updated, and removed the `OperationType` change as it could break the DB layout.

Geod24:
Ping

Geod24:
Rebased on master + removed a commit which would have broken the DB layout

lib/storage/leveldb.go
```diff
@@ -418,3 +415,19 @@ func (st *LevelDBBackend) Walk(prefix string, option *WalkOption, walkFunc WalkF
 
 	return iter.Error()
 }
+
+// Encapsulate serialization method for various functions
+func serialize(i interface{}) ([]byte, error) {
+	if bm, ok := i.(encoding.BinaryMarshaler); ok {
+		return bm.MarshalBinary()
```
Charleslee522: 👍

PR #832, opened 2018-11-27
[WIP] Response return HTML character as-is

kfangw:
### Background
APIs returns Resource as JSON encoded byte.
from GoLang 1.7 (https://golang.org/doc/go1.7#encoding_json)  `&, <, and >` characters are escaped by default.

### Solution
If the client decodes the received string as JSON first there is no problem.
But to make it readable and to not make confuse, the escaping does not necessary.

### DIFF
--- a/lib/common/util.go
+++ b/lib/common/util.go
@@ -83,6 +83,14 @@ func JSONMarshalIndent(o interface{}) ([]byte, error) {
 	return json.MarshalIndent(o, "", "  ")
 }
 
+func JSONMarshalWithoutEscapeHTML(t interface{}) ([]byte, error) {
+	buffer := &bytes.Buffer{}
+	encoder := json.NewEncoder(buffer)
+	encoder.SetEscapeHTML(false)
+	err := encoder.Encode(t)
+	return buffer.Bytes(), err
+}
+
 func ReverseStringSlice(a []string) []string {
 	if len(a) < 1 {
 		return []string{}
--- a/lib/network/httputils/json.go
+++ b/lib/network/httputils/json.go
@@ -37,9 +37,8 @@ func WriteJSONError(w http.ResponseWriter, err error) {
 
 // WriteJSON writes the value v to the http response as json encoding
 func WriteJSON(w http.ResponseWriter, code int, v interface{}) error {
-	if h, ok := v.(HALResource); ok {
+	if _, ok := v.(HALResource); ok {
 		w.Header().Set("Content-Type", "application/hal+json")
-		v = h.Resource()
 	} else if e, ok := v.(error); ok {
 		w.Header().Set("Content-Type", "application/problem+json")
 		v = NewErrorProblem(e, code)
@@ -49,7 +48,13 @@ func WriteJSON(w http.ResponseWriter, code int, v interface{}) error {
 
 	w.WriteHeader(code)
 
-	bs, err := json.Marshal(v)
+	var bs []byte
+	var err error
+	if marshaler, ok := v.(json.Marshaler); ok {
+		bs, err = marshaler.MarshalJSON()
+	} else {
+		bs, err = json.Marshal(v)
+	}
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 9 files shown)


lib/common/util.go
```diff
@@ -83,6 +83,14 @@ func JSONMarshalIndent(o interface{}) ([]byte, error) {
 	return json.MarshalIndent(o, "", "  ")
 }
 
+func JSONMarshalWithoutEscapeHTML(t interface{}) ([]byte, error) {
```
spikeekips: Please update `common.EncodeJSONValue()` too.

lib/network/httputils/json.go
```diff
@@ -49,7 +48,13 @@ func WriteJSON(w http.ResponseWriter, code int, v interface{}) error {
 
 	w.WriteHeader(code)
 
-	bs, err := json.Marshal(v)
+	var bs []byte
+	var err error
+	if marshaler, ok := v.(json.Marshaler); ok {
+		bs, err = marshaler.MarshalJSON()
```
Geod24: `json.Marshal` already takes care of that if you implement the interface

kfangw: By using json.Marshal method, it impossible to set escapeHTML = false
As you can see here, Line 456 opts.escapeHTML is set true by default.
https://github.com/golang/go/blob/22dbc96d8a004bb78a0efd329d0f48b6f0a8b9e5/src/encoding/json/encode.go#L443-L461

Geod24: It is not set to `true` by default, it's just passing the value that was set in the option.
So if you set it to `false` in the caller, it should just work.

kfangw: the `compact` function overwrite the option.

I think it is a bug of the json library. @Geod24

Geod24:
@kfangw : Where did you see the escaping happening ? The current test does not properly check that the escaping occurs.

Geod24:
@kfangw : Looking at the code / documentation, it seems that only https://github.com/Geod24/sebak/commit/b4f2fc038b44c9b39e970a063fe48b88c4c78202 would be necessary. However, I don't know how to test it, as I don't see a ressources which could have those characters.

Geod24:
> from GoLang 1.7 (https://golang.org/doc/go1.7#encoding_json) &, <, and > characters are escaped by default.

They were always escaped by default by the way:
> **As in previous releases**, the encoder defaults to applying this escaping

But 1.7 allows to disable this encoding. They wouldn't change the default as it would break the Go1 compatibility promise.

kfangw:
WIP.

kfangw:
@Geod24 
Default is escapeHTML true
```
// NewEncoder returns a new encoder that writes to w.
func NewEncoder(w io.Writer) *Encoder {
	return &Encoder{w: w, escapeHTML: true}
}
```

To set the escapeHTML false,
It is necessary to implement the interface 
```
// Marshaler is the interface implemented by types that
// can marshal themselves into valid JSON.
type Marshaler interface {
	MarshalJSON() ([]byte, error)
}
```
like this
```
func JSONMarshalWithoutEscapeHTML(t interface{}) ([]byte, error) {
	buffer := &bytes.Buffer{}
	encoder := json.NewEncoder(buffer)
	encoder.SetEscapeHTML(false)
	err := encoder.Encode(t)
	return buffer.Bytes(), err
}
```

However,
The EscapeHTML flag is ignored when the json.Marshal called because of this code
https://github.com/golang/go/blob/22dbc96d8a004bb78a0efd329d0f48b6f0a8b9e5/src/encoding/json/encode.go#L157-L170

In short, let's assume that there is a structure which has field for another structure like this
```
type A struct {

}
type B struct {
    a  A
}
```
If the A does `not` has MarshalJSON method which set the escapeHTML false like the function avobe `JSONMarshalWithoutEscapeHTML` despite of B has the method,
Thre returned byte slice of json.Marshal(A) is result with escapedHTML is true.

kfangw:
When I set escapeHTML false for all MarshalJSON method then,
0x10 ( 16  DLE (data link escape) ) character also remains.
I am not sure what happened in the json library, and it is not intended.

So, I conclude that drop this feature.
@Geod24 @spikeekips @anarcher

PR #835, opened 2018-11-27
Fixed to start consensus after nodes restarting simultaneously 

### LINKED ISSUE #828: Stuck after restarting all the nodes at the same time
I restarted all the nodes at the same time and then consensus was stuck in all the nodes. The reason is, some node broadcasted the received ballot or the proposed ballot before the enough validators are connected.

> The logs of all nodes can be found at https://gist.github.com/spikeekips/94c2d02a0f736c36d45cfbed709aa9a2 .

spikeekips:
### Github Issue
Resolves #828 

### Background

See #828 

This problem is caused by the combination of `NodeRunner.BroadcastBallot()` and `NodeRunner.waitForConnectingEnoughNodes()`. The network was stuck, failed to collect the enough ballots over threshold. Before enough validators are not connected, the incoming ballot or the proposed ballot is received, the node can not broadcast the next ballots.

### Solution

Changes,
* `network.ConnectionManager.IsReady()` added; this new function checks the enough validators are connected
* When broadcasting, if node is in `StateBooting` and does not have enough connected validators, it will wait until having enough validators.

### DIFF
--- a/lib/node/runner/checker.go
+++ b/lib/node/runner/checker.go
@@ -667,7 +667,7 @@ func ACCEPTBallotBroadcast(c common.Checker, args ...interface{}) (err error) {
 		"ballot will be broadcasted",
 		"new-ballot", newBallot.GetHash(),
 		"new-state", newBallot.State(),
-		"voting-hole", checker.VotingHole,
+		"voting-hole", checker.FinishedVotingHole,
 	)
 
 	return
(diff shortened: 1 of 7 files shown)


lib/node/runner/checker.go
```diff
@@ -667,7 +667,7 @@ func ACCEPTBallotBroadcast(c common.Checker, args ...interface{}) (err error) {
 		"ballot will be broadcasted",
 		"new-ballot", newBallot.GetHash(),
 		"new-state", newBallot.State(),
-		"voting-hole", checker.VotingHole,
+		"voting-hole", checker.FinishedVotingHole,
```
Charleslee522: not related, but 👍

PR #837, opened 2018-11-27
Use bufio.Reader in place of bufio.Scanne for large data

anarcher:
### Background

For large data fetch,  `bufio.Scanner` raises error : response failed to unmarshal: bufio.Scanner: token too long

```t=2018-11-27T21:05:57+0900 lvl=eror msg="fetch err" module=sync node=GBVG.MAIO submodule=fetcher err="response failed to unmarshal: bufio.Scanner: token too long" height=243 caller=fetcher.go:82```

### Solution

Replace `bufio.Reader` in place of `bufio.Scanner`.

### CODE REVIEW

**Comment by Geod24** [comment]
Can you add a test for it ?

File: lib/sync/fetcher_test.go [hunk 1]
```diff
@@ -82,3 +83,19 @@ func TestBlockFetcher(t *testing.T) {
...
+func TestLargeFetch(t *testing.T) {
+	f := &BlockFetcher{}
+	f.logger = log
+
+	b := bytes.NewReader([]byte(largeData))
+
+	items, err := f.unmarshalResp(b)
+	require.NoError(t, err)
+	require.Equal(t, len(items), 2)
+
```
Geod24: Extra whiteline

anarcher: IMHO, we need to reduce tx payload.....

Geod24: You are right, but my comment is about the empty line before `}`

PR #839, opened 2018-11-27
Change IsValidVotingBasis to return the reason of the mismatch

Geod24:
```
Make it more obvious what went wrong, also add documentation
```

### DIFF
--- a/lib/consensus/isaac.go
+++ b/lib/consensus/isaac.go
@@ -88,30 +88,43 @@ func (is *ISAAC) SelectProposer(blockHeight uint64, round uint64) string {
 	return is.proposerSelector.Select(blockHeight, round)
 }
 
-func (is *ISAAC) IsValidVotingBasis(basis voting.Basis, latestBlock block.Block) bool {
+//
+// Check if `basis` is a valid one for the current round of consensus
+//
+// Params:
+//   basis = `voting.Basis` received for the current round of consensus
+//   latestBlock = the latest block known to the node
+//
+// Returns:
+//   bool = `true` if it's a valid voting basis, `false` otherwise
+//   string = A string explaining why it's not the latest voting basis (when `bool` == `false),
+//   or empty string ("") if the voting basis is valid (when `bool` == `true`).
+//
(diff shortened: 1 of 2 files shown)


lib/consensus/isaac.go
```diff
@@ -88,30 +88,42 @@ func (is *ISAAC) SelectProposer(blockHeight uint64, round uint64) string {
...
+//
+// Params:
+//   basis = `voting.Basis` received for the current round of consensus
+//   latestBlock = the latest block known to the node
+//
+// Returns:
+//   A string explaining why it's not the latest voting basis,
+//   or empty string ("") if the voting basis is valid.
+//
+func (is *ISAAC) IsValidVotingBasis(basis voting.Basis, latestBlock block.Block) string {
```
Charleslee522: I think `bool` return is better to understand this method. Because it starts with `Is`.
So, how about returning (bool, string)? like
```
return false, "basis' height is different from latest block's height"
return true, ""
return false, "basis' block hash does not match latest block's hash"
return false, return "basis' round is <= to latest voting basis' round"
return true, ""
```

lib/node/runner/checker.go
```diff
@@ -313,19 +313,19 @@ func BallotCheckBasis(c common.Checker, args ...interface{}) (err error) {
...
+	if reason := checker.NodeRunner.Consensus().IsValidVotingBasis(
 		checker.Ballot.VotingBasis(),
 		blk,
-	) {
-		err = errors.InvalidVotingBasis
-		checker.NodeRunner.Log().Debug(
+	); len(reason) > 0 {
+		checker.NodeRunner.Log().Error(
 			"voting basis is invalid",
+			"reason", reason,
```
Charleslee522: 👍

lib/consensus/isaac.go
Geod24: Not a fan of redundant return value (because returning `true` and a non-empty string should never happen), but I can compromise on this :)

anarcher:
Sometimes, l feel that  Reason struct is necessary for various reason.. Haha

PR #845, opened 2018-11-28
Fix a bug about blocks api paging

anarcher:
### Background

- `/api/v1/blocks` 's cursor is height `cursor=12123`. So  this cursor should not do base64 encoding 
- Remove `storage.WalkOption.SkipCursor`.  It's an insane approach. 
 
### Solution

- `PageQueryOption` for base64encoding or not. Default is encoding.  e.g. `p, err := NewPageQuery(r, WithEncodePageCursor(false))`
-  Remove `WalkOption.SkipCursor`
- Without cursor, `storage.Walk` call `iter.Last()` or `iter.First()` depend on `WalkOption.Reverse`.

### DIFF
--- a/lib/node/runner/api/blocks.go
+++ b/lib/node/runner/api/blocks.go
@@ -18,34 +18,34 @@ func (api NetworkHandlerAPI) GetBlocksHandler(w http.ResponseWriter, r *http.Req
 		return
 	}
 
-	var (
-		firstHeight uint64
-		lastHeight  uint64
-		blocks      []resource.Resource
-	)
-
+	var currHeight uint64
 	var option *storage.WalkOption
 	{
 		var cursor string
 		if p.Cursor() != nil {
-			height, err := strconv.ParseUint(string(p.Cursor()), 10, 64)
+			var err error
+			currHeight, err = strconv.ParseUint(string(p.Cursor()), 10, 64)
 			if err != nil {
 				httputils.WriteJSONError(w, errors.BadRequestParameter)
 				return
 			}
-			cursor = block.GetBlockKeyPrefixHeight(height)
+			cursor = block.GetBlockKeyPrefixHeight(currHeight)
 		}
 		option = storage.NewWalkOption(cursor, p.Limit(), p.Reverse())
 	}
 
+	var (
+		firstHeight uint64 = currHeight
+		lastHeight  uint64 = currHeight
+		blocks      []resource.Resource
+	)
 	{
 		var cnt uint64 = 1
 		err := block.WalkBlocks(api.storage, option, func(b *block.Block, key []byte) (next bool, err error) {
-			if cnt >= p.Limit() {
-				lastHeight = b.Height
-			} else if cnt == 1 {
+			if cnt == 1 {
 				firstHeight = b.Height
 			}
+			lastHeight = b.Height
 			blocks = append(blocks, resource.NewBlock(b))
 			cnt++
 			return true, nil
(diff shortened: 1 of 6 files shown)


lib/node/runner/api/blocks.go
```diff
@@ -5,58 +5,67 @@ import (
...
 
 	var option *storage.WalkOption
 	{
-		height, err := strconv.ParseUint(string(p.Cursor()), 10, 64)
-		if err != nil {
-			height = common.GenesisBlockHeight // default cursor is genesis block height
+		var cursor string
+		if p.Cursor() != nil {
+			height, err := strconv.ParseUint(string(p.Cursor()), 10, 64)
+			if err == nil {
```
Geod24: Shouldn't we return an error if the query contains an invalid cursor ?

anarcher: My intent is that it ignore wrong cursor param (error)

Geod24: Is that a good thing ? We usually error out when the query is wrong.

anarcher: Oh, it’s a good point. I will change it! Thanks @Geod24

anarcher:
4cbe905  What do you think? 
If the next cursor is empty, don't display hal `next` .

spikeekips:
> @anarcher : 4cbe905  What do you think? If the next cursor is empty, don't display hal `next` . 

In most cases for 'hal', the 'next' and 'prev' link exists with same link, not empty.

lib/node/runner/api/blocks.go
```diff
@@ -5,58 +5,70 @@ import (
...
+			if cnt > p.Limit() {
+				nextHeight = b.Height
+				return false, nil
 			}
-			cursor = []byte(strconv.FormatUint(height, 10))
-			if len(firstCursor) == 0 {
-				firstCursor = append(firstCursor, cursor...)
+			blocks = append(blocks, resource.NewBlock(b))
+			if cnt == 1 {
+				prevHeight = b.Height - 1
```
kfangw: When I test API with this query `cursor=10&limit=10&reverse=true`
I expect the next cursor is 11, not 9

To resolve the problem, this code needs to be changed
```
if cnt == 1 {
        if option.Reverse {
	    prevHeight = b.Height + 1
        } else {
            prevHeight = b.Height - 1
        }
}
```

when reverse=true
it should be + 1
prevHeight = b.Height + 1

lib/node/runner/api/resource/resource.go
```diff
@@ -36,8 +36,13 @@ func (l ResourceList) Resource() *hal.Resource {
 		rCollection = append(rCollection, apiResource.Resource())
 	}
 	rl.EmbedCollection("records", rCollection)
-	rl.AddLink("prev", hal.NewLink(l.LinkPrev())) //TODO: set prev/next url
-	rl.AddLink("next", hal.NewLink(l.LinkNext()))
+
+	if l.LinkPrev() != "" {
+		rl.AddLink("prev", hal.NewLink(l.LinkPrev())) //TODO: set prev/next url
```
kfangw: Although there are no more items when this link made,
when the client request thru the link, in the meantime, new items may be created.

What do you think? @anarcher

lib/node/runner/api/blocks.go
anarcher: Oh l mistake ㅠ

lib/node/runner/api/blocks.go
```diff
@@ -5,58 +5,70 @@ import (
...
+	var (
+		prevCursor []byte
+		nextCursor []byte
+	)
+	if prevHeight > 0 {
+		prevCursor = []byte(strconv.FormatUint(prevHeight, 10))
+	}
+	if nextHeight > 0 {
+		nextCursor = []byte(strconv.FormatUint(nextHeight, 10))
+	}
```
Charleslee522: 👍

anarcher:
@kfangw Plz review it

PR #848, opened 2018-11-29
Operation Resource `self` to /transactions/{txhash}/operations/{id}

kfangw:
### Background
Operation Resource has `self` link like `/operations/{hash}`
But there is no API support this.

### Solution
* Add the API /transactions/{id}/operations/{opindex}
* Remove operation event - conceptually it is list api. we don't support operations api
* change `self` link for the operation resource


kfangw:
I think that we need more time to consider about getting operation api
So, I postpone merging this.

kfangw:
@anarcher ping

kfangw:
@cmcm2222 Please update the API document
the `self` link of the resource operation is changed

PR #849, opened 2018-11-29
remove self link for operation resource

kfangw:
### Background
Operation Resource has self link like /operations/{hash}
But there is no API support this.

### Solution
Remove self link for operation resource


kfangw:
@spikeekips review please

PR #856, opened 2018-11-29
Add start time to the node info

Geod24:
Allow us to have a better view of nodes running in mainnet


Geod24:
Updated according to @spikeekips ' comment.

PR #858, opened 2018-12-01
Reduce the time to check `BlockOperation` at startup

### LINKED ISSUE #847: `SavingBlockOperations.Check()` takes too long at startup
The elapsed time for `SavingBlockOperations.Check()` will be growing by growing block height.

spikeekips:
### Github Issue

Resolves #847 

### Background

See #847 

### Solution

Up to `100` workers checks the existing Blocks simultaneously.

### DIFF
--- a/lib/node/runner/block_operations.go
+++ b/lib/node/runner/block_operations.go
@@ -1,6 +1,7 @@
 package runner
 
 import (
+	"sync"
 	"time"
 
 	logging "github.com/inconshreveable/log15"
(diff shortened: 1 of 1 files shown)


lib/node/runner/block_operations.go
```diff
@@ -47,58 +47,111 @@ func (sb *SavingBlockOperations) getNextBlock(height uint64) (nextBlock block.Bl
...
+	for i := 1; i <= numWorker; i++ {
+		go sb.checkBlockWorker(i, blocks, errChan)
+	}
+
+	go func() {
+		var blk block.Block
+		for {
+			if blk, err = sb.getNextBlock(blk.Height); err != nil {
+				err = errors.FailedToSaveBlockOperaton.Clone().SetData("error", err)
+				return
```
kfangw: Is it OK to not close the `blocks` here?

spikeekips: @kfangw Oops, you right.

PR #860, opened 2018-12-02
Allow watch with Watcher node

anarcher:
### Background

```
Consensus Node <- Watcher  <- Watcher  ...
```

I think that it's not bad topology.  With this topology, we can protect our consensus nodes. (maybe)

### Solution

- Watcher watches other watchers.

### Possible Drawbacks

- This layered make delayed watching from consensus nodes.. :->

### DIFF
--- a/lib/sync/config.go
+++ b/lib/sync/config.go
@@ -86,6 +86,7 @@ func (c *Config) NewFetcher() Fetcher {
 		c.localNode,
 		func(f *BlockFetcher) {
 			f.fetchTimeout = c.FetchTimeout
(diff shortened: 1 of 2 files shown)


lib/sync/config.go
```diff
@@ -86,6 +86,7 @@ func (c *Config) NewFetcher() Fetcher {
 		c.localNode,
 		func(f *BlockFetcher) {
 			f.fetchTimeout = c.FetchTimeout
+			f.retryInterval = c.RetryInterval
```
Charleslee522: One question. What is it for? :)

Issue #862, opened 2018-12-03
[API] Make `Transaction` resource contains `Operations`


kfangw:
`Transaction` Resource contains `Operations link`.

To retrieve operations, the client should call each `operations link` repeatedly
It leads the client to exceeds the rate limit of the API.

By common sense, when the client calls the transaction API, the client might want to get data about the transaction and related operations.

IMO, hence, the `Transaction` resource would be better to contain operation resources also, not only links.

kfangw:
resolved by #863

PR #863, opened 2018-12-03
tx resource contains operations

kfangw:
### Github Issue
#862 

### Background
see #862 

### Solution
Transaction Resource contains operations


lib/node/runner/api/api.go
```diff
@@ -106,27 +103,3 @@ func TriggerEvent(st *storage.LevelDBBackend, transactions []*transaction.Transa
 	}
 
 }
-
-func renderEventStream(args ...interface{}) ([]byte, error) {
```
kfangw: the renderEventStream function is used only by the subscribe API
and it is impossible to get data from storage with this design.
Hence, this function is moved into subscribe API handler and let the function use the closure.

lib/node/runner/api/resource/transaction.go
```diff
@@ -27,6 +31,7 @@ func (t Transaction) GetMap() hal.Entry {
 		"sequence_id":     t.bt.SequenceID,
 		"created":         t.bt.Created,
 		"operation_count": len(t.bt.Operations),
+		"operations":      t.tx.B.Operations,
```
kfangw: the shape of `operation` in the transaction resource may differ from `operation resource`.
But I think that the `operation resource` will be deprecated and relevant APIs no longer under maintenance.

kfangw:
review please @spikeekips @anarcher @Geod24

lib/node/runner/api/stream.go
```diff
@@ -41,7 +43,33 @@ func (api NetworkHandlerAPI) PostSubscribeHandler(w http.ResponseWriter, r *http
...
 	}
 
-	es := NewEventStream(w, r, renderEventStream, DefaultContentType)
+	rederFunc := func(args ...interface{}) ([]byte, error) {
+		if len(args) <= 1 {
+			return nil, fmt.Errorf("render: value is empty") //TODO(anarcher): Error type
+		}
+		i := args[1]
+
+		if i == nil {
```
Geod24: This check should be folded into line 47 so the error is the same

lib/node/runner/api/stream.go
```diff
@@ -41,7 +43,33 @@ func (api NetworkHandlerAPI) PostSubscribeHandler(w http.ResponseWriter, r *http
...
+			return []byte{}, nil
+		}
+
+		switch v := i.(type) {
```
Geod24: Will this `panic` if no type matches ?

lib/node/runner/api/transaction.go
```diff
@@ -24,25 +24,25 @@ func (api NetworkHandlerAPI) GetTransactionsHandler(w http.ResponseWriter, r *ht
...
-				firstCursor = append(firstCursor, c...)
-			}
-			txs = append(txs, resource.NewTransaction(&t))
+	var txs []resource.Resource
+	iterFunc, closeFunc := block.GetBlockTransactions(api.storage, options)
+	for {
+		t, hasNext, c := iterFunc()
+		if !hasNext {
+			break
+		}
```
Geod24: This should be part of the `for` condition, instead of infinite loop + break

lib/node/runner/api/transaction.go
```diff
@@ -52,28 +52,28 @@ func (api NetworkHandlerAPI) GetTransactionByHashHandler(w http.ResponseWriter,
...
-		if !found {
-			return nil, errors.BlockTransactionDoesNotExists
-		}
-		bt, err := block.GetBlockTransaction(api.storage, key)
-		if err != nil {
-			return nil, err
-		}
-		payload = resource.NewTransaction(&bt)
-		return payload, nil
+	found, err := block.ExistsBlockTransaction(api.storage, key)
```
Geod24: Instead of `exists` + `get`, why don't you get it directly ?

kfangw: To distinguish error in the database and no error but not exist.

kfangw:
Ping @anarcher @spikeekips @Geod24

kfangw:
@cmcm2222 When this merged, please update the API document according to this change.

kfangw:
ping @anarcher @Geod24

lib/node/runner/api/transaction.go
Geod24: [The underlying method already does that](https://github.com/bosnet/sebak/blob/0cc0ebdf63f07b1b2e5dced39ec25b5dfe82dddb/lib/storage/leveldb.go#L136-L162)

kfangw:
> @cmcm2222 Could you test this PR and update the Document?
> I want to merge this when you are ready for the document.

@cmcm2222 ping

cmcm2222:
@kfangw OK.  I will check it

PR #864, opened 2018-12-03
Rename GetBlockOperationsByTxHash to GetBlockOperationsByTx

Geod24:
Because hash is implied when getting an association.


soonkuk:
LGTM!

PR #883, opened 2018-12-20
load CommonAccountAddress with sync start

anarcher:
### Github Issue

`common/Config.CommonAccountAddress` is not loaded when `common.Config` is created. So sync has to load `CommonAccountAddress` and add it to `Config.CommonAccountAddress`

### Background

Block height 348204 has `inflation-pf` operation, and this operation validation needs `common.Config.CommonAccountAddress`. 

### Solution

`sync/Config` loads `CommonAccountAddress` and add it to `common/Config.CommonAccountAddress`.

### Possible Drawbacks

IHMO, `common/Config` itself loads it's own attributes when it is created makes sense.

### DIFF
--- a/cmd/sebak/cmd/run.go
+++ b/cmd/sebak/cmd/run.go
@@ -756,7 +756,10 @@ func runNode() error {
 
 	tp := transaction.NewPool(conf)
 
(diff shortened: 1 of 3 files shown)


soonkuk:
LGTM

PR #885, opened 2018-12-21
Support Time Sync

### LINKED ISSUE #571: Synchronize the time of all nodes
There are two factors of the time in `SEBAK` periodically.
The time is used,
1. to check ballot well-formed.
1. to adjust block time to 5 sec.

spikeekips:
### Github Issue
Resolves #571 

### Background

Time is very important in SEBAK, the creation time of ballot and transaction time is used for its validation and node discovery also validates its time.

If time is different from the other nodes,
* The node will not join the network, discovery will be failed
* The ballots from problem node will be ignored

### Solution

At starting time, `common.TimeSync` compares local time with the NTP server(`--ntp`). If it's different is over `common.MaxTimeDiffAllow`(4 seconds by default), the node tries to sync time using `--time-sync-command` command.

The `TimeSync` will check and sync time periodically(every 1 minute).

#### Changes

* `--ntp` option added; set the NTP server, by default `time.bora.net`
* `--time-sync-command` option added; this command will be executed when time difference is over `common.MaxTimeDiffAllow`. For example, `rdate time.bora.net` will be used.

### DIFF
--- a/lib/common/time.go
+++ b/lib/common/time.go
@@ -1,6 +1,13 @@
 package common
 
-import "time"
+import (
+	"bytes"
+	"fmt"
+	"os/exec"
+	"time"
+
+	"github.com/beevik/ntp"
+)
 
 const (
 	TIMEFORMAT_ISO8601 string = "2006-01-02T15:04:05.000000000Z07:00"
@@ -16,4 +23,93 @@ func NowISO8601() string {
 
 func ParseISO8601(s string) (time.Time, error) {
 	return time.Parse(TIMEFORMAT_ISO8601, s)
+}
+
+type TimeSync struct {
+	ntpServer string
+	syncCmd   string
+}
+
+func NewTimeSync(ntpServer string, syncCmd string) (*TimeSync, error) {
+	ts := &TimeSync{
+		ntpServer: ntpServer,
+		syncCmd:   syncCmd,
+	}
+
+	if _, err := ts.Query(); err != nil {
+		return nil, err
+	}
+
+	return ts, nil
+}
+
+func (ts *TimeSync) Start() {
+	log.Debug("starting to sync time periodically")
+
+	go func() {
+		for _ = range time.NewTicker(time.Second * 60).C {
+			go func() {
+				if err := ts.Sync(); err != nil {
+					log.Error("failed to sync time", "error", err)
+				}
+			}()
+		}
+	}()
+}
+
+func (ts *TimeSync) Query() (*ntp.Response, error) {
+	return ntp.Query(ts.ntpServer)
+}
+
+func (ts *TimeSync) checkNTPOffset() (bool, error) {
+	resp, err := ts.Query()
+	if err != nil {
+		return false, err
+	}
+
+	log.Debug("check time difference", "offset", resp.ClockOffset, "allowed", MaxTimeDiffAllow)
+	if resp.ClockOffset > MaxTimeDiffAllow || resp.ClockOffset < (MaxTimeDiffAllow*-1) {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+func (ts *TimeSync) Sync() error {
+	if len(ts.syncCmd) < 1 {
+		return nil
+	}
+
+	log.Debug("trying to sync time")
+
+	var b bytes.Buffer
+	cmd := exec.Command("sh", "-c", ts.syncCmd)
+	cmd.Stderr = &b
+	cmd.Stdout = &b
+
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to run time-sync-command: `%s`: %s", ts.syncCmd, b.String())
+	} else {
+		log.Debug("time-sync-command was executed", "output", b.String())
+	}
+
+	var err error
+	var synced bool
+	for i := 0; i < 3; i++ {
+		time.Sleep(time.Second * 3)
+		if synced, err = ts.checkNTPOffset(); err != nil {
+			return err
+		}
+		if synced {
+			break
+		}
+	}
+
+	if !synced {
+		return fmt.Errorf("time-sync-command was executed, but still not synced")
+	}
+
+	log.Debug("time synced")
+
+	return nil
 }
--- a/lib/common/time.go
+++ b/lib/common/time.go
@@ -1,9 +1,7 @@
 package common
 
 import (
-	"bytes"
 	"fmt"
-	"os/exec"
 	"time"
 
 	"github.com/beevik/ntp"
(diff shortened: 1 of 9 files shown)


lib/common/time.go
```diff
@@ -17,3 +24,92 @@ func NowISO8601() string {
...
+
+func (ts *TimeSync) Sync() error {
+	if len(ts.syncCmd) < 1 {
+		return nil
+	}
+
+	log.Debug("trying to sync time")
+
+	var b bytes.Buffer
+	cmd := exec.Command("sh", "-c", ts.syncCmd)
```
Geod24: `sebak` really shouldn't be running `sh` on our computers...

Geod24: And it even uses `flagTimeSyncCommand` directly, which means it will run any command provided by the user as `sebak` ? What happens when someone exploit that by making a clueless user exports the wrong command ? This is a huge security hole.

spikeekips: @Geod24 Yes, as you pointed, it is not safe only when anonymous can set that command. It only can be set by a node operator. It is totally up to a node operator's choice.

Geod24: The thing is, node operators have varying degree of knowledge. Even the most experienced ones can fall victim of scams / trick, that's why we should aim to reduce the area of attack. Same reasoning applies to bugs. And exposing a way to run a command through the shell is a big no.

Last, but not least, it's not portable to Windows.

spikeekips: @Geod24 Yes, current one is not good for Windows. I can change this part to support Windows.

spikeekips:
@Geod24 I can not understand what you mean, running `sh` can be a security breach. The 'time sync command' is executed with `sh`, but it only set by the node operator and it runs under the system account. It can not be accessed from a remote.

Geod24:
@spikeekips :
Using `sh` means that the damage one can do if this part of the code is compromised is virtually unlimited. By doing a direct shell call, you open the possibility, if someone was to find a buffer overflow / way to corrupt the data to have any arbitrary command run. It does not in itself means that it's less secure, but it means some data corruption would be absolutely catastrophic.

Second, doing this adjustment at a global level for a single program is bad. Imagine if two programs were to do this, what would happen if they don't sync to the same time ? Your clock would constantly be shifting, and that would most likely affect other programs. Which is a big no.

And as it turns out, there is already a way to do this. It's the NTP daemon available on each system. Something that respects the user preference and just works.

spikeekips:
> @Geod24 : Using sh means that the damage one can do if this part of the code is compromised is virtually unlimited. By doing a direct shell call, you open the possibility, if someone was to find a buffer overflow / way to corrupt the data to have any arbitrary command run. It does not in itself means that it's less secure, but it means some data corruption would be absolutely catastrophic.

Using which sync command is under control of a node operator. If his/her command has (security) problem, he/she owns that risk. The node operator should be responsible for choosing valid command.

> Second, doing this adjustment at a global level for a single program is bad. Imagine if two programs were to do this, what would happen if they don't sync to the same time ? Your clock would constantly be shifting, and that would most likely affect other programs. Which is a big no.

If multiple system processes run from the same shell process, synced time can affect to the others, but this can be solved by the notice or guide to the node operator.

> And as it turns out, there is already a way to do this. It's the NTP daemon available on each system. Something that respects the user preference and just works.

Originally running NTP daemon in the system is an ideal solution. This PR is the minimum guarantee to keep the consensus to work. Basically, the time syncing process consists of 2 parts. First, checking time offset from time server at starting time and next is periodically syncing time. As I mentioned, a node should check whether it is under the correct time for participating consensus.

Geod24:
> Using which sync command is under control of a node operator. If his/her command has (security) problem, he/she owns that risk. The node operator should be responsible for choosing valid command.

That was not the point. We are creating an easy to exploit attack vector. It is like limiting a password to lowercase letters only and then blaming the user for not using one secure enough...

> If multiple system processes run from the same shell process, synced time can affect to the others, but this can be solved by the notice or guide to the node operator.

Noticing the operator is not solving the problem.

> Originally running NTP daemon in the system is an ideal solution. This PR is the minimum guarantee to keep the consensus to work. Basically, the time syncing process consists of 2 parts. First, checking time offset from time server at starting time and next is periodically syncing time. As I mentioned, a node should check whether it is under the correct time for participating consensus.

The protocol should not behave incorrectly if the time sync is not guaranteed. The node should merely not be alive, and the network should be alive as long as the number of nodes is under the fault tolerance threshold.

spikeekips:
>> @Geod24 : Using which sync command is under control of a node operator. If his/her command has (security) problem, he/she owns that risk. The node operator should be responsible for choosing a valid command.
>
>That was not the point. We are creating an easy to exploit attack vector. It is like limiting a password to lowercase letters only and then blaming the user for not using one secure enough...

I also think setting the external command to set the time, is just like a compromise plan. SEBAK needs to be based on the synced time and I believe this approach can be a good solution at this time.

>> If multiple system processes run from the same shell process, synced time can affect to the others, but this can be solved by the notice or guide to the node operator.
>
>Noticing the operator is not solving the problem.

Hmm, I don't know who can be in charge of operating node and it's the system. Setting external command is holy managed by him/her and its potential risk will be owed to him/her. We can guide and provide the proper way to set the safe way.

>> Originally running NTP daemon in the system is an ideal solution. This PR is the minimum guarantee to keep the consensus to work. Basically, the time syncing process consists of 2 parts. First, checking time offset from time server at starting time and next is periodically syncing time. As I mentioned, a node should check whether it is under the correct time for participating consensus.
>
>The protocol should not behave incorrectly if the time sync is not guaranteed. The node should merely not be alive, and the network should be alive as long as the number of nodes is under the fault tolerance threshold.

There are no ways to sync time on a cosmic level. NTP has been the de facto standard to sync time in the internet level and is believed to work well. If we can not find a better solution, NTP is the only option.

spikeekips:
> @Geod24 : The whole point I am making here is exactly that. Let's use NTP. But do not set the system time through sebak. If you want to rely on system time, ask the operator to correctly configure his NTP daemon. If you don't want to rely on the operator properly configuring his NTP daemon, provide a virtual clock in sebak which is NTP sync, but do not change the system's clock from sebak.

We can reach the CONSENSUS :) At first, I also think that the syncing (system) time is done by a system(as you said, NTP daemon like that), so this PR just checks the system time is correct using NTP. In this PR, if the external command is empty(by default **empty** 😄), there is no further action. Node operator intentionally will set the external command; he/she will take responsibility for the security concerns.

lib/common/time.go
```diff
@@ -17,3 +29,106 @@ func NowISO8601() string {
...
+		return err
+	}
+
+	go func() {
+		for _ = range time.NewTicker(timeSyncInterval).C {
+			if synced, err := ts.checkNTPOffset(); err != nil {
+				ts.log.Error("failed to check ntp offset", "error", err)
+			} else if synced {
+				ts.log.Debug("time synced")
+				return
```
kfangw: Isn't it should be `continue` instead of `return`?

kfangw: I expect that the `TimeSync` runs continuously until the node shutdown.

spikeekips: @kfangw Yes you right.

Geod24:
@spikeekips : I don't agree it's even remotely close to a solution, let alone a good one. But since I seem to be the only one thinking so, feel free to merge.

PR #886, opened 2018-12-21
Support log rotate

### LINKED ISSUE #873: Log rotating HTTP log

spikeekips:
### Github Issue

Resolves #873 

### Background

In debug mode, SEBAK produces huge log messages including HTTP requests and responses log. Under docker, we can use `max-size` setting in log driver, but a node can keep the entire log messages, it's the matter of choice.

### Solution

There are 2 logs, node log, and HTTP log. Each log will have rotate options,

* `--log-rotate-max-size`, `--http-log-rotate-max-size`: if log file is over, rotate log
* `--log-rotate-max-age`, `--http-log-rotate-max-age`: if log file is over days, rotate log
* `--log-rotate-max-count`, `--http-log-rotate-max-count`: if the number of rotated logs over, remove olds
* `--log-rotate-uncompress`, `--http-log-rotate-uncompress`: compress or not

### DIFF
--- a/lib/common/log.go
+++ b/lib/common/log.go
@@ -3,11 +3,13 @@ package common
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"reflect"
 	"time"
 
 	logging "github.com/inconshreveable/log15"
+	lumberjack "gopkg.in/natefinch/lumberjack.v2"
 
 	"boscoin.io/sebak/lib/errors"
 )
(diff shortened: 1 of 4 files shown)


lib/common/log.go
```diff
@@ -112,3 +114,27 @@ func (l nopLogger) Info(msg string, ctx ...interface{})  {}
...
+type closingHandler struct {
+	io.WriteCloser
+	logging.Handler
+}
+
+func (h *closingHandler) Close() error {
+	return h.WriteCloser.Close()
+}
+
+type RotateHandler struct {
```
Geod24: Unused ?

Geod24:
I thought you wanted to use logrorate directly ? At least that's what we discussed in [the issue](https://github.com/bosnet/sebak/issues/873#issuecomment-445815875)

spikeekips:
@Geod24 Hmm, `logrotate` utility is one of our options. It can be a simple solution to support a rotating log, but under docker, environment to make `logrotate` to be working needs some additional efforts. I think the rotating log should be the native feature for considering the various deploying environments.

Geod24:
@spikeekips : What kind of additional effort ? It should be a few (<10) LoC in sebak and a configuration file.

spikeekips:
@Geod24 :) For instance, like these 
https://www.google.co.kr/search?q=do+ker+logrotate&ie=UTF-8&oe=UTF-8&hl=en-kr&client=safar . Basicslly docker runs one entrypoint, so the daemon like logrotate should be deployed by task manager in docker.

PR #893, opened 2018-12-30
Save BlockTransaction and BlockOperation when sync 

anarcher:
### Background

> The operation of above a blockheight is congress-voting-result and it needs to check the operation of congress-voting which should be saved already.

This PR intent that when syncer finishes a block, the operations and the transactions of this block are saved with finishing the block in sync/watcher node.

### Solution

when syncer finishes a block, the operations and the transactions of this block are saved with finishing the block in sync/watcher node.

### Possible Drawbacks

Without `SavingBlockOperations`, It makes slower  sync/watcher time than with `SavingBlockOperations`.


anarcher:
> Dependency changes are unrelated?

@Geod24 Oh, I removed this dependency changes in comments :->

anarcher:
Ping~

PR #896, opened 2019-01-03
Reset si.Bts in fetch function

### LINKED ISSUE #895: fetcher add tx to si.Bts repeatedly
In watcher mode, fetcher function can add tx to si.Bts repeatedly if some error occurs in validation.
It can result mismatched block hash error.

soonkuk:
### Github Issue
Resolves #895.

### Solution
Reset si.Bts when fetch function is called.

### DIFF
--- a/lib/sync/fetcher.go
+++ b/lib/sync/fetcher.go
@@ -97,6 +97,7 @@ func (f *BlockFetcher) fetch(ctx context.Context, si *SyncInfo) error {
 		height    = si.Height
 		nodeAddrs = si.NodeAddrs()
 	)
+	si.Bts = si.Bts[:0]
 	f.logger.Debug("start fetch", "height", height, "nodes", nodeAddrs)
 
 	if len(nodeAddrs) <= 0 {


anarcher:
Thanks! @soonkuk , I missed this. T_T

lib/sync/fetcher.go
```diff
@@ -97,6 +97,7 @@ func (f *BlockFetcher) fetch(ctx context.Context, si *SyncInfo) error {
 		height    = si.Height
 		nodeAddrs = si.NodeAddrs()
 	)
+	si.Bts = si.Bts[:0]
```
Charleslee522: This is just initialize `si.Bts`, right?

anarcher: @Charleslee522 It's that a slice shares the same backing array and capacity as the original, so the storage is reused.

Charleslee522: @anarcher I got it. 👍 thanks
https://tour.golang.org/moretypes/11

soonkuk: @Charleslee522 Yes it is

PR #898, opened 2019-01-04
Refactor/document hash-related functions

Geod24:
A rather minor PRs with a few commits that have been longing on other PRs for a while that I'd like to get in instead of constantly rebasing them :)

### DIFF
--- a/lib/common/hash.go
+++ b/lib/common/hash.go
@@ -38,9 +38,4 @@ func MustMakeObjectHashString(i interface{}) string {
 }
 
 type Hash = ethcommon.Hash
-
-func BytesToHash(b []byte) Hash {
-	var h Hash
-	h.SetBytes(b)
-	return h
-}
+var BytesToHash = ethcommon.BytesToHash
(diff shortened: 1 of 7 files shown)


lib/common/hash.go
```diff
@@ -1,3 +1,8 @@
+//
+// This package defines hash-related functions used in Sebak
+// The primary hash function used in Sebak is a double sha256,
+// like Bitcoin.
+//
```
Charleslee522: 👍

lib/common/hash.go
```diff
@@ -7,12 +12,26 @@ import (
...
+
+// Generate a double SHA-256 hash out of the supplied binary data
 func MakeHash(b []byte) []byte {
 	first := sha256.Sum256(b)
 	second := sha256.Sum256(first[:])
 	return second[:]
 }
 
+// Interface recognized by RLP encoder
+type Encoder = rlp.Encoder
```
spikeekips: `Encoder` is not used, is it for future usage?

Geod24: Yes!

PR #899, opened 2019-01-04
Operation: Use slightly more clever code to deserialize body

Geod24:
```
Instead of copy-pasting blocks of code inside a `switch`,
we can just have a function instantiating the type and
greatly reduce duplication.

As `Body` is defined to be an interface to value instead
of an interface to pointer, a bit of magic is needed
with the `reflect` module to make it work.
```

### DIFF
--- a/lib/transaction/operation/operation.go
+++ b/lib/transaction/operation/operation.go
@@ -2,6 +2,7 @@ package operation
 
 import (
 	"encoding/json"
+	"reflect"
 
 	"boscoin.io/sebak/lib/common"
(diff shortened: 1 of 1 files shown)


Geod24:
@spikeekips : For this usage there will really have no impact, because we're essentially dereferencing a pointer.
We are already depending on `reflect` very heavily through `json` and `rlp` encoding, to name a few things, who are doing real work with reflect (like looking up struct's fields).
Also, as obvious in #836, it allows to reuse the same code for JSON and RLP decoding instead of copy/pasting this function.

PR #903, opened 2019-01-08
Remove EncodeList, use anon struct instead

Geod24:
```
By shifting the type to the call site, we can greatly simplify the code
without loosing in readability.
```

### DIFF
--- a/lib/common/hash.go
+++ b/lib/common/hash.go
@@ -6,7 +6,6 @@
 package common
 
 import (
-	"bytes"
 	"crypto/sha256"
 	"io"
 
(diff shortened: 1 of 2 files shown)


Geod24:
> `totalLength` is necessary?

In which regard ? I left the variable because it makes it slightly more obvious what it is for. Can remove it if you want.

Geod24:
> To make it simpler, how about providing EncodeRLP/DecodeRLP pair to Header and Body of the Operation?

I thought about this, but I am wondering if it would not imply too much copying. I will look into it for further improvement, thanks for the idea!

Issue #842, opened 2018-11-28
Database migration strategy


kfangw:
- Related Issue
#834 

- Background
* Invariant `internal` data structure is only Block and Transaction
* Block
    * 	Version
    *	PrevBlockHash
    *	TransactionsRoot
    *	ProposedTime
    *	Height
    *	TotalTxs
    *	TotalOps
* Transaction
    *  Source
    *  Fee
    *  SequenceID
    *  Operations

Except above data structure and its fields, every other can be changed.
IMHO,
BlockAccount and BlockOperation data structure are the caches for speed up getting data.
It means that these data structure can be changed anytime.
For example, Index or model for the data.

To support this, we need a database migration strategy.
Here is a suggestion. It is originated from @anarcher 's idea

* The database structure is changed
* Launch a node as `watcher` mode with the change.
* Every validator is watched by at least one watcher node
* Wait until the watcher node sync all data from validator node in the changed data structure
* Kill the validator and elaborate the watcher node to the validator

I want to listen to your opinion that this scenario can be possible or not.
@spikeekips @anarcher @zzim2x @outsideris @Geod24 

Especially @zzim2x 's opinion is very crucial.

spikeekips:
At first, 👍 .

This suggestion is focused on the maintenance by us, if watcher node operator see the db migration, the story should be changed. Of course, they can use this way, but we should provide the explicit way for them, I think.

This suggested way and plus, I suggest new sub command, `$ sebak db migrate` command, it checks the existing db status and if needed, migrate to the latest version.

kfangw:
> At first, 👍 .
> 
> This suggestion is focused on the maintenance by us, if watcher node operator see the db migration, the story should be changed. Of course, they can use this way, but we should provide the explicit way for them, I think.
> 
> This suggested way and plus, I suggest new sub command, `$ sebak db migrate` command, it checks the existing db status and if needed, migrate to the latest version.

`sebak db migrate` command is good 👍 

To do, we need to manage and refactor the `block` package to be more suitable for this purpose.

Geod24:
> Except above data structure and its fields, every other can be changed.

[...]

PR #904, opened 2019-01-09
Enable binary serialization of BlockOperation (+ add upgrade command)

Geod24:
**~Still working / testing this, it should be ready but I want hard data on the amount of space saved~**

~Please review commit-by-commit.~
~The most important commit is the last one (f52a7c5 ATM).~

This patch enables binary serialization on `BlockOperation`, saving some disk space.
```
In the LevelDB serialization code, `BinaryMarshaller` takes priority over `TextMarshaller`.
This allows us to simply switch on binary serialization with a few lines of code.

However, it is a breaking change for existing deployments,
as old text-serialized data won't be deserializable.

There are multiple way to approach this problem, among other things:
- Trying one of the method, then falling back if it doesn't work
- Setting a given block height in the future for the transition
- Replay all the transaction data to recreate the DB
- Force conversion of all the existing data

We choose, for simplicity, and to help reducing existing load on nodes,
to go with the later approach.
We introduce `sebak upgrade`, which will do the job only if needed.
In addition, a check was introduced in `run`,
to avoid the node starting with the new code and old data,
which could lead to memory corruption or crash.

Note that the code in `sebak upgrade` is not guaranteed to work in a future release.
Doing so would require specifying exactly what transformation is to be made,
which would overly complicate the process for little benefit.
It can be done in the future if the user base justifies the need.
```

### DIFF
--- a/lib/block/operation.go
+++ b/lib/block/operation.go
@@ -9,6 +9,8 @@ import (
 	"boscoin.io/sebak/lib/storage"
 	"boscoin.io/sebak/lib/transaction"
 	"boscoin.io/sebak/lib/transaction/operation"
+
+	"github.com/ethereum/go-ethereum/rlp"
 )
 
 // BlockOperation is `Operation` data for block. the storage should support,
@@ -34,6 +36,16 @@ type BlockOperation struct {
 	seqID   uint64
 	linked  string
 	isSaved bool
+}
+
+// Implement `encoding.BinaryMarshaler`
+func (bo BlockOperation) MarshalBinary() (data []byte, err error) {
+	return rlp.EncodeToBytes(bo)
+}
+
+// Implement `encoding.BinaryUnmarshaler`
+func (bo *BlockOperation) UnmarshalBinary(data []byte) error {
+	return rlp.DecodeBytes(data, bo)
 }
 
 func NewBlockOperationKey(opHash, txHash string) string {
--- a/cmd/sebak/cmd/run.go
+++ b/cmd/sebak/cmd/run.go
@@ -755,6 +755,11 @@ func runNode() error {
 		log.Crit("failed to initialize storage", "error", err)
 		return err
 	}
+	// Check that no upgrades are needed
+	if needsDBUpgrade(st) {
+		log.Crit("cannot start node - Upgrades are needed. Run sebak upgrade [--storage ...]")
+		return fmt.Errorf("Upgrades to the database needed")
+	}
 
 	// get the initial balance of geness account
 	initialBalance, err := runner.GetGenesisBalance(st)
--- a/cmd/sebak/cmd/upgrade.go
+++ b/cmd/sebak/cmd/upgrade.go
@@ -1,0 +1,113 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/syndtr/goleveldb/leveldb/util"
+
+	cmdcommon "boscoin.io/sebak/cmd/sebak/common"
+	"boscoin.io/sebak/lib/block"
+	"boscoin.io/sebak/lib/common"
+	"boscoin.io/sebak/lib/storage"
+)
+
+var (
+	flagStorage string
+)
+
+func init() {
+	upgradeCmd := &cobra.Command{
+		Use:   "upgrade",
+		Short: "Upgrade the DB from a previously existing version (or do nothing)",
+		Run:   runUpgrade,
+	}
+
+	flagStorage = common.GetENVValue("SEBAK_STORAGE", cmdcommon.GetDefaultStoragePath(upgradeCmd))
+	upgradeCmd.Flags().StringVar(&flagStorage, "storage", flagStorage, "storage uri")
+
+	rootCmd.AddCommand(upgradeCmd)
+}
+
+// Entry point of the command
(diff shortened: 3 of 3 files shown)


spikeekips:
Before commenting, this is the good start for building stretegy of upgrading db and migration. I wish this PR will help to make the upgrade(or db migration) process.

lib/block/operation.go
```diff
@@ -38,6 +40,16 @@ type BlockOperation struct {
...
+// Implement `encoding.BinaryMarshaler`
+func (bo BlockOperation) MarshalBinary() (data []byte, err error) {
+	return rlp.EncodeToBytes(bo)
+}
+
+// Implement `encoding.BinaryUnmarshaler`
+func (bo *BlockOperation) UnmarshalBinary(data []byte) error {
+	return rlp.DecodeBytes(data, bo)
+}
+
```
anarcher, line 50: I like this approach! (encoding. BinaryMarshaler)  :-D Thanks, @Geod24

anarcher (approved):

spikeekips:
the `upgrade` cmd should be consider #842, db migration strategy, so instead of `upgrade` cmd and before building strategy, I think to provide migration script will be enough at this time.

anarcher (changes requested):
I'm not sure that RLP is good serialization for storage and P2P.

- "The RLP format is overcomplicated" :  https://github.com/ethereum/wiki/wiki/Wishlist#rlp 
- P2P Message Serialization: https://notes.ethereum.org/15_FcGc0Rq-GuxaBV5SP2Q
- vbuterin's suggestion  Simple Serialize(SSZ) :  https://github.com/ethereum/research/tree/master/py_ssz

anarcher:
- https://github.com/prysmaticlabs/prysm/issues/150
- https://ethresear.ch/t/discussion-p2p-message-serialization-standard/2781/2
- https://github.com/ethereum/beacon_chain/blob/master/beacon_chain/utils/simpleserialize.py

anarcher:
I am not sure but RLP hashing but using other sereilzation format Like http://prysmaticlabs.com is not bad.

For now backward compatibility, l intent that for storage and p2p

spikeekips (changes requested):
Please check https://github.com/bosnet/sebak/pull/904#issuecomment-452967401 .

Geod24:
> I think to provide migration script will be enough at this time.

I'm not sure what you mean by migration script ? I plan to remove `apply20190107UpgradeBlockOperation` and related code after a while.

spikeekips:
> > I think to provide migration script will be enough at this time.
> 
> @Geod24 : I'm not sure what you mean by migration script ? I plan to remove `apply20190107UpgradeBlockOperation` and related code after a while.

I mean, the migration script can be written easily with separate go application(if possible, python can do), I think we need to develop how to upgrade storage.

Geod24:
@spikeekips : I thought so at first, but the problem I had is `needsDBUpgrade`. I wanted to provide a way to tell user they need to run the script. And since the logic is similar, it was just simpler to put it in sebak (also, sebak has all the parsing utilities).

spikeekips:
@Geod24 Before we have a more concrete plan for upgrading, we can provide the upgrade script and guide for users. I think it will be enough. To watcher user, we can support the upgrade script and guide.

Geod24:
Tested on a dump of the MainNet DB:

Before, `du` reports a size of `11,216,008` (5.3G), and after, `10,743,864` (5.1G), so about 5% gain.
It is quite surprising that the gains are so small, I will look into what is using so much space then.

Geod24:
> @Geod24 Before we have a more concrete plan for upgrading, we can provide the upgrade script and guide for users. I think it will be enough. To watcher user, we can support the upgrade script and guide.

What do you mean by script ? We need to have it integrated in sebak.

Geod24:
Ping

Charleslee522 (approved):
It's only about 5%, but I think it's a good attempt.

spikeekips:
@Geod24 I think the current storage data need more enhancements, so this patch is meaningful. I wish the storage upgrade(or migration process) should be a systematic process like flyway(spring) or django db migration(python). As I suggested, before systematic migration, the simple script for this PR seems to be enough.

lib/block/operation.go
```diff
@@ -36,6 +38,16 @@ type BlockOperation struct {
 	isSaved bool
 }
 
+// Implement `encoding.BinaryMarshaler`
+func (bo BlockOperation) MarshalBinary() (data []byte, err error) {
```
kfangw, line 42: The migration should be applied before this code deployed. right?
This should be handled carefully.

kfangw (approved):
LGTM.
But please care about what I left below.

lib/block/operation.go
Geod24: The migration *is* this code. `sebak` will refuse to start if the migration isn't applied. See the changes to `cmd/sebak/cmd/run.go`.

kfangw: I mean that the upgrade command should be applied to exiting database before the node deployed with this commit.

Did you mentioned this critical point to SRE team?

I just worried that.

Geod24: > Did you mentioned this critical point to SRE team?

I didn't see any reason to do so, as long as this PR is not merged. And even so, the error message is pretty clear.

Geod24:
@spikeekips : This has 2 approval. Should I merge it or close it ?

PR #916, opened 2019-01-21
[WIP] IAVL store wrapper

kfangw:


anarcher:
How about that we make a branch for it? and make a pr of this branch (for master rebasing) :->

anarcher:
IMHO, `state.DB` and `StateDB` are confusing by same naming`DB`.     I would like have layering naming. e.g. `State` -(use)-> `Store` -(use)-> `DB(leveldb,rocksdb)`.

anarcher:
I am not clear that we have account (only) based states. 
As you know, Cosmos style is that multi-substores.  With this approaches, we can have accounts and contracts and storages each other. (not having account.Code and account.Storages).

Block.StateRoot = MultiStore's Roothash and each of substores can be  accounts(one of StoreKVKey) and contracts, storages... (staking or gov...)

kfangw:
> I am not clear that we have account (only) based states?
> As you know, Cosmos style is that multi-substores. With this approaches, we can have accounts and contracts and storages each other. (not having account.Code and account.Storages).
> 
> Block.StateRoot = MultiStore's Roothash and each of substores can be accounts(one of StoreKVKey) and contracts, storages... (staking or gov...)

@anarcher 
Good idea.
Balance, Code, and Storage can be divided into substores.
But, for the Storage, each account can hold multiple (k,v) pairs.
To solve this, IMO, PrefixDB, PrefixStore may be adaptable.

/balance/(address, balance) pair for a data item
/code/(address, code) pair
/storage/(addressprefixed data key, value) pair

Is it makes sense?

This is a example code snippet
```

func TestMulti(t *testing.T) {
	db := dbm.NewMemDB()

	st := store.NewCommitMultiStore(db)

	key1 := sdk.NewKVStoreKey("balance")
	key2 := sdk.NewKVStoreKey("code")
	key3 := sdk.NewKVStoreKey("storage")

	st.MountStoreWithDB(key1, sdk.StoreTypeIAVL, nil)
	st.MountStoreWithDB(key2, sdk.StoreTypeIAVL, nil)
	st.MountStoreWithDB(key3, sdk.StoreTypeIAVL, nil)

	err := st.LoadLatestVersion()
	require.Nil(t, err)
	
	addrExample := []byte{0x00, 0x00, 0x01, 0x02, 0xFF}

	balanceStore := st.GetCommitKVStore(key1)
	codeStore := st.GetCommitKVStore(key2)
	storageStore := st.GetCommitKVStore(key3)
	theAccountStorageStore := storageStore.Prefix(addrExample)
	
	
	balanceStore.Set(addrExample, []byte("1000"))
	codeStore.Set(addrExample, []byte("print hello world"))
	theAccountStorageStore.Set([]byte("variableKey"), []byte("variableValue"))
	
	commitID := st.Commit()
	
	t.Log(commitID.Hash) //Hash for root Storage
	t.Log(balanceStore.LastCommitID().Hash)
	t.Log(codeStore.LastCommitID().Hash)
	t.Log(storageStore.LastCommitID().Hash)
}
```

anarcher:
My first about storage is that storage treated as []byte from storage type. ( accAddress -> storage bytes ) 

It means that we storage these items saved one byte array in store (with amino) So we don’t need to prefix if it cares like this. But l am not sure about it. We need to have times to consider it :-)

PR #919, opened 2019-01-30
Adapt the testsuite to use SSE

### LINKED ISSUE #909: Use Server Sent Event for integration tests
Currently, the tests use the API, which forces it to follow a pull model: whenever we want to get informations about an item in the blockchain (usually, the balance of account), we have to query, see if the result is what we expected, sleep a bit, and try again.

This is not efficient, nor scalable and we should switch to using `/subscribe` instead.

Geod24:
Also includes two improvements to SSE themselves.

Fixes #909


tools/accstreamer/main.go
```diff
@@ -0,0 +1,88 @@
...
+	var addresses []string
+	for i := 0; i < len(args); i += 2 {
+		addresses = append(addresses, args[i])
+		exps = append(exps, Expectation{ address: args[i], balance: common.MustAmountFromString(args[i + 1]) })
+	}
+
+	cli := client.MustNewClient(server)
+	ctx, cancel := context.WithCancel(context.Background())
+
+	// Handler function
```
Charleslee522: Do we need this comment?

tools/accstreamer/main.go
```diff
@@ -0,0 +1,88 @@
...
+	handler := func(tx client.Account) {
+		// We log the changes so if something fail, we have an history of what the client saw
+		tnow := time.Now();
```
Charleslee522: There is a semicolon in the end of the line. Is this needed?

tools/accstreamer/main.go
```diff
@@ -0,0 +1,88 @@
...
+			// This is an address we care about
+			if exp.address == tx.Address {
+				if exp.balance != current {
+					// Not the right balance yet, bail out
+					return
+				}
+				exps[idx].address = ""
+				// If there's still active watcher,
+				// there is no point in iterating further
+				if activeWatcher {
```
Charleslee522: How about making it `activeWatcher == true` or modifying 79 line for code consistency?

Charleslee522: Could you explain what does the `activeWatcher` mean? :)

Geod24: Sure! So this utility can listen to multiple accounts at the same time. `activeWatcher == true` just means that there is still at least one account which didn't reach the expected balance.

Geod24:
Updated according to @Charleslee522 's comments

PR #924, opened 2019-02-01
Simplify publishURL parsing in cmd

Geod24:
Pretty simple, instead of having two separated code path, we only branch out if there is no value.

### DIFF
--- a/cmd/sebak/cmd/run.go
+++ b/cmd/sebak/cmd/run.go
@@ -374,18 +374,17 @@ func parseFlagsNode() {
 	queries.Add("IdleTimeout", "3s")
 	bindEndpoint.RawQuery = queries.Encode()
 
-	if len(flagPublishURL) > 0 {
-		if p, err := common.ParseEndpoint(flagPublishURL); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/sebak/cmd/run.go
```diff
@@ -374,18 +374,17 @@ func parseFlagsNode() {
...
 	bindEndpoint.RawQuery = queries.Encode()
 
-	if len(flagPublishURL) > 0 {
-		if p, err := common.ParseEndpoint(flagPublishURL); err != nil {
-			cmdcommon.PrintFlagsError(nodeCmd, "--publish", err)
-		} else {
-			publishEndpoint = p
-			flagPublishURL = publishEndpoint.String()
-		}
+	// If there is no endpoint, just set as default
```
Charleslee522: This comment is the best part :)

Geod24:
Ping @Charleslee522 @spikeekips

cmd/sebak/cmd/run.go
Geod24: Except for the little typo (set a default, not "as default")

PR #926, opened 2019-02-07
Various code cleanup around test code

Geod24:
The title says it all. Review commit-by-commit to better see what's happening there.

### DIFF
--- a/lib/sync/fetcher_test.go
+++ b/lib/sync/fetcher_test.go
@@ -35,9 +35,9 @@ func TestBlockFetcher(t *testing.T) {
 	st := block.InitTestBlockchain()
 	defer st.Close()
 
-	kp := keypair.Random()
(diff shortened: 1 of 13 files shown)


lib/sync/fetcher_test.go
```diff
@@ -36,9 +35,9 @@ func TestBlockFetcher(t *testing.T) {
 	st := block.InitTestBlockchain()
 	defer st.Close()
 
-	kp := keypair.Random()
-	_, _, localNode := network.CreateMemoryNetwork(nil)
+	localNode := node.NewTestLocalNode0()
 
+	kp := keypair.Random()
```
soonkuk: Why did you remove from above line and inserted again?

Geod24: I just grouped it with its usage (`NewValidator`). Because a local node also takes a keypair.

PR #927, opened 2019-02-07
Simplify BlockOperation and remove unused members

Geod24:
One of my current goals is to merge `BlockOperation` into `BlockTransaction`. This should make the code much simpler and the DB much less heavy. However, there's a few things that are currently blocking this:
- Usage of operation hashes
- The million way operations are indexed with something else
- The amount of duplication between operation and transaction

This addresses some of the 2nd and 3rd point. Operation hashes should be dealt with by using an unified way to access blocks/txs/ops (e.g. hash/hash/idx, hash/idx/idx or height/idx/idx).

### DIFF
--- a/lib/block/operation.go
+++ b/lib/block/operation.go
@@ -31,10 +31,9 @@ type BlockOperation struct {
 	Height uint64                  `json:"block_height"`
 
 	// bellows will be used only for `Save` time.
-	seqID     uint64
-	operation operation.Operation
-	linked    string
-	isSaved   bool
+	seqID   uint64
+	linked  string
+	isSaved bool
 }
 
 func NewBlockOperationKey(opHash, txHash string) string {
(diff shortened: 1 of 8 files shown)


lib/block/operation.go
```diff
@@ -313,26 +309,15 @@ func GetBlockOperation(st *storage.LevelDBBackend, hash string) (bo BlockOperati
...
-			break
-		}
-	}
-	if !found {
-		err = errors.OperationNotFound
+// Looks up the operation referenced by `txHash`, then get the operation's hash from it
+func GetBlockOperationByIndex(st *storage.LevelDBBackend, txHash string, opIndex int) (BlockOperation, error) {
+	if bt, err := GetBlockTransaction(st, txHash); err != nil {
+		return BlockOperation{}, err
+	} else if opIndex >= len(bt.Operations) {
```
spikeekips: Unlike the existing code, `opIndex` can be under 0, so we need to check.
```go
} else if opIndex < 0 || opIndex >= len(bt.Operations) {
```

Geod24: In which case could it be `< 0` ?

spikeekips: @Geod24 `GetBlockOperationByIndex()` is called at `GetOperationsByTxHashOpIndexHandler` and it's `opindex` is just treated as `int`. Client can set the `opindex` as negative.

Geod24: Good point. A shame the type system doesn't allow us to express this :(

Geod24:
Updated according to @spikeekips 's comment

